Issue #11, opened 2016-01-29
Create nicer main page for the examples


ibgreen:
@wwwtyro - The local server (`nmp run start`) shows a directory listing including a lot of non-example related files. It would be nice to have a more visually appealing and clear landing page.

I am also wondering if we can now show multiple lessons in one page - with your latest changes we now support multiple parallel gl contexts. It'd be neat to just be able to scroll through all the lessons on a single page rather than click back and forth.

ibgreen:
Closed by @wwwtyro

PR #14, opened 2016-01-30
landing

wwwtyro:
@ibgreen @philogb 

Simple landing page.

screenshot from 2016-01-30 13-16-49


ibgreen:
@wwwtyro, 

I don't mind the carousel however I tend to update the README relatively frequently.  This makes it look like I will need to update in two places, README and index.html. I know that it is possible to put an image link directly in the markdown, that would probably be simpler long term for me rather than introducing an HTML file.

In the open issue regarding landing page #11, I was not referring to the README.md. I was mostly concerned about what happens when the user does an "npm run start" and opens the browser - he sees an ugly directory listing and has to find examples or lessons directory and click in there, instead of just starting one of the examples and being able to use the controls to go to the next one.

While we need to support `npm run start` for development purposes, I think the ultimate design for showcasing our demos would be to use something like github pages where we could serve the examples without asking the user to run a server. And then we could just put a link to the demo pages in our README.md. Then anyone could see the demos without even cloning the repo.

wwwtyro:
Okay, perhaps I misunderstood. With this PR, if someone runs `npm run start` and goes to localhost:3000, they'll be presented with the webpage above. Would you like something else?

Agree re: gh-pages, we'll just need to make sure the gh-pages repo has a built version of the code and that we keep it updated. Once this PR reaches a state you're happy with, set up a gh-pages branch and I'll make a pull request against it to get that started.

wwwtyro:
Rereading your comment, it sounds like you just want a redirect to a lesson?

ibgreen:
> Okay, perhaps I misunderstood. With this PR, if someone runs npm run start and goes to localhost:3000, they'll be presented with the webpage above. Would you like something else?

Sorry, the misunderstanding was mine. Just goes to show that I should always test run your changes before commenting. 

Yes this is a good start, I did want a nice landing page for the `npm run start`. One that looks good and makes it super easy to start the examples. My first impressions:
 - Most importantly, how do I start the examples from this page? That should be a big "click me" button.
 - Also, I don't think we need to duplicate all the contents in the README file. We could include the overview section/design goals for now, just to have some contents.

> Agree re: gh-pages, we'll just need to make sure the gh-pages repo has a built version of the code and that we keep it updated. 

Should we create a `prepublish` npm script that places built versions in dist/lumagl.js and dist/lumagl.min.js. And remove those from .gitignore? Then the built version could be picked up directly from the main repo, and it would be easier to use lumagl in simple web pages.

wwwtyro:
Yes, all that sounds good. I'll patch it up and ping you when I'm done. Thanks Ib!

wwwtyro:
Is that more what you had in mind, @ibgreen? I'm out of time this week, so I'll pick up the gh-pages stuff on Monday.

ibgreen:
@wwwtyro Yes, this looks awesome. Thanks for sorting it so quickly and enjoy your weekend!

ibgreen:
Closes #11

wwwtyro:
No problem! You too! :)

PR #15, opened 2016-02-02
gh-pages

wwwtyro:
@ibgreen @philogb 

Reorganization to get `gh-pages` running. See http://wwwtyro.github.io/luma.gl/ for a demo.

With this PR, we need only keep `gh-pages` and `master` synced in order to keep `gh-pages` updated. So if/when you merge this, Ib, can you make a new branch `gh-pages` off of `master` and push it to github? Then we should be good to go.


ibgreen:
gh-pages branch created.

wwwtyro:
Woo it's working! :D

philogb:
woohoo!

PR #31, opened 2016-02-08
Custom picking

wwwtyro:
@ibgreen @philogb 

Restores support for custom picking colors and adds an example. I'm sure this will need to be altered a little to get it to play nice with deck.gl, but the functionality is in place.

screenshot from 2016-02-07 21-51-27


wwwtyro:
This is included in #32.

PR #32, opened 2016-02-13
Babel 6 / deck.gl support

wwwtyro:
Adds support for babel 6 and updates for deck.gl interop.


src/scene.js
```diff
@@ -343,6 +338,58 @@ export default class Scene {
     return hash[[r,g,b]];
   }
 
+  pickCustom(x, y, opt = {}) {
+    const gl = this.gl;
+
+    if (this.pickingFBO === undefined) {
+      this.pickingFBO = new Framebuffer(gl, {
+        width: gl.canvas.width,
```
ibgreen: As I remember it, the picking buffer was previously quite small, we just a couple of pixels around the one we want to sample?

wwwtyro: Yes, I changed it to a simpler scissoring scheme, hence the buffer size change. There may be a way to do that with a small buffer; I can research it if it is an issue.

PR #43, opened 2016-03-04
Adds a bare-bones shadowmap example using es6 imports.

wwwtyro:
@ibgreen @mikolalysenko

Should address https://github.com/uber-common/luma.gl/issues/22 and https://github.com/uber-common/luma.gl/issues/8

image

### DIFF
--- a/examples/shadowmap/package.json
+++ b/examples/shadowmap/package.json
@@ -1,0 +1,34 @@
+{
+  "name": "luma.gl-examples-shadowmap",
+  "version": "0.0.0",
+  "description": "An es6 shadowmap example for luma.gl.",
+  "license": "MIT",
+  "author": "Ib Green <ib@uber.com>",
+  "dependencies": {
+    "babel-polyfill": "6.5.0",
+    "gl-format-compiler-error": "^1.0.2",
+    "global": "^4.3.0",
+    "glslify": "^5.0.2"
+  },
+  "browserify": {
+    "transform": [
+      "babelify",
+      "glslify"
+    ]
+  },
+  "devDependencies": {
+    "babel-cli": "6.5.1",
+    "babel-core": "6.5.1",
+    "babel-eslint": "4.1.8",
+    "babel-plugin-transform-builtin-extend": "1.1.0",
+    "babel-preset-es2015": "6.5.0",
+    "babel-preset-stage-2": "^6.5.0",
+    "babelify": "^7.2.0",
+    "browserify": "^13.0.0",
+    "budo": "^8.0.3"
+  },
+  "scripts": {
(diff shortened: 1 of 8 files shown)


wwwtyro:
Whoops, referenced the wrong issue, I meant https://github.com/uber-common/luma.gl/issues/22.

examples/shadowmap/src/index.js
```diff
@@ -0,0 +1,138 @@
+import 'babel-polyfill';
+
+const glslify = require('glslify');
+import * as LumaGL from '../../../src/index.js';
+
+window.webGLStart = function() {
+
+  var createGLContext = LumaGL.createGLContext;
```
ibgreen: If we are taking the trouble to run this file through babel (as opposed to just loading a prebuilt script) it could look nicer if we included the luma stuff through the import, as opposed to ES5 style destructuring.

import (createGLContext, Program, ... } from '../../../src';

(Note you can also drop the 'index.js' from the import.)

examples/shadowmap/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "luma.gl-examples-shadowmap",
+  "version": "0.0.0",
+  "description": "An es6 shadowmap example for luma.gl.",
+  "license": "MIT",
+  "author": "Ib Green <ib@uber.com>",
```
ibgreen: As long as we are keeping separate package.json files, I think you should take the full credit for this one.

examples/shadowmap/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "babel-eslint": "4.1.8",
+    "babel-plugin-transform-builtin-extend": "1.1.0",
+    "babel-preset-es2015": "6.5.0",
+    "babel-preset-stage-2": "^6.5.0",
+    "babelify": "^7.2.0",
+    "browserify": "^13.0.0",
+    "budo": "^8.0.3"
+  },
+  "scripts": {
+    "build": "browserify src/index.js > bundle.js",
```
ibgreen: What's the thinking here? Should the built version of this go into a local dist directory and be checked in so we can load this example directly from our gh-pages?

ibgreen:
@wwwtyro - Thanks, it looks like we are off to wonderful start on shadow mapping in luma.gl!

From a more practical point of view, how do we make these ES6 examples available in gh-pages (ideally in an "examples carousel"). See the comment about prebuilds above.

wwwtyro:
@ibgreen Thanks for the review! Re: gh-pages, yes, that was exactly what I had in mind, but I forgot to do it in my rush to push this last night. I'll have time tonight to fix that and address your other points here. :+1:

PR #46, opened 2016-03-05
Feature/shadowmap es6 example.1

wwwtyro:
@ibgreen @mikolalysenko

Response to comments in last PR.


ibgreen:
@wwwtyro Looks great, and I think it could make sense to organize all our examples like this, with separate package.json files. It allows us to show Luma.GL's intended ES6 programming style in the example code, and makes the examples more free standing and easier to copy and build from.

The only concern I have at this point is the size of the built bundles (currently 13K lines, 420K bytes), especially if we will be doing this for every example and every time we update we'll store a new version in the repo.

So it might be worth a quick investigation into ways to use multiple browserify bundles, essentially bundling only the example's code, and requiring luma.gl from the standard pre-built bundle (see for example: https://lincolnloop.com/blog/speedy-browserifying-multiple-bundles/)

Not sure if budo will work with multiple bundles though.

wwwtyro:
@ibgreen Yikes, yeah, that's gigantic - thanks for not merging this! :grinning: 

I hate to get rid of the es6 import syntax, though, since that was largely the point. Let me look into this and see what I can find.

Just brainstorming: how about another npm script command that goes through the examples and builds the ones that need building, and only putting the builds ones in gh-pages? Then if someone wanted to try them locally, they could run that script, and we'd only have the ballooning in the gh-pages branch, and they'd still be viewable on our gh-pages site.

wwwtyro:
@ibgreen Actually, along those lines... how do you feel about doing the same for dist/LumaGL.js?

ibgreen:
Before we go there, can we explore my proposal that we'd just exclude luma.gl when budo is building the example's bundle? Something like:
```
    "start": "budo main.js:dist/main.js --live --open --port 3000 -- -x luma.gl",
```
   and then having an index.html that includes both the standard luma.gl bundle and the example's own bundle
```
   <script "../../dist/LumaGL.js>  
   <script "./dist/main.js">
```

- Per the article I quoted, it should be possible to exclude a module from being bundled by browserify by using the `-x` command line option.

- We can pass through arguments from budo to browserify after `--`
https://github.com/mattdesl/budo/blob/master/docs/command-line-usage.md#browserify-arguments

wwwtyro:
@ibgreen Sure, no problem.

AFAICT, we'll need to add luma.gl as a dependency in each of the examples we do this with, since it seems to be using npm require semantics to resolve the module to be browserified when using the `-r` flag. e.g., this doesn't work:

```sh
$ npm run build-dev

> luma.gl@0.0.2 build-dev /home/rye/Dropbox/src/luma.gl
> browserify -r src/bundle.js -o dist/LumaGL.js

Error: Cannot find module 'src/bundle.js' from '/home/rye/Dropbox/src/luma.gl'
    at /home/rye/Dropbox/src/luma.gl/node_modules/resolve/lib/async.js:46:17
    at process (/home/rye/Dropbox/src/luma.gl/node_modules/resolve/lib/async.js:173:43)
    at ondir (/home/rye/Dropbox/src/luma.gl/node_modules/resolve/lib/async.js:188:17)
    at load (/home/rye/Dropbox/src/luma.gl/node_modules/resolve/lib/async.js:69:43)
    at onex (/home/rye/Dropbox/src/luma.gl/node_modules/resolve/lib/async.js:92:31)
    at /home/rye/Dropbox/src/luma.gl/node_modules/resolve/lib/async.js:22:47
    at FSReqWrap.oncomplete (fs.js:82:15)
```

But this does (using our already-installed `global` module as an example):

```sh
$ npm run foo

> luma.gl@0.0.2 foo /home/rye/Dropbox/src/luma.gl
> browserify -r global > foo.js
```

Just so there's no confusion - there's two steps to this process: first, browserify the external modules with the `-r` flag to make them requireable, then second, browserify the example source flagging the external modules with the `-x` flag to prevent them from being included. 

What are your thoughts?

ibgreen:
@wwwtyro Yes I suspect that you are right, we'd need luma.gl mentioned in the example's `package.json` for the `-x luma.gl` option to work.

I think it is possible to use relative paths in package.json dependencies, something like:
```
dependencies: {
   "luma.gl": "../../",
   ...
};
```
- This would presumably install a copy of luma.gl in the examples `node_modules` dir, which is OK since it would be is gitignored.
- This should also remove the need for the other dependencies in example.js, as they would be part of the luma.gl bundle.

ibgreen:
Another concern is whether the budo server would be able to serve a script file from outside it's start directory:
```
   <script "../../dist/LumaGL.js>  
   <script "./dist/main.js">
```
We need to prototype this to make sure we have found all the wrinkles...

wwwtyro:
:+1: Thanks, I'll see if I can get it to work with this technique.

wwwtyro:
Sorry @ibgreen, I'm not having a lot of luck with this. What I have right now is really hacky, and it doesn't even address the issue of budo not handling multiple bundles, so we lose the handy-dandy live reload. Would you like to give it a shot? I've spent a couple hours on it without a lot of luck.

wwwtyro:
Just cleaning some things up. :)

PR #47, opened 2016-03-05
Fix up jekyll site generation

mikolalysenko:
This PR fixes the static site generation for the docs on the GitHub pages and adds documentation links.

To run the static site generation, you need to install the `jekyll` and `pygments` ruby gems which can be done with the following command:

```sh
gem install jekyll pygments.rb
```

Once that's done, you can rebuild the static site using the following npm script:

```
npm run jekyll
```

This will create a folder called `_site` in the root directory which contains all the static content for the site.  To view the docs locally, you can also run jekyll as a standalone server with the command:

```
jekyll serve --source=docs/
```

CC @ibgreen @wwwtyro

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Luma.gl comes with 16 lessons, a number of examples, and a full set of
 reference documenation.
 
 To run examples:
-```
+```sh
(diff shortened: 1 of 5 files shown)


wwwtyro:
:+1:

PR #51, opened 2016-03-07
update event docs

wwwtyro:
@ibgreen @mikolalysenko

### DIFF
--- a/docs/event.md
+++ b/docs/event.md
@@ -35,7 +35,7 @@ Creates a set of events for the given domElement that can be handled through a c
 
 ### Arguments:
 
-1. app  - (*element*) A [LumaGL application](core.html#LumaGL:constructor).
+1. gl  - (*object*) A WebGL context. Events are handled for the context's canvas element.
 5. options - (*object*) An object containing the following options:
 
 ### Options:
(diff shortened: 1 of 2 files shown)


docs/event.md
```diff
@@ -35,7 +35,7 @@ Creates a set of events for the given domElement that can be handled through a c
 
 ### Arguments:
 
-1. app  - (*element*) A [LumaGL application](core.html#LumaGL:constructor).
+1. gl  - (*object*) A WebGL context. Events are handled for the context's canvas element.
```
ibgreen: The WebGL objects all have well defined types. In this case [WebGLRenderingContext](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext)

Let's embrace that and use it in our documtenation.

docs/event.md
```diff
@@ -94,7 +88,7 @@ Setting rotation and zoom to a moon object with drag and drop and mousewheel eve
 
     //create and assign variables to objects...
 
-    LumaGL.Events.create(app, {
+    LumaGL.Events.create(gl, {
```
ibgreen: To make justice to our ES6 port, I would prefer that we use the namespace less version in examples:
```
import {Events} from 'luma.gl';
Events.create(gl, ...
```

PR #59, opened 2016-03-16
Tool/download bitmaps

wwwtyro:
@ibgreen @mikolalysenko @philogb 

Adds a function to luma that will force a file download of a canvas in png format. Note that I built this on top of the particle demo PR, so this diff is needlessly large. [Here's the relevant commit.](https://github.com/uber-common/luma.gl/commit/04670d8cdea7e0763a0b914eef6c5f4417e06223)

[Here's a video I made with this functionality.](https://youtu.be/pOTyWZI2iX8)


examples/particles/index.html
```diff
@@ -0,0 +1,303 @@
+<html>
+
+  <head>
+    <title>Particle demo</title>
+    <meta http-equiv="content-type" content="text/html; charset=ISO-8859-1">
+
+    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/css/bootstrap.min.css" integrity="sha384-1q8mTJOASx8j1Au+a5WDVnPi2lkFfwwEAa8hDDdjZlpLegxhjVME1fgjWPGmkzs7" crossorigin="anonymous">
+    <link href="../examples.css" type="text/css" rel="stylesheet" media="screen" />
```
Qix-: Just a nit - you self-close 8 but don't self-close 7 here.

examples/particles/index.html
```diff
@@ -0,0 +1,303 @@
...
+
+    <script src="http://d1n0x3qji82z53.cloudfront.net/src-min-noconflict/ace.js" type="text/javascript" charset="utf-8"></script>
+    <script type="text/javascript" src="../../dist/LumaGL.js"></script>
+    <script type="text/javascript" src="../examples.js"></script>
+    <script type="text/javascript" src="index.js"></script>
+
+
+
+    <script id="scene-vs" type="x-shader/x-vertex">
+
```
Qix-: Why all these extra lines? :dancer:

examples/particles/index.html
```diff
@@ -0,0 +1,303 @@
...
+      #render-canvas {
+        width: 100%;
+        height: 100%;
+      }
+    </style>
+
+    <script src="http://d1n0x3qji82z53.cloudfront.net/src-min-noconflict/ace.js" type="text/javascript" charset="utf-8"></script>
+    <script type="text/javascript" src="../../dist/LumaGL.js"></script>
+    <script type="text/javascript" src="../examples.js"></script>
+    <script type="text/javascript" src="index.js"></script>
```
Qix-: Not sure you need the `type` attributes on here; could be argued the `charset` attribute could be left off as well.

examples/particles/index.html
```diff
@@ -0,0 +1,303 @@
...
+                      + dot(m1*m1, vec2( dot( p3, x3 ), dot( p4, x4 ) ) ) ) ;
+      }
+
+    </script>
+
+
+
+
+
+
```
Qix-: More empty lines :dancer:

ibgreen:
@wwwtyro Rye, this is great stuff! I think this PR covers the bitmap generation only, not the video assembly phase. Could you make an example, e.g. a node script or shell script or just a writeup, showing how you combined the generated bitmaps into a video, which tools you used etc?

@twojtasz and @heshan0131 are planning to use this in a couple of their apps, so they would benefit from anything you can put together

wwwtyro:
@ibgreen @twojtasz @heshan0131

Before I get to making the video, one quick note about generating the images: this doesn't play well with requestAnimationFrame. For that reason, I recommend the following pattern:

```javascript
function render() {
   // Render to your canvas. 

   // Instead of calling RAF, use setTimeout while you're recording frames:
   if (tick < 60) { // Here I'm just capturing the first 60 frames. You'll need to decide when to record a frame.
      saveBitmap(canvas, 'test-' + tick + '.png');
      setTimeout(function() {
        render(); // This could be requestAnimationFrame(render), if you're more comfortable with that, but it would still need to be inside the setTimeout callback.
      }, 0);
      return; // <-- Note that the return here prevents RAF being called below.
    }

    requestAnimationFrame(render);
}
```

To make the video from the resulting frames (which look like test-1.png, test-2.png, etc from the example above) I did the following:

```sh
ffmpeg -framerate 60 -start_number 0 -i test-%d.png -refs 5 -c:v libx264 -preset veryslow -crf 18 test.mp4
```

The details of that one liner are outside of my area of expertise & there's plenty of stuff online about it, so I'll leave that research to you guys. I think that it will probably be sufficient for your purposes as it is, though.

Oh, one more thing - consider fixing your canvas size to some normal video resolution like 1920x1080 or 1280x720 while you're recording frames. That'll prevent funky aspect ratios.

Let me know if you guys have any questions!

ibgreen:
@wwwtyro @mikolalysenko - This is great. I just wonder if it would be easy to "package this up" a little more? 
E.g. would this be easy? Create an example Node.js micro-server that has a couple of routes
/startVideo
/addFrame
- Use a data URL?
/endVideo filename
- On the server side we could use node's spawn to run the `ffmpeg` incantation?

And on the client side a small client class that starts a renderloop based on setTimer, sets up a render target with good default resolution, calls a render callback for specified number of frames, and hits the server routes above.

PR #69, opened 2016-04-10
Feature/fix examples.b

wwwtyro:
@ibgreen @philogb Gets lessons 3 - 8, & 10 - 12 mostly working. It looks like there might be an off-by-one error or the like somewhere that is preventing the rendering of a poly in meshes, see lesson 10 for an obvious example.

Instead of trampling your refactor, Ib, I've fixed bugs here in a very hacky way to highlight where the issues are. You'll want to come up with fixes that better align to your vision. I'll leave comments in the diff to try to explain some of the issues I found.


examples/lessons/11/index.js
```diff
@@ -28,15 +29,15 @@ window.webGLStart = function() {
   gl.depthFunc(gl.LEQUAL);
   gl.viewport(0, 0, +canvas.width, +canvas.height);
 
-  var program = makeProgramFromDefaultShaders(gl);
+  var program = new Program(gl, {});
```
wwwtyro: The `merge` function needs to be fixed so that it can handle undefined objects.

package.json
```diff
@@ -72,6 +72,9 @@
     "build-scripts": "npm run build-script-main && npm run build-script-addons",
     "build-script-main": "browserify src/bundle.js --debug -o dist/lumagl.js && cat dist/lumagl.js | uglifyjs > dist/lumagl.min.js",
     "build-script-addons": "browserify src/addons/index.js -o dist/lumagl-addons.js --debug | uglifyjs > dist/lumagl-addons.min.js",
+    "watch-scripts": "npm run watch-script-main & npm run watch-script-addons",
+    "watch-script-main": "watchify src/bundle.js -v --debug -o dist/lumagl.js",
+    "watch-script-addons": "watchify src/addons/index.js -v --debug -o dist/lumagl-addons.js",
```
wwwtyro: I added these in lieu of fixing the budo script, not sure if/how that can be done.

src/geometry.js
```diff
@@ -48,9 +48,15 @@ export default class Geometry {
     for (const attributeName in attributes) {
       const attribute = attributes[attributeName];
       if (isTypedArray(attribute)) {
+        let size = 3;
+        if (attributeName === 'instanced') {
+          size = 1;
+        } else if (attributeName === 'texCoords') {
+          size = 2;
+        }
```
wwwtyro: Might want some way to manually set the attribute size.

src/model.js
```diff
@@ -220,7 +220,8 @@ export default class Model extends Object3D {
         data: attribute.value,
         size: attribute.size,
         instanced: attribute.instanced ? 1 : 0,
-        bufferType: attribute.bufferType || program.gl.ARRAY_BUFFER,
+        bufferType: attribute.bufferType || attributeName === 'indices' ?
+          program.gl.ELEMENT_ARRAY_BUFFER : program.gl.ARRAY_BUFFER,
```
wwwtyro: Might want some way to indicate an ELEMENT_ARRAY_BUFFER is being created other than using the 'indices' string.

src/webgl/program.js
```diff
@@ -90,7 +90,18 @@ export default class Program {
...
+      if (buffer.attribute === 'vertices') {
+        location = this.attributeLocations.position;
+      } else if (buffer.attribute === 'colors') {
+        location = this.attributeLocations.color;
+      } else if (buffer.attribute === 'normals') {
+        location = this.attributeLocations.normal;
+      } else if (buffer.attribute === 'texCoords') {
+        location = this.attributeLocations.texCoord1;
+      }
+    }
```
wwwtyro: There's a mismatch between the shader attribute names and the ones being used in geometry. Don't use this horrible hack! :D

philogb:
Thanks for this @wwwtyro ! @ibgreen is out but will be back next week. I'll make sure he goes through this.

wwwtyro:
Spring cleaning!

PR #70, opened 2016-04-11
fix model.js geometry.js for none instanced geometry rendering

heshan0131:


src/model.js
```diff
@@ -318,7 +333,8 @@ export default class Model extends Object3D {
         Instanced: attribute.instanced,
         Verts: attribute.value.length / attribute.size,
         Size: attribute.size,
-        Bytes: attribute.value.length * attribute.value.BYTES_PER_ELEMENT
+        Bytes: attribute.value.length * attribute.value.BYTES_PER_ELEMENT,
+        location: program && program.attributeLocations[attributeName]
```
ibgreen: Use capital (Location) for consistency with other table titles?

Also, if location is undefined, it would be cool to check the bufferType and if it is ELEMENTS_ARRAY_BUFFER print something to indicate that this is the indices, perhaps simply print "ELEMENTS_ARRAY_BUFFER"?

If so, we should probably move the Location string calculation logic up above the object.

dist/webgl/program.js
```diff
@@ -244,8 +244,8 @@ function getUniformSetter(gl, glProgram, info, isArray) {
 
   var matrix = false;
   var vector = true;
-  var glFunction = undefined;
-  var TypedArray = undefined;
+  var glFunction = void 0;
```
ibgreen: Interesting. I wonder why these changed? Did you just pull in a later version of babel?

heshan0131: im in babelify 7.2.0

PR #71, opened 2016-04-16
Set up travis CI to work with headless-gl.

mikolalysenko:
This PR sets up travis CI to work with luma.gl.  As a proof-of-concept I wrote a simple test case for buffers to show that things are working.

I tried a few times to get testtron working, but kept having weird issues where the test process would hang and not exit for like 20 minutes.  headless-gl works, though I need to fix the prebuilt binaries to work with the correct version of GLIBC on linux.

CC @ibgreen @philogb

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "eslint": "^1.9.0",
     "eslint-plugin-react": "^4.2.3",
     "faucet": "0.0.1",
+    "gl": "^2.1.5",
     "glslify": "^5.0.2",
     "husky": "^0.10.2",
     "tap-browser-color": "^0.1.2",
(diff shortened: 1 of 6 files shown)


ibgreen:
@mikolalysenko this looks very good, thanks.
- FWIW, @vp also had problems when trying out testron with react-map-gl. Even when his test cases detected errors testron didn't always cause the build to fail, so Travis would be green in spite of failures. 
- And yes, I also saw cases where testron wouldn't close down. 
It's great that we are getting headless-gl in place, although ideally, we'd want both test approaches to work (i.e. a full browser env like testron and a webgl only env like headless-gl).

package.json
```diff
@@ -72,8 +73,8 @@
     "build-dev": "browserify src/bundle.js -o dist/LumaGL.js",
     "lint": "uber-standard",
     "precommit": "npm run lint -s",
-    "test": "browserify test/electron.js | testron | faucet",
-    "test-node": "babel-node --include='' test/node.js | faucet",
+    "test": "babel-node --include='' test/node.js | faucet",
```
ibgreen: @mikolalysenko I can change this later, but just to share my thoughts: 
I think we may end up with an evolving menu of test approaches. Naming them all and simply have "test" reference one of them by name makes it clear and easy to switch between them:
```
   "test": "npm run test-node",
   "test-node": "babel-node --include='' test/node.js | faucet",
   "test-headless": "babel-node --include='' test/webgl.js | faucet",
   "test-electron...", 
   "test-browser...", 
   ...
```
Not sure if it still makes sense to separate test-node and test-headless anymore. The idea was test-node would only test non-webl-gl dependent code, so it would always be runnable. If headless-gl is reliably installable on both mac and linux, we might be able to drop it, if not I'd consider keeping the separation for now.

mikolalysenko:
Yeah, I'll continue investigating what is going on with testron.  It is weird that it won't shut down sometimes, but even more frustrating that this situation only seems to happen on travis.  That said, a full DOM implementation would be very useful to have in larger integration tests.

PR #73, opened 2016-04-28
Add glslify-babel transform

mikolalysenko:
This PR switches the glslify transform in luma.gl to use the newly created [glslify-babel](https://github.com/stackgl/glslify-babel).  The advantage of this approach is that it should support ES6 syntax like imports and tagged template strings.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,10 +29,12 @@
     "global": "^4.3.0"
   },
   "browserify": {
-    "transform": [
-      "babelify",
-      "glslify"
-    ]
+    "transform": [[
+      "babelify", {
+        "presets": ["es2015"],
+        "plugins": ["glslify-babel"]
+      }
+    ]]
   },
   "standard": {
     "ignore": [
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -58,6 +60,7 @@
     "faucet": "0.0.1",
     "gl": "^2.1.5",
     "glslify": "^5.0.2",
+    "glslify-babel": "^1.0.0",
```
ibgreen: FWIW, I notice that these tend to be named `babel-plugin-...` But should be OK as is.

package.json
```diff
@@ -29,10 +29,12 @@
...
   },
   "browserify": {
-    "transform": [
-      "babelify",
-      "glslify"
-    ]
+    "transform": [[
+      "babelify", {
+        "presets": ["es2015"],
+        "plugins": ["glslify-babel"]
```
ibgreen: I moved this to `.babelrc`

ibgreen:
@mikolalysenko the action is on dev branch, I pushed similar changes there, please check out and try `npm run build`

ibgreen:
Closing as this was merged separately into dev

PR #77, opened 2016-05-05
Implement rest of image loading operations and add test case

mikolalysenko:
This PR adds a test case for image loading and also finishes the work in progress version on the browser side. 

Right now this PR works fine with browser and electron tests, but currently node is failing.  The reason for this is that I am getting some weird error involving `FileSaver.js`.  Here is a stack trace:

```
/Users/mikolalysenko/GitHub/luma.gl/node_modules/filesaver.js/FileSaver.js:22
		  doc = view.document
		            ^

TypeError: Cannot read property 'document' of undefined
    at /Users/mikolalysenko/GitHub/luma.gl/node_modules/filesaver.js/FileSaver.js:22:15
    at Object.<anonymous> (/Users/mikolalysenko/GitHub/luma.gl/node_modules/filesaver.js/FileSaver.js:241:2)
    at Module._compile (module.js:413:34)
    at Module._extensions..js (module.js:422:10)
    at Object.require.extensions.(anonymous function) [as .js] (/Users/mikolalysenko/GitHub/luma.gl/node_modules/babel-register/lib/node.js:166:7)
    at Module.load (module.js:357:32)
    at Function.Module._load (module.js:314:12)
    at Module.require (module.js:367:17)
    at require (internal/module.js:16:19)
    at Object.<anonymous> (browser-fs.js:3:1)
```

I think something is going wrong with babel-node, though I'm not sure.  Time permitting I can investigate more deeply tomorrow.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ _docs
 examples/winds/data/*.html
 node_modules
 npm-debug.log
+test/data/test.png
--- a/io/browser/browser-fs.js
+++ b/io/browser/browser-fs.js
@@ -17,7 +17,7 @@ const Blob = window.Blob;
  * @param {Function} callback - Standard node (err, data) callback
  * @return {Promise} - promise, can be used instead of callback
  */
-export function writeFile(file, data, options, callback) {
+function writeFile(file, data, options, callback) {
   // options is optional
   if (callback === undefined && typeof options === 'function') {
     options = undefined;
@@ -51,7 +51,7 @@ export function writeFile(file, data, options, callback) {
  * @param {File|Blob} file  HTML File or Blob object to read as string
  * @returns {Promise.string}  Resolves to a string containing file contents
  */
-export function readFile(file) {
+function readFile(file) {
   return new Promise((resolve, reject) => {
     try {
       assert(File, 'window.File not defined. Must run under browser.');
@@ -104,4 +104,9 @@ function getFileErrorMessage(e) {
   default:
     return 'Read error.';
   }
+}
+
+export default {
+  writeFile,
+  readFile
 }
--- a/io/browser/image-io.js
+++ b/io/browser/image-io.js
@@ -2,6 +2,7 @@
 /* global document, HTMLCanvasElement, Image */
 /* eslint-disable guard-for-in, complexity, no-try-catch */
 import assert from 'assert';
+import through from 'through';
 
 /*
  * Returns data bytes representing a compressed image in PNG or JPG format,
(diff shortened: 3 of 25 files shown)


ibgreen:
@mikolalysenko 
>adds a test case for image loading 

Thanks, this is a great start! This test case (TC1) (reading from/saving to a string) is a good cross-environment (node+browser) test. I think we need a few more TCs, some ideas:
- **TC2**: Roundtrip load/save: a node specific test case that actually reads from/saves to files (rather than strings) under node. 
    - We could place a reference image in test/data/test.png, and read it and then write it back to a temp folder, and then do a byte compare on the two files.
    - This TC would go in the test/node folder
-  **TC3**: actually create a texture out of a loaded image file. This shows that the DOM-less WebGL integration for image load is working.
    - This TC would go in the test/gpu folder.
- **TC4**: A render target is rendered with a test pattern, and then the contents of that render target are saved and byte compared with a reference file.
   - This TC would essentially prove that we have the all the image "io" we need for the video generation use case.
    - This TC would go in the test/gpu folder.
- **TC5** Roundtrip load/texture/render/save - Load a texture, render it to a render target, save it, byte compare files.
    - This TC would go in the test/gpu folder.

> I am getting some weird error involving `FileSaver.js`

- My guess is that FileSaver.js simply makes indiscriminate access to the DOM (window, document globals etc) and has not been tested under node. 
- In fact, I am not even sure we need a big polyfill for window.saveAs anymore.
- For now you can just comment out the fileSaver.js import and stub out any reference to it. I will look into that later.

> and also finishes the work in progress version on the browser side. 

This is probably bike shedding, and I assume you were mainly trying to get things working (i.e. something in the existing, somewhat unfinished, import/export structure was broken), but the export changes might be a step backwards. 

In ES6 I usually export individual named symbols and often import only the ones I want using destructuring imports. 
```
import {readFile} from './browser-fs';
```
If I want them all then I import them with something like
```
import * as browserFs from './browser-fs';
```
In fact, apart from the exporting code looking less readable (IMHO), I am actually not sure that destructing import still works from an `export default` statement, as babel transpiles the default export to a custom object with a `default` field on the same level as the named exports.

Conclusion: the changed import/exports are not a blocker, once the additional TC is in place. But I would probably restore the original import/export structure after merging the PR.

mikolalysenko:
Yeah, the reason I changed the imports around was that babel was getting screwed up with the `export {foo} from 'bar';` syntax.  I'll take another look once all the features are implemented and tested.

mikolalysenko:
Alright, I added all of the cases except for TC5 because I still don't have a great solution for saving the image from the browser side.  At the moment, save-image returns a stream while the reading interface requires a file.

I think we could fix this by making the save-image case only work in node or by modifying the read interface to accept a stream.  I am kind of leaning toward the former case, since the stream interface would increase overhead in the browser, but I'm open to either solution.

ibgreen:
@mikolalysenko This looks great. 

> I think we could fix this by making the save-image case only work in node or by modifying the read interface to accept a stream.

About TC5, I definitely want to support a stream api in the browser, the IO module is there to take care of the differences between the browser and node environments after all. And I don't really see it as overhead since browser support for streams is coming, the new "fetch" api will support opening request streams directly. 

I don't know if this module is a good option, I am sure there are many others https://github.com/devongovett/blob-stream.

PR #84, opened 2016-06-26
Add npm script that helps push to gh-pages.

episodeyang:
### Intention

@ibgreen asked me to add a script to luma.gl to help publish the newest version of it's documentation that has been in a long-lived branch for a while.

### Usage 

run `npm run publish-gh-pages` to publish. Whether this works depend on many states.

### Demo Link

Live demo here: [http://www.episodeyang.com/luma.gl/](http://www.episodeyang.com/luma.gl/)

**note that the domain is not intentional.** GitHub Pages of my fork is automatically published under [episodeyang.github.io/luma.gl](episodeyang.github.io/luma.gl), where `episodeyang.github.io` is redirected to my blog. This page will disappear as soon as I delete my fork of `luma.gl` after this PR is accepted.

### Explanation of Method

This publication method uses the `git subtree` command. This is the most convenient method, however it does require one to include the generated `./_docs` folder in git whenever a new version is updated. Many think this is a worthwhile sacrifice for the convenience.

You can look at the source of that script, and remove some of the `--no-verify` flag when your git pre-commit hook works. 

The `gh-pages` script is the one that is called. However when there is an existing gh-page upstream version (the case atm), a forced version of the same command need to be used (`gh-pages-forced`). 

Because the git `subtree` command does not have a `--force` option, this is realized with:

```shell
git push origin `git subtree split --prefix _docs dev`:gh-pages --force
``` 

@ibgreen @chrisirhc


package.json
```diff
@@ -80,6 +80,9 @@
     "build-compile": "babel src -d dist --source-maps inline --copy-files",
     "build-script": "browserify src/bundle.js --debug -o dist/lumagl.js -t babelify -t glslify && cat dist/lumagl.js | uglifyjs > dist/lumagl.min.js",
     "docs": "npm run build-docs && open ./_docs/index.html",
+    "gh-pages": "git subtree push --prefix _docs origin gh-pages",
```
ibgreen: Maybe have a common 'doc' prefix for all these scripts. When the list of script grows it gets easy to lose track of what's what.
```
doc-gh-pages
doc-gh-pages-force
doc-publish
 ```

ibgreen:
@episodeyang 
Thanks for doing the research and for the detailed explanation.
- I really like the approach of creating a script for this, and I think including the generated docs in the main branch should be fine.
Before I pull
- it would be great if we could find a way using .gitattributes or github's linguist to defeat the generated files from showing up in the PR diffs, as they overwhelm the diff and make it very hard to review for any other changes.
- To that end it would be great if this PR could be squashed into two commits (one with script etc changes, and one with only generated docs.

PR #90, opened 2016-08-01
Fixes to examples

ibgreen:
All examples have been fixed on the `fix-examples` branch (except lesson 15/16). 
But I haven't been able to update gh-pages using the new scripts.
@manassra When you come online again perhaps you can take a quick look?
@philogb


ibgreen:
npm install --save autobind-decorator should fix the CI issue

PR #91, opened 2016-08-10
docs path updated

doug:


ibgreen:
@doug Thanks for the PR. We ended up doing a bigger overhaul of the examples docs which changed these links, but it took longer than we expected to complete. Examples and documentation links should work now.

PR #93, opened 2016-09-02
Attribute manager: remove separate updater argument, remove need for …

ibgreen:
…positional tags


ibgreen:
@gnavvy

ibgreen:
This is a breaking change for deck.gl will be published as a 3.0.0-beta release.

PR #100, opened 2016-09-12
isomorphic rendering fix - restore navigator override

ibgreen:


ibgreen:
@gnavvy @abmai  Ensures the navigator object is restored after the filesaver hack.

PR #111, opened 2016-11-30
Add test case and fix for glGetDebugInfo regression under Node.

ibgreen:
@shaojingli @gnavvy - quick regression fix.


test/webgl/index.js
```diff
@@ -1,24 +1,27 @@
+// No headless
+// import './context-no-headless.spec';
```
gnavvy: do we need this line to include the new test?

ibgreen: Good question
No that file is imported from a different test run (without headless-gl imported).
We run with and without headless gl. I have pushed a commit that removes the commented line in this index.js file.

PR #116, opened 2016-12-16
Very basic test that addEvents doesn't crash.

tehwalris:
> In addition to manual tests, it would be great with a CI sanity test script that just checked that imports are working and that addEvents is called without crashing.

I'm not sure how valuable this alone is. I can't see a way to add more useful tests without starting to mock whole DOM events - that doesn't seem like a good test. Is this single simple test worth adding?

### DIFF
--- a/src/experimental/event/test/headless/add-events.spec.js
+++ b/src/experimental/event/test/headless/add-events.spec.js
@@ -1,0 +1,19 @@
+import test from 'tape-catch';
+import {addEvents} from '../../index.js';
+
+class MockDomElement {
+  _listenersByType = {}
+
+  addEventListener(type, listener) {
+    this._listenersByType[type] = listener;
+  }
+}
+
+test('Experimental#addEvents with no user handlers', t => {
+  const element = new MockDomElement();
+
+  addEvents(element, {});
+  t.pass('addEvents did not throw');
+
+  t.end();
+});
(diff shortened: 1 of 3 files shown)


src/experimental/event/test/headless/add-events.spec.js
```diff
@@ -0,0 +1,19 @@
...
+  addEventListener(type, listener) {
+    this._listenersByType[type] = listener;
+  }
+}
+
+test('Experimental#addEvents with no user handlers', t => {
+  const element = new MockDomElement();
+
+  addEvents(element, {});
+  t.pass('addEvents did not throw');
```
ibgreen: There is a tape "assert" for checking exceptions:
```
 t.doesNotThrow(
    () => addEvents({}),
   'addEvents did not throw'
);
```

ibgreen:
> I'm not sure how valuable this alone is. I can't see a way to add more useful tests without starting to mock whole DOM events - that doesn't seem like a good test. Is this single simple test worth adding?

Yes absolutely - Even just a test that checks that a symbol is correctly imported is a valuable safety net in future refactors.

If we could have called `addEvents` with a couple of event specifiers instead of an empty object, that would have been neat, but if it would require pulling in something like JSDom that would probably be too much.

FWIW, we used to have electron based tests (`npm run test-electron`) but they haven't been maintained. I suspect that could have worked great in this case. I'd be happy to support if you'd be interested in trying that.

Another option would be to add internal tests - i.e. unit tests of the various event manager classes and utilities.

PR #121, opened 2017-01-26
Performance query using EXT_disjoint_timer_query

shaojingli:
Timer query using EXT_disjoint_timer_query. The implementation uses direct webgl calls instead of the TimerQuery class for simplicity. I'll do a cleanup of the TimerQuery class and migrate the code on top of it in the future.

### DIFF
--- a/src/core/model.js
+++ b/src/core/model.js
@@ -60,6 +60,7 @@ export default class Model extends Object3D {
     render = null,
     onBeforeRender = () => {},
     onAfterRender = () => {},
+    timerQueryEnabled = false,
     ...opts
   } = {}) {
     // assert(program || program instanceof Program);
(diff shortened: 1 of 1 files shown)


src/core/model.js
```diff
@@ -116,6 +117,16 @@ export default class Model extends Object3D {
 
     this.onBeforeRender = onBeforeRender;
     this.onAfterRender = onAfterRender;
+
+    this.timeElapsedQuery = undefined;
```
ibgreen: A bit too much code to add here, would be best to have a separate method/util to handle this.

src/core/model.js
```diff
@@ -60,6 +60,7 @@ export default class Model extends Object3D {
     render = null,
     onBeforeRender = () => {},
     onAfterRender = () => {},
+    timerQueryEnabled = false,
```
ibgreen: Maybe just call it:
```
    profile: true
```

src/core/model.js
```diff
@@ -262,6 +280,38 @@ export default class Model extends Object3D {
...
+
+        if (available) {
+          const timeElapsed = this.ext.getQueryObjectEXT(this.timeElapsedQuery,
+            this.ext.QUERY_RESULT_EXT) / 1e6;
+          this.accumulatedFrameTime += timeElapsed;
+          this.averageFrameTime = this.accumulatedFrameTime / this.profileFrameCount;
+          // Do something useful with the time.  Note that care should be
+          // taken to use all significant bits of the result, not just the
+          // least significant 32 bits.
+          log.log(2, 'program.id: ', this.program.id);
```
ibgreen: +1.

For the long-term, a nice way to handle this would be to update a global `perf` or `stats` object, and make a stats widget that can show this in the UI.

There is three.js stats widget and regl-stats widget etc we could investigate, maybe we could use directly.

ibgreen:
Just saw your comment on `TimerQuery`. Sounds good.

PR #123, opened 2017-02-15
Fix instancing to work with webgl2 context

philogb:
@ibgreen wasn't sure whether there was a better way to capture if we're working with a webgl 2 context. I'm working on webgl2 and had to patch luma this way to be able to work with it.

### DIFF
--- a/src/webgl/context.js
+++ b/src/webgl/context.js
@@ -79,14 +79,12 @@ export function createGLContext(opts = {}) {
 
     // Prefer webgl2 over webgl1, prefer conformant over experimental
     if (webgl2) {
-      gl = canvas.getContext('webgl2', opts);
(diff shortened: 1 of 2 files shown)


ibgreen:
I have a pretty comprehensive set of polyfills, capability detections, and new classes to support new WebGL2 objects but haven't had time to merge them. 

This change is fine in the meantime. Also, there is a function called isWebGL2Context that might be better than setting a flag, especially as the context object is replaced when activating debug mode.
```
import {isWebGL2Context} from './webgl-checks'; 
```

philogb:
That's a good point - will use that instead and update the PR

philogb:
@ibgreen please let me know if this new code works.

PR #129, opened 2017-03-03
Fix global variable initialization

ibgreen:


ibgreen:
@pessimistress I think the problem here might be related to module syntax and possibly tree shaking - i.e. importing a single object from luma.gl does not execute all the imports. I made all the files that referencer to the global state import `luma` from `init.js` to force initialization.

PR #130, opened 2017-03-15
3.0.0 release - dev rebased to master

ibgreen:
Current `master` has been branched to `2.0-release`, time to merge `dev` into `master`, publish an official 3.0.0 release.

@shaojingli @gnavvy @Pessimistress 

Note: This will allow us to start importing luma.gl with semver wildcards again (e.g. "^3.0.0") rather than hard coding alpha versions, which is important for upcoming the deck.gl release. We'll need to update examples etc there once this is published.


ibgreen:
> Not sure how to review this 8000-file diff...

Yes, the large number of files is due to the removal of a bunch of unused demos inherited from the initial PhiloGL fork. They were never ported to luma so no need to keep them around in the tree.

PR #134, opened 2017-03-23
add CORS setting for image loading

gnavvy:
Current config prevents from loading images from another domain (e.g. S3) as texture to layers like the IconLayer.

ref:
https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_settings_attributes

### DIFF
--- a/src/io/browser-load.js
+++ b/src/io/browser-load.js
@@ -9,7 +9,6 @@ export function loadFile(opts) {
 /*
  * Loads images asynchronously
  * returns a promise tracking the load
- * TODO - CORS support
  */
 export function loadImage(url) {
   return new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


src/io/browser-load.js
```diff
@@ -9,14 +9,14 @@ export function loadFile(opts) {
...
  * returns a promise tracking the load
- * TODO - CORS support
  */
 export function loadImage(url) {
   return new Promise((resolve, reject) => {
     try {
       const image = new Image();
       image.onload = () => resolve(image);
       image.onerror = () => reject(new Error(`Could not load image ${url}.`));
+      image.crossOrigin = 'anonymous';
```
ibgreen: Ideally it should be the default value of an optional parameter.

PR #136, opened 2017-04-22
WebGL refactor

ibgreen:
@shaojingli - Apologies for this monster diff, the reason is that I've been working on a parallel track for a long time, but due to other priorities have not had the opportunity to merge back.

In spite of appearances, I have actually broken my work into several PRs. 

This is the first PR that contains the trivial, mostly uninteresting refactoring, so in spite of its size it should be sufficient to scan through it pretty quickly.

Once we get this in, I can open small focused PRs on new features like state management, WebGL2 support etc.


src/core/model.js
```diff
@@ -455,4 +458,17 @@ export default class Model extends Object3D {
   setTextures(textures = []) {
     throw new Error('model.setTextures replaced: setUniforms({sampler2D: new Texture2D})');
   }
+
+  isPickable() {
```
shaojingli: Since picking is not generally implemented in luma.gl, do we still want to have this property here

src/webgl/buffer.js
```diff
@@ -120,7 +83,7 @@ export default class Buffer {
   setData({
     data,
     bytes,
-    target = GL.ARRAY_BUFFER,
+    target = this.target || GL.ARRAY_BUFFER,
```
shaojingli: Is this.target still here?

src/webgl/context.js
```diff
@@ -23,21 +32,46 @@ and import 'luma.gl/headless' before importing 'luma.gl'.`;
 const ERR_HEADLESSGL_NOT_AVAILABLE =
 'Cannot create headless WebGL context, headlessGL not available';
 
-const ERR_HEADLESSGL_FAILED =
-'headlessGL failed to create headless WebGL context';
+const ERR_HEADLESSGL_FAILED = 'headlessGL failed to create headless WebGL context';
+
+export function isWebGLContext(gl) {
+  return gl && (gl instanceof WebGLRenderingContext ||
+    gl.ARRAY_BUFFER === GL_ARRAY_BUFFER);
```
shaojingli: Is this the only way to detect if a debug context is WebGL 1 or WebGL 2? I'm sure it works, but make me feel uncomfortable

src/webgl/context.js
```diff
@@ -196,17 +235,18 @@ export function glContextWithState(gl, {scissorTest, framebuffer}, func) {
...
-  // checking whether info is null here. Firefox doesn't follow the
-  // specs by returning null for unsupported extension. Instead,
-  // it returns an object without GL_UNMASKED_VENDOR_WEBGL and GL_UNMASKED_RENDERER_WEBGL.
-  return {
-    vendor: (info && info.UNMASKED_VENDOR_WEBGL) ?
-      gl.getParameter(info.UNMASKED_VENDOR_WEBGL) : 'unknown',
-    renderer: (info && info.UNMASKED_RENDERER_WEBGL) ?
-      gl.getParameter(info.UNMASKED_RENDERER_WEBGL) : 'unknown'
-  };
+  return getGLContextInfo(gl);
```
shaojingli: Why do we create a new method getGLContextInfo() here and make glGetDebugInfo() simply call it?

src/webgl/framebuffer.js
```diff
@@ -1,67 +1,48 @@
...
-    this.stencilBuffer = null;
-    this.texture = null;
-    this.userData = {};
-    this.width = 0;
-    this.height = 0;
+    super(gl, opts);
+    this.initialize(opts);
     Object.seal(this);
-
     this.resize(opts);
```
shaojingli: I don't really get it... why do we call resize (which is essentially a this.initialize() call,  right after another this.initilize() call?

src/webgl/framebuffer.js
```diff
@@ -77,37 +58,37 @@ export default class Framebuffer {
...
-    .setImageData({
+    const colorBuffer = new Texture2D(this.gl, {
       data: null,
+      format,
+      type,
       width,
       height,
-      type,
-      format
+      [GL.TEXTURE_MIN_FILTER]: this.minFilter,
```
shaojingli: didn't really understand the JS syntax used here...

src/webgl/framebuffer.js
```diff
@@ -158,28 +136,48 @@ export default class Framebuffer {
...
     pixelArray = null
   }) {
-    const {gl} = this;
-
     // Deduce type and allocated pixelArray if needed
     if (!pixelArray) {
       // Allocate pixel array if not already available, using supplied type
       type = type || GL.UNSIGNED_BYTE;
-      const ArrayType = glTypeToArray(type);
+      const ArrayType = getTypedArrayFromGLType(type);
```
shaojingli: ArrayType => TypedArray? or TypedArrayClass?

src/webgl/functions.js
```diff
@@ -0,0 +1,149 @@
+/* eslint-disable */
```
shaojingli: This file name is to ambiguous to me. I feel the previous src/webgl/draw.js just works... If lumping functions with various purposes into a single file is what we want, we might call it utility.js, or wrapper.js or similar.

Just for discussion:
I like the way gl.readPixels got wrapped with a readPixels() function, so that deck.gl or any users of luma.gl don't need to call WebGL APIs directly.

But, since one of the major goals of luma.gl is to bridge WebGL's "bind-and-operate-on-global-state" model with more user-friendly "operate-on-object" mode, is it possible to have clear() and readPixels() function part of some objects instead of standalone functions on a global objects?

src/webgl/gl-constants.js
```diff
@@ -0,0 +1,870 @@
+// WEBGL BUILT-IN TYPES
```
shaojingli: These are some incredible work! Did you actually go through the GL header and copied them line by line? wow...

src/webgl/texture-3d.js
```diff
@@ -0,0 +1,176 @@
+import GL from './api';
+import {assertWebGL2Context, withParameters} from './context';
+import Texture from '../webgl/texture';
+import Buffer from './buffer';
+
+export default class Texture3D extends Texture {
+  /**
+   * @classdesc
+   * 2D WebGL Texture
```
shaojingli: nit: 3D WebGL Texture

src/webgl/texture.js
```diff
@@ -1,60 +1,341 @@
...
+
+  /**
+   * Defines a two-dimensional texture image or cube-map texture image with
+   * pixels from the current framebuffer (rather than from client memory).
+   * (gl.copyTexImage2D wrapper)
+   *
+   * Note that binding a texture into a Framebuffer's color buffer and
+   * rendering can be faster.
+   */
+  copyFramebuffer({
```
shaojingli: copyFromFramebuffer?

src/core/model.js
ibgreen: It is used but in deprecated code.
https://github.com/uber/luma.gl/search?utf8=%E2%9C%93&q=isPickable&type=

Also there is a picking example that uses this (the one that shows the different planets)

Agree we should review this, but I think it would add too much to this already big PR.

src/webgl/buffer.js
ibgreen: > Is this.target still available for the Buffer class?

In my latest test I was surprised to see that I couldn't change the target of a Buffer once created. So setting the target is important.

Is that the answer or is it the use of `this.target` as a default parameter that is the concern? It works well, although I can't find a clear guarantee for it in the standard.

src/core/model.js
shaojingli: Agree

src/webgl/framebuffer.js
ibgreen: The dynamic object keys? They allow us to use a variable (in this case a "constant") as the key in an object.

```
const a = "dynamicKey";
const obj = {[a]: "value"}
```

```
var a = "dynamicKey";
var obj = {};
obj[a] = "value";
```

The idea is that instead of defining hundreds of new luma.gl specific keys (in this case e.g. 'textureMinFilter`) we just use the GL constants people are already familiar with.

src/webgl/functions.js
ibgreen: >  clear() and readPixels() function part of some objects instead of standalone functions on a global objects?

I like this idea, I also think they look a little odd here. The question is just "which object"?

I have started to think that `draw` should probably be a method on `Program` - it runs the shaders in the program selected with `Program.use`

`clear` and `readPixels` seem harder to associate with any existing object. I believe that they work on the currently selected framebuffer - and we don't have an object for the global framebuffer.

I suppose they could be static methods on `Framebuffer`...

src/webgl/gl-constants.js
ibgreen: > Did you actually go through the GL header and copied them line by line? wow...

Yep, every character was artfully typed by hand...

Truth be told, Sublime has some great party tricks for editing hundreds of lines at the same time.

PR #142, opened 2017-04-25
Add benchmarking scaffolding plus a benchmark test for Program

ibgreen:
Add benchmarking support

Initial test case confirms that we need to get shader caching sorted asap.

Even for trivial shaders we see a big cost for recompiling shaders. For many instances of a big shader (like multiple copies of a deck.gl layer), the cost will be much higher.

```
Starting bench...
* Program from shader source x 121 ops/sec ±1.64% (52 runs sampled)
* Program from cached shaders x 180 ops/sec ±1.95% (58 runs sampled)
```


src/packages/math/src/euler.js
```diff
@@ -1,8 +1,8 @@
 import MathArray from './math-array';
 import {checkNumber, clamp} from './common';
-import {Matrix4} from './matrix4';
-import {Quaternion} from './quaternion';
-import {Vector3} from './vector3';
+import Matrix4 from './matrix4';
+import Quaternion from './quaternion';
+import Vector3 from './vector3';
```
Pessimistress: +1 for fixing this class, has been broken for a while.
But should be on a separate PR?

demo/index.html
```diff
@@ -0,0 +1,306 @@
+<!DOCTYPE html>
```
Pessimistress: What's changed with this file, just moving?

src/packages/math/src/euler.js
ibgreen: > But should be on a separate PR?

Yes, I suppose that could have made sense. This PR could have been broken into three really small pieces. I'll be a little less trigger happy with the merge button going forward.

demo/index.html
ibgreen: Yes webpack-dev-server needs a short index.html in the root directory to run the bench tests.

Since examples are broken on master and we are moving to a demo structure anyway, I moved this index file into the intended target directory now.

PR #146, opened 2017-04-25
GL state and limit management

ibgreen:
Initial GL context and limit management.

Documentation started in 
* [context](https://github.com/uber/luma.gl/blob/state-management/docs/webgl/context.md)
* [extensions](https://github.com/uber/luma.gl/blob/state-management/docs/webgl/extensions.md)
* [limits](https://github.com/uber/luma.gl/blob/state-management/docs/webgl/limits.md)

### DIFF
--- a/src/webgl/context-state.js
+++ b/src/webgl/context-state.js
@@ -1,0 +1,540 @@
+/* eslint-disable no-inline-comments, max-len */
+import GL from './webgl-constants';
+import assert from 'assert';
+
+// WebGL specification 'types'
+const GLenum = 'GLenum';
+const GLfloat = 'GLfloat';
+const GLint = 'GLint';
+const GLuint = 'GLint';
+const GLboolean = 'GLboolean';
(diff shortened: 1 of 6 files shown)


src/webgl/context-state.js
```diff
@@ -0,0 +1,540 @@
...
+class GLState {
+  // Note: does not maintain a gl reference
+  constructor(gl, {copyState = false} = {}) {
+    this.state = {};
+    if (copyState) {
+      this._copyWebGLState(gl);
+    } else {
+      this._getInitialState();
+    }
+    this.overrides = [];
```
shaojingli: I feel that the name "overrides" is not intuitive enough. Can we call it stateStack? or similar?

ibgreen: > I feel that the name "overrides" is not intuitive enough. Can we call it stateStack? or similar?

Good suggestion, done.

PR #150, opened 2017-04-26
feat(dev): seer integration

Apercu:
Add integration with the soon-to-be born seer.
No performance drawback for production on normal users given the checks over its presence.

Pass the model uniforms and allow to modify some of it. Does not persist across renders given the deck.gl overriding most of it, but can be done if deemed necessary.

### DIFF
--- a/src/core/model.js
+++ b/src/core/model.js
@@ -21,8 +21,11 @@ import {
 
 import Object3D from '../deprecated/scenegraph/object-3d';
 import {log, formatValue} from '../utils';
+import {window} from '../utils/globals';
 import {SHADERS} from '../experimental/shaders';
+import {addModel, removeModel} from '../debug/seer-integration';
 import assert from 'assert';
+import seer from 'seer';
 
 const MSG_INSTANCED_PARAM_DEPRECATED = `\
 Warning: Model constructor: parameter "instanced" renamed to "isInstanced".
@@ -136,6 +139,7 @@ export default class Model extends Object3D {
   /* eslint-enable complexity */
 
   destroy() {
+    removeModel(this.id);
   }
 
   setNeedsRedraw(redraw = true) {
@@ -241,6 +245,9 @@ export default class Model extends Object3D {
    */
   /* eslint-disable max-statements */
   render(uniforms = {}) {
+    if (window.__SEER_INITIALIZED__) {
+      addModel(this);
+    }
     const resolvedUniforms = this.addViewUniforms(uniforms);
 
     this.setUniforms(resolvedUniforms);
@@ -377,8 +384,14 @@ export default class Model extends Object3D {
         program: this.program,
         uniforms: Object.assign({}, this.uniforms, uniforms)
       });
+
       log.table(priority, table);
       log.log(priority, `${unusedCount || 'No'} unused uniforms `, unusedTable);
+    }
+
+    if (window.__SEER_INITIALIZED__) {
+      const uniformsObject = Object.assign({}, this.uniforms, uniforms);
+      seer.indexedListItem('luma.gl', this.id, uniformsObject);
     }
   }
 
--- a/test/node.js
+++ b/test/node.js
@@ -1,6 +1,10 @@
 // Enables ES2015 import/export in Node.js
 require('reify');
 
+// Mock addEventListener on window, required for seer
+const {window} = require('../src/utils/globals');
+window.addEventListener = () => {};
+
 // Registers an alias for this module
 const path = require('path');
 const moduleAlias = require('module-alias');
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
   "dependencies": {
     "gl-constants": "^1.0.0",
     "gl-matrix": "^2.3.2",
+    "seer": "^0.0.4",
     "webgl-debug": "^1.0.2"
   },
   "devDependencies": {
--- a/src/debug/seer-integration.js
+++ b/src/debug/seer-integration.js
@@ -1,0 +1,45 @@
+import seer from 'seer';
+
+const models = {};
+
+/**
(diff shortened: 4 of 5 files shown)


src/core/model.js
```diff
@@ -241,6 +246,9 @@ export default class Model extends Object3D {
    */
   /* eslint-disable max-statements */
   render(uniforms = {}) {
+    if (window.__SEER_INITIALIZED__) {
+      addModel(this);
```
ibgreen: Why add it here instead of constructor? A comment to explain?

src/core/model.js
```diff
@@ -136,6 +140,7 @@ export default class Model extends Object3D {
   /* eslint-enable complexity */
 
   destroy() {
+    removeModel(this.id);
```
ibgreen: As discussed this method is rarely (if ever) called today. But can't really see other ways of doing it. 

Maybe this is a motivation to improve deck.gl in this regard.

src/core/model.js
```diff
@@ -376,9 +384,21 @@ export default class Model extends Object3D {
...
     }
+
+    if (window.__SEER_INITIALIZED__) {
+      const uniformObject = getUniformsObject(
+        this.program,
+        Object.assign({}, this.uniforms, uniforms)
+      );
+
+      if (uniformObject) {
+        seer.indexedListItem('luma.gl', this.id, uniformObject);
```
ibgreen: Something about this method name - it is not very intuitive. Consider something like `addInfoToItem` ?

src/core/models.js
```diff
@@ -0,0 +1,33 @@
+import seer from 'seer';
```
ibgreen: Should the file name be `seer-integration.js` or similar?

src/core/models.js
```diff
@@ -0,0 +1,33 @@
...
+    return;
+  }
+  models[model.id] = model;
+};
+
+export const removeModel = id => {
+  delete models[id];
+};
+
+const recursiveSet = (obj, path, value) => {
```
ibgreen: Needs documentation, not clear what this is doing.

src/core/models.js
```diff
@@ -0,0 +1,33 @@
...
+  if (path.length > 1) {
+    recursiveSet(obj[path[0]], path.slice(1), value);
+  } else {
+    obj[path[0]] = value;
+  }
+};
+
+seer.listenFor('luma.gl', payload => {
```
ibgreen: Suggest being very generous with documentation here. Both header and inside method.

src/webgl/uniforms.js
```diff
@@ -130,6 +130,21 @@ function checkUniformValue(value) {
...
 
+/**
+ * Create the current uniform state object to be sent to the devtools
+ */
+export const getUniformsObject = (program, uniforms) => {
+  if (!program) {
+    return null;
+  }
+
+  return Object.keys(program._uniformSetters).reduce((out, uniformName) => {
```
ibgreen: The rest of the code uses for statements rather than reduce, would be nice to keep convention/style per repo.

test/node.js
```diff
@@ -1,6 +1,10 @@
 // Enables ES2015 import/export in Node.js
 require('reify');
 
+// Mock addEventListener on window
```
ibgreen: Explain that this is for seer

src/core/model.js
Apercu: Looking at the docs will make more sense

src/core/models.js
Apercu: Nope, seer is what it should be

src/core/models.js
Apercu: Sure, will do

test/node.js
Apercu: ok

src/core/model.js
ibgreen: > Looking at the docs will make more sense

I agree that this discussion should be had in the Seer repo. It's a reminder that Seer itself also needs to reviewed. We should review API and implementation before launching and update here if needed.

src/core/models.js
ibgreen: >>Should the file name be seer-integration.js or similar?
> Nope, seer is what it should be

Of course, was not suggesting we change the module name. 

I was actually referring to name of the new file that this PR creates in luma.gl: `src/core/models.js`. 

I feel it would make more sense as something like `src/debug/seer-integration.js`

Apercu: Oh I see, yes definitely

src/webgl/uniforms.js
Apercu: It's quite disgusting to use `for in` since you also have to check if it's not an inherited property, but fine

ibgreen: > It's quite disgusting to use for in since you also have to check if it's not an inherited property, but fine

You don't, that linter rule is disabled since it is never an issue in practice.

Also you can do a `for of` on `Object.keys` if that feels better. It is more the `reduce` that looks out of place.

src/core/model.js
Apercu: That was because in the constructor, it appears sometimes that id's are not yet properly initialized, or using temporary names like `Model-1`, but maybe I'm doing something wrong?

src/webgl/uniforms.js
Apercu: I replaced it by `for in` already, tell me if I need to use `for of` instead or if it looks fine

src/core/model.js
ibgreen: Hmm, maybe yes, maybe some models are overwriting it in their constructors rather than passing it in as a parameter.

If you make an ES6 Map you could key on objects instead of ids.

It wouldn't work in deck.gl but it would work here.

PR #154, opened 2017-04-28
Fix rendering. Buffer and Program updates

ibgreen:
* Mainly fixes related to Buffer target type.
* Fixed a small WebGL2 porting issue in Program. 
* Updated the instancing example to make sure we have working test case.


ibgreen:
Hotfix, merging. Will address any comments in second PR.

PR #155, opened 2017-04-28
Add coverage support

ibgreen:
Trying to copy the deck.gl setup to luma.gl.


heshan0131:
We should add coverage/ to .gitignore

ibgreen:
> We should add coverage/ to .gitignore

Yes good catch, done. At this point, I can run the `cover` script, but I seem to get an empty report...

PR #160, opened 2017-05-02
WebGL2 updates

ibgreen:
Adds WebGL2 features for v4


ibgreen:
@shaoningli - I could reproduce this error in luma.gl, should be fixed now.

src/core/model.js
```diff
@@ -239,15 +223,15 @@ export default class Model extends Object3D {
     return Object.assign({}, uniforms, cameraUniforms, viewUniforms);
   }
 
-  /*
-   * @param {Camera} opt.camera=
-   * @param {Camera} opt.viewMatrix=
-   */
-  /* eslint-disable max-statements */
-  render(uniforms = {}) {
+  draw({uniforms = {}, attributes = {}, settings = {}, samplers = {}}) {
```
shaojingli: Is the draw() function here for compatibility reasons? I see that it just calls Model.render() with all arguments passed over. 

Also, it seems that all deck.gl layers calls, Model.render().

src/webgl/framebuffer.js
```diff
@@ -45,90 +35,176 @@ export default class Framebuffer extends Resource {
...
       });
-      this.attachRenderbuffer({
-        attachment: GL.DEPTH_ATTACHMENT,
-        renderbuffer: depthBuffer
-      });
+      attachments[gl.DEPTH_ATTACHMENT] = depthBuffer;
 
-      if (this.opts.depthBuffer) {
-        // this.opts.depthBuffer.delete();
+      if (this.depthBuffer) {
```
shaojingli: Remove this?

src/core/model.js
ibgreen: > Is the draw() function here for compatibility reasons? I see that it just calls Model.render() with all arguments passed over.

Good question!

Actually, `draw` is new. The old function (`render`) only takes uniforms. As you can see, we want to be able to supply more parameters.  But everyone already uses `render` so if we change that function's signature, we will break a lot of code. 

So one way is simply to define a new function, `draw` and maintain both in parallel. But that's only one way. We could also do dynamic checking on the `render` arguments.

I will create an API-AUDIT doc and add bullets about API design issues that need to be settled and list this question here.

src/webgl/framebuffer.js
ibgreen: > // this.opts.depthBuffer.delete(); 
> Remove this?

The idea is that the autogenerated targets would be autodeleted when e.g. resizing the framebuffer. Just need a little more time to get the logic correct.

PR #163, opened 2017-05-05
Tree shaking

ibgreen:
* Set up a new "tree-shaking" example with minification & analysis configs for webpack and rollup. 
* Size reductions in core library based on iteration with both tools.
* Source maps are no longer inline, to reduce bundle size during debug.


ibgreen:
Experiments to mitigate https://github.com/uber/deck.gl/issues/621

src/experimental/shader-modules/index.js
```diff
@@ -1,16 +1,15 @@
...
+// import * as material from './material';
 
-registerShaderModules([
-  fp64,
-  fog,
-  picking,
-  // lighting,
-  material
-]);
+// TODO - This probably defeats tree-shaking
```
Apercu: Can we register the modules one by one in their own file so that only importing them will trigger the register?

src/utils/globals.js
```diff
@@ -1,6 +1,6 @@
 /* global window, global, document */
-module.exports = {
```
Apercu: why the code duplication between `globals-internal.js` and `globals.js`?

ibgreen: > why the code duplication between globals-internal.js and globals.js?

Let me double check - something here was causing problems with rollup I think. I'll add comments if I need to keep this change.

src/experimental/shader-modules/index.js
ibgreen: > Can we register the modules one by one in their own file so that only importing them will trigger the register?

The problem is that to support tree shaking we need double dist folders - one for non-tree-shaking aware bundlers, and one with imports/export left untranspiled for tree-shaking bundlers.

In this case, keeping multiple top level import files is complicated, we can't just put a file in the root that imports './dist/...' since dist might be different depending on bundler.

A good solution will be to just export a function registerPackages that the app needs to call. As long as we don't register them globally we are good.

Apercu: I know how tree shaking works, but I feel like giving this responsibility to the user is a bit of overhead we should be able remove.

By simply registering each module one by one in their respective file, we would prevent that. Maybe just rename `registerShaderModules` into `registerShaderModule` and only pass only one module instead of a list.

Or is there something I'm missing?

ibgreen: > Or is there something I'm missing?

Well, index.js imports all files. Any statements on the top or "global" level will be executed during that import.

Tree shaking only comes into play after everything has been imported (in the case of webpack it doesn't happen until minification). The bundler starts looking for things that are not being imported/used. But then it is already too late. The bundler can no longer tree shake these things since they have indeed been used by the program.

In fact, this type of side effects is an huge unsolved problem for the tree-shakers and defeats much of their value for code that hasn't been carefully tuned.

As an example, even just setting a static member on a class (or the mere transpilation of the class which generates assignments to the prototype) can generate side effects that disable tree-shaking. 

That's why I added the babel plugin that injects /* PURE */ comments but also that prevented me from removing comments from the code, so I will need to find a configurable comment remover, sigh...

ibgreen:
Did a cherry-pick rebase due to extensive changes. New branch so will close and open again.

PR #164, opened 2017-05-05
Remove assert checks in draw function when non-debug context is used

shaojingli:
Cache glGetDebugInfo queries results

ATTN: this diff is targeting the 3.0-release branch

### DIFF
--- a/src/webgl/context.js
+++ b/src/webgl/context.js
@@ -195,18 +195,24 @@ export function glContextWithState(gl, {scissorTest, framebuffer}, func) {
  * @param {WebGLRenderingContext} gl - context
  * @return {Object} - 'vendor' and 'renderer' string fields.
  */
+let lastGLContext = null;
+let lastDebugInfo = null;
 export function glGetDebugInfo(gl) {
-  const info = gl.getExtension('WEBGL_debug_renderer_info');
-  // We can't determine if 'WEBGL_debug_renderer_info' is supported by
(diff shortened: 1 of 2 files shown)


src/webgl/context.js
```diff
@@ -195,18 +195,24 @@ export function glContextWithState(gl, {scissorTest, framebuffer}, func) {
  * @param {WebGLRenderingContext} gl - context
  * @return {Object} - 'vendor' and 'renderer' string fields.
  */
+let lastGLContext = null;
```
ibgreen: Nit: Technically the JSDoc contextually belong to the next definition `lastGLContext`. We are not generating docs so maybe not a big deal.

src/webgl/draw.js
```diff
@@ -23,12 +23,12 @@ export function draw(gl, {
 
   drawMode = glGet(drawMode);
   indexType = glGet(indexType);
-
-  assertDrawMode(drawMode, 'in draw');
-  if (isIndexed) {
-    assertIndexType(indexType, 'in draw');
+  if (gl.debug) {
```
ibgreen: +1

Actually, I think we should just delete these, no need to keep them in an if statement. This type of checks just make the executable bigger and any errors will already be caught in debug mode by the `WebGLDebug` package.

ibgreen:
@shaojingli - remove the if statement and merge?

ibgreen:
@shaojingli merged - Does this need to be cherry picked into master?

PR #166, opened 2017-05-07
Better explanation of examples setup. 

ibgreen:
Remove yarn.lock files in examples.


package.json
```diff
@@ -30,17 +30,12 @@
     "headless.js"
   ],
   "scripts": {
-    "start-webpack": "(cd examples/main && yarn && npm run start-local)",
-    "start": "npm run build-watch && budo dist/bundle.js:dist/lumagl.js --live --open --port 3000 --watch-glob '**/*.{html,css,js,glsl}' ",
-    "build-watch": "babel src -d dist --source-maps inline --copy-files --watch & babel examples -d dist --source-maps inline --copy-files --watch &",
+    "start": "cat < docs/get-started/examples.md",
```
ericsoco: What's going on here?

ericsoco: Don't we want `npm start` to run examples in a browser?

Apercu: Since we only have standalone examples, `cat`'ing the get started to run these examples makes sense to me.

ericsoco: Ah, I forgot there's no "layer browser" equivalent in luma.
And actually viewing examples.md, it makes even more sense. 👌

ibgreen: > Ah, I forgot there's no "layer browser" equivalent in luma.

I am starting to think this might be the best setup for all our repos (i.e also deck.gl/react-map-gl).

* We see from issues being opened that many users are confused by the double install setup in the current `start` script (I think we knew this). 
* Also users don't seem to easily figure out how to work with the standalone examples, including how to build against local source.
So the idea was, why not just tell them to go into the examples and install and run there.

ericsoco: I think that's a good idea. One more bullet point:
- The layer browser / single-entry-point-for-all-examples requires maintenance

My understanding is that the layer browser also exists to ensure compatibility between layers, so if we remove it as an example entry point we'll still need to keep it around for that use.

ibgreen: > My understanding is that the layer browser also exists to ensure compatibility between layers, so if we remove it as an example entry point we'll still need to keep it around for that use.

Good point. Wasn't planning on removing any examples (least of all the layer browser). Just changing the top level `npm start` command from running the layer browser to showing the docs to avoid the technical complications with dual installs.

This seems more like a process issue, e.g.:
* At the moment we mainly test that all the examples work before we release but since an increasing amount of users are working on master it is becoming important to keep it in good condition.
* We could publish a test guideline for contributions and ask contributors do a manual test run using the layer browser on any PRs they submit. 
* Or update our internal guidelines to require that we run the layer browser before merging non-trivial PRs.

ericsoco: Probably good to do both 2 & 3 -- PR authors test and reviewers test.
To get back on-track re: this PR tho, agreed (w/ you and @Apercu) that just showing the docs suffices and will save trouble with a complex `npm start` entry point.

However, we now need to make this clear in the root `README` as well. Should probably have been part of this PR in retrospect >.<

PR #172, opened 2017-05-15
Library fixes to make examples run under new API

ibgreen:


src/webgl/framebuffer.js
```diff
@@ -8,6 +8,24 @@ import {getTypedArrayFromGLType, getGLTypeFromTypedArray} from '../utils/typed-a
 import {log} from '../utils';
 import assert from 'assert';
 
+// Local constants - will collapse during minification
+const GL_FRAMEBUFFER = 0x8D40;
```
shaojingli: Why we decide to have these GL constants defined here instead of extracting them from our "GL" struct?

src/webgl/texture.js
```diff
@@ -344,47 +344,77 @@ export default class Texture extends Resource {
...
       format, type, dataFormat, compressed, data, width, height}));
 
-    // Support ndarrays
-    if (data && data.data) {
-      const ndarray = data;
-      data = ndarray.data;
-      width = ndarray.shape[0];
-      height = ndarray.shape[1];
+    const {gl} = this;
+    gl.bindTexture(this.target, this.handle);
```
shaojingli: We are using `this.target` in some occasions and `target` in some others

src/webgl/framebuffer.js
ibgreen: > Why we decide to have these GL constants defined here instead of extracting them from our "GL" struct?

The comment is a little unclear, sorry. It's related to making the libraries smaller. The GL object contains all flags and is pretty big (15Kb+). If we have local constants here, they collapse to small numbers during minification/uglification. Then the app can decide whether to import GL or not. Small apps won't have to take the hit.

shaojingli: sorry for the confusion but you did answer my question. So it's for reduction of the library size then.

src/webgl/texture.js
ibgreen: Yes we should be consistent here. There are four subclasses of texture and they need different targets, that is what we need to accommodate. Let's review after landing as you suggest.

PR #173, opened 2017-05-16
Fixes to examples

ibgreen:


ibgreen:
> LGTM. Could we create an issue on github tracking which examples are working and which are under construction?

Will do, in the mean time run the demo and check manually. Many need to be updated to the new math library, some need to move to animation loop structure. Plus there are some changes in texture and framebuffer apis that cause some breakages.

PR #176, opened 2017-05-22
Remove null parameters given to Float32Array constructor

### LINKED ISSUE #175: deck.gl layer browser example broken in Safari/FF
On master, Safari and FF first throw an error on [this `Float32Array` initialization](https://github.com/uber/luma.gl/blob/master/src/webgl/context-state.js#L109). Getting past that, `this.state.model` is `null` in `ScatterplotLayer` in deck.gl.

I'm not clear on how much effort we're putting into keeping master unbroken across repos, so feel free to file this away for later. However, it's blocking investigation into https://github.com/uber/deck.gl/issues/652.

shaojingli:
@ericsoco @ibgreen

Fix #175

### DIFF
--- a/src/webgl/context-state.js
+++ b/src/webgl/context-state.js
@@ -106,7 +106,7 @@ const GL_STATE = {
 
   depthRange: {
     type: new Float32Array(2),
-    value: new Float32Array(2), // TBD
+    value: new Float32Array([0, 1]), // TBD
     object: ['min', 'max'],
     params: GL.DEPTH_RANGE,
     setter: (gl, value) => gl.depthRange(...value)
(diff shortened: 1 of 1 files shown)


src/webgl/context-state.js
```diff
@@ -106,7 +106,7 @@ const GL_STATE = {
 
   depthRange: {
     type: new Float32Array(2),
-    value: new Float32Array(null, null), // TBD
+    value: new Float32Array(2), // TBD
```
ibgreen: It is supposed to be the default value for `depthRange` for a new context, so that we don't have to query it. 

I think it should be `new FloatArray([0, 1])`

PR #178, opened 2017-05-23
UniformBufferLayout class

ibgreen:
Basic support for `std140` uniform buffer layouts.

### DIFF
--- a/src/webgl/uniform-buffer-layout.js
+++ b/src/webgl/uniform-buffer-layout.js
@@ -1,0 +1,165 @@
+/* eslint-disable camelcase */
+import assert from 'assert';
+
+const ERR_ARGUMENT = 'UniformBufferLayout illegal argument';
+
+// Local constants - these will "collapse" during minification
+const GL_INT = 0x1404;
+const GL_UNSIGNED_INT = 0x1405;
+const GL_FLOAT = 0x1406;
+
+const GL_FLOAT_VEC2 = 0x8B50;
+const GL_FLOAT_VEC3 = 0x8B51;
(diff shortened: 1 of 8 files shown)


src/webgl/uniform-buffer-layout.js
```diff
@@ -0,0 +1,165 @@
...
+  [GL_FLOAT_MAT2x4]: [GL_FLOAT, 8], // 8
+
+  [GL_FLOAT_MAT3]: [GL_FLOAT, 12], // 9
+  [GL_FLOAT_MAT3x2]: [GL_FLOAT, 12], // 6
+  [GL_FLOAT_MAT3x4]: [GL_FLOAT, 12], // 12
+
+  [GL_FLOAT_MAT4]: [GL_FLOAT, 16], // 16
+  [GL_FLOAT_MAT4x2]: [GL_FLOAT, 16], // 8
+  [GL_FLOAT_MAT4x3]: [GL_FLOAT, 16] // 12
+};
```
shaojingli: Just a note, so this layout class will not support array types as uniform. It's okay for now but will be limiting sometimes, esp. considering that GLSL 3.0 added support of dynamic indexing of vector and matrix arrays.

ibgreen: > this layout class will not support array types as uniform

Yes good observation. The standard (non-buffer) uniform support in luma.gl has similar limitation.

I feel that it is partly a question of putting together a set of good test cases - there are a lot of data type + array combinations, easy for some cases to break. This would benefit from a systematic test rig.

Will add to a tracker task.

PR #180, opened 2017-05-23
Add probe package - will enable cross session logging settings

ibgreen:


1chandu:
@ibgreen , this seems very old, is this still relevant or should we close it?

PR #187, opened 2017-05-31
add disable rasterizer, drawMode to primitiveMode mapping, and a weba…

gnavvy:
- added disable rasterizer to program;
- added drawMode to primitiveMode mapping;
- a minor webpack.config bug fix

### DIFF
--- a/examples/core/transform-feedback/package.json
+++ b/examples/core/transform-feedback/package.json
@@ -7,11 +7,11 @@
     "start-local": "webpack-dev-server --env.local --progress --hot --open"
   },
   "dependencies": {
-    "luma.gl": "^4.0.0"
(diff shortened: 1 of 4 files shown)


examples/core/transform-feedback/package.json
```diff
@@ -7,11 +7,11 @@
     "start-local": "webpack-dev-server --env.local --progress --hot --open"
   },
   "dependencies": {
-    "luma.gl": "^4.0.0"
+    "luma.gl": ">=4.0.0-alpha.6"
```
shaojingli: Good change! We should do this for every examples in luma.gl and deck.gl, and the frameworks themselves. This has been mentioned by @heshan0131

PR #188, opened 2017-06-01
Fix for lessons 1,2 and 3

1chandu:


shaojingli:
Congratulations @1chandu

PR #189, opened 2017-06-01
Fix the texture storage mode settings

shaojingli:
Fix the texture storage mode settings after recent refactor
This is also the root cause of https://github.com/uber/deck.gl/issues/665

@ibgreen @Pessimistress @1chandu


src/webgl/context.js
```diff
@@ -218,21 +219,17 @@ export function withParameters(gl, {scissorTest, framebuffer, nocatch = true}, f
     if (framebuffer) {
       // TODO - was there any previously set frame buffer?
       // TODO - delegate "unbind" to Framebuffer object?
-      gl.bindFramebuffer(gl.FRAMEBUFFER, null);
+      framebuffer.unbind();
     }
   }
 
   let value;
-  if (nocatch) {
```
shaojingli: Remove nocatch temporarily as withState() function does not support nocatch path yet

src/webgl/context-state.js
```diff
@@ -535,26 +536,15 @@ export function setParameter(gl, key, value) {
 export function withState(gl, params, func) {
   // assertWebGLContext(gl);
   const state = getState(gl);
-
```
shaojingli: Remove framebuffer settings from withState as framebuffer state currently is not tracked by our state manager. The ideal solution would be having GL state manager track it but remove *for now* for consistency.

src/webgl/context-state.js
```diff
@@ -317,25 +318,25 @@ const GL_STATE = {
   [GL.UNPACK_ALIGNMENT]: {
     type: GLint,
     params: GL.UNPACK_ALIGNMENT,
-    setter: (gl, value) => gl.pixelStorei(value)
+    setter: (gl, value) => gl.pixelStorei(GL.UNPACK_ALIGNMENT, value)
```
ibgreen: Good fix! 

Gave me an idea, maybe the setter could get the flag as an argument, I'll look into that.

src/webgl/context-state.js
ibgreen: While I agree with your reasoning, I am a little worried that there are places within the code that might be affected.

I should probably just prioritize a fix in that does this the right way. Can you comment it out for now and add a TODO with my name?

src/webgl/context-state.js
```diff
@@ -535,26 +536,15 @@ export function setParameter(gl, key, value) {
...
-
-  const {frameBuffer} = params;
-  // TODO - was there any previously set frame buffer we need to remember?
-  if (frameBuffer) {
-    frameBuffer.bind();
-  }
-
   state.pushValues(gl, params);
-
+  let value;
```
ibgreen: Good fix!

src/webgl/context.js
ibgreen: Can you comment it out instead of deleting - I added it recently since all these catches can interfere with debugging so I want to disable it in places where I am reasonably confident that there will not be exceptions.

texture.diff
```diff
@@ -0,0 +1,966 @@
+diff --git a/src/webgl/texture.js b/src/webgl/texture.js
```
ibgreen: Can we delete this file?

src/webgl/context.js
shaojingli: Did what you said

src/webgl/context-state.js
```diff
@@ -536,25 +537,28 @@ export function withState(gl, params, func) {
   // assertWebGLContext(gl);
   const state = getState(gl);
 
-  const {frameBuffer} = params;
-  // TODO - was there any previously set frame buffer we need to remember?
-  if (frameBuffer) {
-    frameBuffer.bind();
-  }
+  // TODO (@ibgreen): Make GL state manager tracking framebuffer state and
```
shaojingli: Also added this piece of code back, commented out.

src/webgl/context-state.js
shaojingli: I missed the first case..... :( since no one is using it...I guess I'll just commit but not hurry to publish it

PR #194, opened 2017-06-02
Fix lesson-7 example.

1chandu:
- Lesson 6: Add WebGL example link.
- Lesson 7: 
   Remove unnecessary texture code (probably copied over from lesson-6)
   Remove dependency on Scene object , use shaders for lighting.

### DIFF
--- a/examples/lessons/06/app.js
+++ b/examples/lessons/06/app.js
@@ -45,35 +45,7 @@ const animationLoop = new AnimationLoop({
   // .context(() => createGLContext({canvas: 'lesson05-canvas'}))
   onInitialize: ({canvas, gl}) => {
     addControls();
-
-    addEvents(canvas, {
-      onKeyDown(e) {
-        switch (e.key) {
-        case 'f':
-          filter = (filter + 1) % 3;
-          break;
-        case 'up':
-          xSpeed -= 0.02;
-          break;
-        case 'down':
-          xSpeed += 0.02;
-          break;
-        case 'left':
-          ySpeed -= 0.02;
-          break;
-        case 'right':
-          ySpeed += 0.02;
-          break;
-        // andle page up/down
-        default:
-          if (e.code === 33) {
-            z -= 0.05;
-          } else if (e.code === 34) {
-            z += 0.05;
-          }
-        }
-      }
-    });
+    addKeyboardHandler(canvas);
 
     // context.set(gl, {
     //   clearColor: [0, 0, 0, 1],
@@ -144,9 +116,8 @@ const animationLoop = new AnimationLoop({
 
     // draw Cube
 
-    // update element matrix
+    // update element matrix to rotate cube on its center
     cube
-      .setPosition([0, 0, z])
       .setRotation([xRot, yRot, 0])
       .updateMatrix();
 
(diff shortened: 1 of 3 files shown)


examples/lessons/07/app.js
```diff
@@ -3,205 +3,238 @@
...
-  <br>
-  refraction
-  <input class="valign" id="refraction"
-    type="range" min="0.0" max="1.0" value="1.0" step="0.01">
-  <br>
+    The classic WebGL Lessons in luma.gl
     `;
   }
 }
+
```
shaojingli: Is this piece of code kept for adding functionality in the future? If yes, we should add a comment indicating that start with 'TODO' . Otherwise, we should remove it.

examples/lessons/07/app.js
```diff
@@ -3,205 +3,238 @@
...
+  gl_Position = uPMatrix * uMVMatrix * vec4(positions, 1.0);
+  vTextureCoord = texCoords;
+
+  if (!uUseLighting) {
+      vLightWeighting = vec3(1.0, 1.0, 1.0);
+  } else {
+      // Perform lighting in world space
+      // we should use 'transpose(inverse(mat3(uMVMatrix)))', but
+      // 'inverse' matrix operation not supported in GLSL 1.0
+      mat3 normalMatrix = mat3(uMVMatrix);
```
ibgreen: Does this take the top left submatrix? Maybe add comment...

examples/lessons/06/app.js
```diff
@@ -44,6 +44,7 @@ var filters = ['nearest', 'linear', 'mipmap'];
 const animationLoop = new AnimationLoop({
   // .context(() => createGLContext({canvas: 'lesson05-canvas'}))
   onInitialize: ({canvas, gl}) => {
+    addControls();
 
     addEvents(canvas, {
```
ibgreen: Is this supposed to be here, or only in lesson 7?

examples/lessons/07/app.js
```diff
@@ -3,205 +3,238 @@
...
+var xSpeed = 0.01;
+var yRot = 0;
+var ySpeed = 0.0;
+var z = -5.0;
 
+const animationLoop = new AnimationLoop({
+  onInitialize: ({canvas, gl}) => {
+    addControls();
+
+    addEvents(canvas, {
```
ibgreen: Break out into a function `addKeyboardHandler` or similar, like `addControls`?

examples/lessons/07/app.js
shaojingli: Yeah. We could live with normalMatrix being the same as uMVMatrix as long as the world-to-view transformation doesn't involve non-uniform scaling among different axes.

examples/lessons/06/app.js
1chandu: We need this here, lesson 6 is not showing link to the WebGL example, and this fixes it.

ibgreen: Oh I see, yes makes sense. But how about `addEvents`?

1chandu: Moved addEvents to seperate method 'addKeyboardHandler()' , same as lesson7, if that is what you are referring to.

PR #196, opened 2017-06-03
Fix lesson-08

1chandu:
-Remove dependency on Scene object and add custom VS and PS.
-Fix cube rotation to match with WebGL example.

### DIFF
--- a/examples/lessons/08/app.js
+++ b/examples/lessons/08/app.js
@@ -1,33 +1,83 @@
 /* eslint-disable no-var */
 /* eslint-disable max-statements, array-bracket-spacing, no-multi-spaces */
 /* global window, document */
-import {createGLContext, loadTextures, Cube, Program, addEvents, Fx} from 'luma.gl';
+import {GL, AnimationLoop, Cube, Matrix4, Texture2D,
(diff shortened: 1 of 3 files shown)


examples/lessons/08/app.js
```diff
@@ -1,181 +1,230 @@
...
-    loadTextures(gl, {
+  onInitialize: ({canvas, gl}) => {
+    addControls();
+    addKeyboardHandler(canvas);
+
+    gl.clearColor(0, 0, 0, 1);
+    gl.clearDepth(1);
+    gl.enable(GL.DEPTH_TEST);
+    gl.depthFunc(GL.LEQUAL);
+    gl.blendFunc(gl.SRC_ALPHA, gl.ONE);
```
1chandu: As I enabled blending in this lesson, now when I switch to other lessons, blending is enabled there too, does all lessons use same GL context?

ibgreen: > do all lessons use same GL context?

Yes they all share a context.

> As I enabled blending in this lesson, now when I switch to other lessons, blending is enabled there too, does

We have a new state management feature in luma v4, to enable local settings, let's try to use it here.

PR #201, opened 2017-06-07
Lesson09: fixes

1chandu:
- Fix lesson 09 and enable it.

### DIFF
--- a/examples/lessons/09/app.js
+++ b/examples/lessons/09/app.js
@@ -1,82 +1,100 @@
 /* eslint-disable no-var, max-statements */
 /* eslint-disable array-bracket-spacing, no-multi-spaces */
 /* global document */
-import {AnimationLoop, loadTextures, PerspectiveCamera, Group, addEvents} from 'luma.gl';
+import {GL, AnimationLoop, loadTextures, addEvents, Matrix4} from 'luma.gl';
 import {Star} from './star';
-
-var tStar;
-var twinkle = $id('twinkle');
 
 var zoom = -15;
 var tilt = 90;
 
 const animationLoop = new AnimationLoop({
-  // var canvas = document.getElementById('lesson09-canvas');
-  // canvas.width = canvas.clientWidth;
-  // canvas.height = canvas.clientHeight;
-  // var gl = createGLContext({canvas});
+  onInitialize: ({canvas, gl}) => {
+    addControls();
+    addKeyboardHandler(canvas);
 
-  onInitialize: ({gl, canvas}) => {
+    gl.clearColor(0, 0, 0, 1);
+    gl.clearDepth(1);
     gl.blendFunc(gl.SRC_ALPHA, gl.ONE);
     gl.enable(gl.BLEND);
+//    gl.disable(gl.DEPTH_TEST);
 
-    var scene = new Group(gl);
-
-    addEvents(canvas, {
-      onKeyDown(e) {
-        switch (e.key) {
-        case 'up':
-          tilt -= 1.5;
-          break;
-        case 'down':
-          tilt += 1.5;
-          break;
-        // handle page up/down
-        default:
-          if (e.code === 33) {
-            zoom -= 0.1;
-          } else if (e.code === 34) {
-            zoom += 0.1;
-          }
-        }
-      }
-    });
-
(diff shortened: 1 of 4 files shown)


examples/lessons/09/app.js
```diff
@@ -1,82 +1,100 @@
...
     })
     .then(textures => {
-      tStar = textures[0];
-
-      // Load all world objects
+      var stars = [];
       var numStars = 50;
       for (var i = 0; i < numStars; i++) {
-        scene.add(new Star(i / numStars * 5.0, i / numStars));
+        stars.push(new Star({
```
ibgreen: Just realized that this is creating 50 copies of the program and shaders. Do you see any startup delay? 
If so the shader cache might be a worth while introduction here.

It might also be worth deleting these in `onFinalize`

examples/lessons/09/app.js
```diff
@@ -1,82 +1,100 @@
...
+    const cameraZ = Math.sin(radTilt) * zoom;
 
-    scene.children.forEach(function(star) {
-      star.animate();
-    });
+    gl.clear(GL.COLOR_BUFFER_BIT | GL.DEPTH_BUFFER_BIT);
+
+    for (var i in stars) {
+      var uMVMatrix = Matrix4
+        .lookAt({eye: [0, cameraY, cameraZ]})
```
ibgreen: I am sure you are just following the example but it seems a bit weird to combine the view and model matrices here, doesn't it.

examples/lessons/09/app.js
```diff
@@ -1,82 +1,100 @@
...
+    The classic WebGL Lessons in luma.gl
+    `;
+  }
+}
+
+function addKeyboardHandler(canvas) {
+
+  addEvents(canvas, {
+    onKeyDown(e) {
+      switch (e.key) {
```
ibgreen: I think it would be nice to briefly describe the key bindings in the text.This is probably overkill but could give a nice look:
screen shot 2017-06-06 at 5 24 22 pm

```
        <table>
          <tbody>
            <tr>
              <td><div className="key">&#x2192;</div></td>
              <td>advance one frame (0.1s)</td>
            </tr>
            <tr>
              <td><div className="key">&#x2190;</div></td>
              <td>back up one frame (0.1s)</td>
            </tr>
            </tr>
          </tbody>
        </table>
```

With style as
```
 table {
    /* Give key table a little tighter row spacing */
    td {
      padding-top: 2px;
      padding-bottom: 2px;
      padding-left: 0px;
    }
    /* "keyboard key" style */
    .key {
        display: inline;
        display: inline-block;
        min-width: 1em;
        padding: .5em .6em .4em .6em;
        margin: 2px;
        font-weight: normal;
        font-size: 11px;
        text-align: center;
        text-decoration: none;
        line-height: 0.8rem;
        -moz-border-radius: .3em;
        -webkit-border-radius: .3em;
        border-radius: .3em;
        border: none;
        cursor: default;
        -moz-user-select: none;
        -webkit-user-select: none;
        user-select: none;
        background: #ffffff;
        border: 1px solid #ccc;
    }
  }
```

examples/lessons/09/app.js
1chandu: I don't see any noticeable delay during startup, but isn't the caching taken cared by either WebGL Framework or Graphics Drivers layers? Have we seen this problem before.

examples/lessons/09/app.js
1chandu: this code is building ModelView matrix, which is a uniform, hence including model and view matrices, are you suggesting to have two different uniforms?

examples/lessons/09/app.js
shaojingli: Yes. It's cached on the system level.

ibgreen:
```
[luma.gl](master)$npm run bench

> luma.gl@4.0.0-alpha.10 bench /Users/ib/Documents/luma.gl
> node test/bench/node.js

Starting bench...
Program from shader source x 510 ops/sec ±1.17% (84 runs sampled)
Program from cached shaders x 1,268 ops/sec ±1.00% (87 runs sampled)
Fastest is Program from cached shaders
```

PR #202, opened 2017-06-07
Fix blending in lessons

1chandu:
All lessons are sharing same canvas hence same GL context. When switching from a lesson that enables blending to another lesson that doesn't causes problem. Disable blending on initialize for all lessons that don't need blending.


shaojingli:
For this, I also would like to see luma.gl state manager being used but I understand that we should have this quick fix to test if each example is being fixed properly for now. 

@1chandu @ibgreen I'll create an issue on github to track the task of adopting luma.gl state manager

PR #204, opened 2017-06-09
Context state management fixes.

1chandu:
Context state management changes:
- 	Add 'resetContext' to set all gl states to default values.
- 	Call either glEnable or glDisable based on the value to be set.
- 	Fix default values for several gl states.


test/webgl/context-state.spec.js
```diff
@@ -65,3 +66,20 @@ test('WebGLState#withState', t => {
   t.end();
 });
 
+test('WebGLState#resetContext', t => {
+  const {gl} = fixture;
+
+  setParameter(gl, 'clearColor', [0, 1, 0, 1]);
```
ibgreen: For another PR: Would be great to add a complete list of alternate values here so we could test all settings.

PR #206, opened 2017-06-09
Reorganize state(parameter) code

ibgreen:
@1chandu Inspired by your work yesterday, I did a pass on the state/parameter handling reorganizing the code to provide a better base for further work.


src/webgl/context-state.js
```diff
@@ -1,456 +1,32 @@
...
-import GL from './gl-constants';
 import assert from 'assert';
 
 // WebGL specification 'types'
-const GLenum = 'GLenum';
-const GLfloat = 'GLfloat';
-const GLint = 'GLint';
-const GLuint = 'GLint';
-const GLboolean = 'GLboolean';
+import GL_PARAMETERS from './api/parameters';
```
shaojingli: 👍 on breaking it out to a separate file

docs/api-reference/webgl/context-state.md
```diff
@@ -2,7 +2,13 @@
 
 luma.gl provides WebGL state management which enabling a 'stateless' and 'conflict-free' WebGL programming model. In the stateless model, settings can be passed as parameters to rendering commands rather than being set directly on the global state. And in the conflict-free model, even when setting global state, one part of the code does not need to worry about whether other parts are changing the global state.
 
-Note that to fully support the conflict-free model and detect changes done e.g. in other WebGL libraries, luma.gl needs to hook into the WebGL context to track state changes.
+The following functions are provided:
+* `getParameter` - Returns the value(s) of a GL context parameter
```
shaojingli: using the ending 's' to distinguish APIs is usually very confusing and error-prone...

src/webgl/context-state.js
```diff
@@ -511,50 +91,67 @@ function getState(gl, {copyState = false} = {}) {
...
+// Get the parameter value(s) from the context
+// Might return an array of multiple values.
+// The return value will be acceptable input to setParameter
 export function getParameter(gl, key) {
-  const parameterDefinition = GL_STATE[key];
-  if (!parameterDefinition) {
-    throw new Error(`Unknown GL state parameter ${key}`);
-  }
-  // Get the parameter value(s) from the context
+  // Get the parameter definition for the key
```
shaojingli: So the getParameter() function still queries the WebGLRenderingContext and is slow? Should we consider storing current values of parameters so that query operations do land on WebGLRenderingContext? We could store it in GL_PARAMETERS and add a new `currentValue` prop for each entry.

docs/api-reference/webgl/context-state.md
ibgreen: > using the ending 's' to distinguish APIs is usually very confusing and error-prone...

I'm open to better suggestions!

I do see a need for asking for single values. But also want to offer the ability to get multiple values. All the luma.gl objects that derive from `Resource` follow the same pattern so if we change we could do it across the board.

src/webgl/context-state.js
ibgreen: Yes caching is only partly implemented. It would require some additional work and test cases to make it into luma v4.

Hoping to collaborate with @1chandu on this.

docs/api-reference/webgl/context-state.md
shaojingli: Okay... if you feel that an API for updating a single value is useful, then that's okay. I was thinking about just having one version of the APIs that can take both single or multiple values

PR #207, opened 2017-06-09
Misc fixes

1chandu:
Fixes I found while debugging lesson16.

- animation-loop.json: When merging the appContext (returned from initializer), use the correct merging order so existing props in current context are overwritten by whats in appContext.
- uniforms.js: Fix a typo.


1chandu:
Revert macbook.json changes.

PR #209, opened 2017-06-12
Re-write lesson 16.

1chandu:
- Re write lesson 16 to support "render to texture" and "point lighting".
- All of the features work similar to the corresponding webgl example except , specular highlights on moon and cube. Added 'TODO' comments that help debug this issue.

### DIFF
--- a/examples/lessons/16/app.js
+++ b/examples/lessons/16/app.js
@@ -1,7 +1,7 @@
 /* eslint-disable max-statements, indent, no-multi-spaces */
 import {GL, AnimationLoop, Cube, Sphere, Matrix4, Texture2D,
   loadTextures, Model, loadFiles, parseModel,
-  Program, Renderbuffer, Framebuffer, Geometry} from 'luma.gl';
+  Program, Renderbuffer, Framebuffer, Geometry, resetParameters} from 'luma.gl';
 
 // TODO: Remaining issues
 // 1. Specular highlights don not match with example:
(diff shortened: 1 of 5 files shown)


examples/lessons/16/app.js
```diff
@@ -46,254 +213,285 @@ export const animation = new AnimationLoop()
...
-    });
+function generateTextureForLaptopScreen(gl, tick, aspect, moon, cube, tSquare) {
+  moonAngle += moonAngleDelta;
+  cubeAngle += cubeAngleDelta;
+  gl.viewport(0, 0, FB_WIDTH, FB_HEIGHT);
+
+  if (!DISABLE_FB) {
+    rttFramebuffer.bind();
+  }
+
```
shaojingli: We should consider put gl.viewport() calls into Framebuffer.bind() to minimize the surprise

examples/lessons/16/app.js
```diff
@@ -1,36 +1,203 @@
...
+const FB_WIDTH = 512;
+const FB_HEIGHT = 512;
+
+var moonAngle = 0.0;
+var cubeAngle = Math.PI;
+var laptopAngle = 0.0;
+var moonAngleDelta = 0.01; // * Math.PI / 180.0;
+var cubeAngleDelta = 0.01; // * Math.PI / 180.0;
+var laptopAngleDelta = -0.005; // * Math.PI / 180.0;
+const moonCubeMaterial = {
```
ibgreen: Nit: We typically use CAPITAL_CASE for constants in JS.

examples/lessons/16/app.js
```diff
@@ -46,254 +213,285 @@ export const animation = new AnimationLoop()
...
     ])
     .then(function(results) {
-      var program = results[0];
-      var tMoon = results[1][0];
-      var tCrate = results[1][1];
-
-      var screenWidth = 512;
-      var screenHeight = 512;
-      var screenRatio = 1.66;
+      let resultIndex = 0;
```
ibgreen: How about using ES6 array destructuring?
```
let [macbookJson] = results[0];
const [tMoon, tCrate] = results[1];
```

examples/lessons/16/app.js
```diff
@@ -46,254 +213,285 @@ export const animation = new AnimationLoop()
...
+
+      // Fix attribute name to match with Shaders
+      macbookJSON = macbookJSON.replace('vertices', 'positions');
+      const program = new Program(gl, {vs: VERTEX_SHADER, fs: FRAGMENT_SHADER});
+      const macbook = parseModel(gl, {
+        id: 'macbook',
+        file: macbookJSON,
+        program,
+        uniforms: {
+          uMaterialAmbientColor: laptopMaterial.ambientColor,
```
ibgreen: almost feels like a `getMaterialUniforms` function would be appropriate?

examples/lessons/16/app.js
```diff
@@ -46,254 +213,285 @@ export const animation = new AnimationLoop()
...
-      uniforms: {
-        hasTexture1: true,
-        sampler1: tCrate,
-        shininess: 5,
-        enableSpecularHighlights: false,
-        materialAmbientColor: [1, 1, 1],
-        materialDiffuseColor: [1, 1, 1],
-        materialSpecularColor: [0, 0, 0],
-        materialEmissiveColor: [0, 0, 0]
+    rttFramebuffer = new Framebuffer(gl, {
```
ibgreen: We usually type out variable names.

examples/lessons/16/app.js
ibgreen: > We should consider put gl.viewport() calls into Framebuffer.bind() to minimize the surprise, tracked by our state manager

Good point. I think we have mostly used FBs of the same size as the screen.

I had a crazy idea - what if we made the `draw` call as a member of the `Framebuffer` class?

const screen = Framebuffer.getDefault();
screen.draw(...);
const fb = new Framebuffer(...);
fb.draw(...); // binds buffer and sets viewport temporarily...

There would be no need to bind framebuffers or `viewport` etc. The code would make it super clear what framebuffer we were working with.

We'd need to support backward compatibility of course...

examples/lessons/16/app.js
1chandu: addressed, thanks!

shaojingli:
@1chandu github says `src/webgl/index.js` is still conflicting

PR #210, opened 2017-06-13
State management : Unit tests, fixes and cleanup.

1chandu:
Unit tests, fixes and cleanup.

### DIFF
--- a/src/webgl-utils/parameters.js
+++ b/src/webgl-utils/parameters.js
@@ -97,7 +97,7 @@ export const GL_PARAMETER_SETTERS = {
   [GL.DEPTH_CLEAR_VALUE]: (gl, value) => gl.clearDepth(value),
   [GL.DEPTH_FUNC]: (gl, value) => gl.depthFunc(value),
   [GL.DEPTH_RANGE]: (gl, value) => gl.depthRange(...value),
-  [GL.DEPTH_WRITEMASK]: enable,
+  [GL.DEPTH_WRITEMASK]: (gl, value) => gl.depthMask(value),
   [GL.DITHER]: enable,
   [GL.FRAGMENT_SHADER_DERIVATIVE_HINT]: hint,
   [GL.FRONT_FACE]: (gl, value) => gl.frontFace(value),
--- a/src/webgl-utils/parameters.js
+++ b/src/webgl-utils/parameters.js
@@ -78,6 +78,8 @@ const enable = (gl, value, key) => value ? gl.enable(key) : gl.disable(key);
 const hint = (gl, value, key) => gl.hint(key, value);
 const pixelStorei = (gl, value, key) => gl.pixelStorei(key, value);
 
+// NOTE: When value type is a string, it will be handled by 'GL_PARAMETER_COMPOSITE_SETTERS'
+
 export const GL_PARAMETER_SETTERS = {
   [GL.BLEND]: enable,
   [GL.BLEND_COLOR]: (gl, value) => gl.blendColor(...value),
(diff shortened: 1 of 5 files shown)


src/webgl-utils/parameters.js
```diff
@@ -92,13 +94,13 @@ export const GL_PARAMETER_SETTERS = {
   [GL.CULL_FACE]: enable,
   [GL.CULL_FACE_MODE]: (gl, value) => gl.cullFace(value),
   [GL.DEPTH_TEST]: enable,
-  [GL.DEPTH_CLEAR_VALUE]: 1,
+  [GL.DEPTH_CLEAR_VALUE]: (gl, value) => gl.clearDepth(value),
   [GL.DEPTH_FUNC]: (gl, value) => gl.depthFunc(value),
-  [GL.DEPTH_RANGE]: (gl, value) => gl.depthRange(value),
-  [GL.DEPTH_WRITEMASK]: true,
+  [GL.DEPTH_RANGE]: (gl, value) => gl.depthRange(...value),
+  [GL.DEPTH_WRITEMASK]: enable,
```
ibgreen: I think this should call `gl.depthMask`. [link]( https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/depthMask).

src/webgl-utils/track-context.js
```diff
@@ -137,6 +124,12 @@ export const GL_STATE_SETTERS = {
     });
   },
 
+  viewport(setter, x, y, width, height) {
```
ibgreen: Should we order these alphabetically?

src/webgl-utils/track-context.js
```diff
@@ -336,3 +323,10 @@ export function popContextState(gl) {
   assert(gl.state);
   gl.state.pop();
 }
+
+export function setParameters(gl, params) {
+  assert(gl.state);
+  // Note: should we not pass gl.state.cache and let glSetParameters
```
ibgreen: The reason I've been avoiding this is that would create a circular dependency between params and track-context.

test/webgl-utils/parameters.spec.js
```diff
@@ -0,0 +1,168 @@
+import {createTestContext} from '../setup';
+import {GL} from 'luma.gl';
+import {glSetParameters, glGetParameter, glCopyParameters, GL_PARAMETER_DEFAULTS}
+  from '../../src/webgl-utils/parameters';
+import trackContext from '../../src/webgl-utils/track-context';
```
ibgreen: params does not depend on trackContext. It is a bit circular to import trackContext here.

test/webgl-utils/track-context.spec.js
```diff
@@ -1,78 +1,155 @@
...
-  scissorBox: [0, 0, 1024, 1024],
-
-  stencilTest: false,
-  stencilClearValue: 0,
-  stencilMask: [0xFFFFFFFF, 0xFFFFFFFF],
-  stencilFunc: [GL.ALWAYS, 0, 0xFFFFFFFF, GL.ALWAYS, 0, 0xFFFFFFFF],
-  stencilOp: [GL.KEEP, GL.KEEP, GL.KEEP, GL.KEEP, GL.KEEP, GL.KEEP],
-
-  viewport: [0, 0, 1024, 1024],
+const GL_PARAMETER_SETTINGS_ONE = {
```
ibgreen: One thing we are not testing here is "raw WebGL calls"...

e.g. `gl.viewport`, that it gets intercepted correctly.

test/webgl-utils/parameters.spec.js
```diff
@@ -0,0 +1,168 @@
...
+import test from 'tape-catch';
+
+// Settings test, don't reuse a context
+const fixture = {
+  gl: createTestContext({debug: true})
+};
+
+// eslint-disable-next-line
+const GL_PARAMETER_SETTINGS_ONE = {
```
ibgreen: We could copy these into a data file since we are using them in multiple test files.

src/webgl-utils/parameters.js
1chandu: Nice catch, will fix it.

PR #211, opened 2017-06-14
refactor(dev): use latest seer api

Apercu:


src/debug/seer-integration.js
```diff
@@ -10,6 +12,54 @@ export const addModel = model => {
     return;
   }
   models[model.id] = model;
+
+  seer.listItem('luma.gl', model.id);
+};
+
+const getType = obj => Object.prototype.toString.call(obj).slice(8, -1);
```
ibgreen: Too clever, let's add a comment explaining what this line is intended to do.

src/debug/seer-integration.js
```diff
@@ -10,6 +12,54 @@ export const addModel = model => {
...
+
+const getAttribute = (key, attr) => {
+  const isArray = getType(attr).includes('Array');
+  if (isArray) {
+    return Array.prototype.slice.call(attr);
```
ibgreen: Again explain why not `attr.slice()` to the rest of us who are not JS one liner fiends.

src/debug/seer-integration.js
```diff
@@ -10,6 +12,54 @@ export const addModel = model => {
...
+  }
+  return attr;
+};
+
+/**
+ * Trim functions properties and convert typed arrays into simple ones
+ * to be able to pass the payload through postMessage.
+ */
+const transformPayload = payload => Object.keys(payload)
+  .reduce((acc, attrKey) => {
```
ibgreen: Just a comment regarding this nested reduce on objects. As much as I would love to prefer reduce, this would have been much easier for me to grok if this was a nested for-in loop.

PR #212, opened 2017-06-15
State improvements

1chandu:
- Add new context management support for 'withParameters'.
- Add/update unit tests.


src/webgl-utils/track-context-state.js
```diff
@@ -241,22 +258,13 @@ class GLState {
   push(values = {}) {
     this.stateStack.push({});
     // this.setValues(values);
-    // glSetParameters(this.gl, oldValues, this.state);
+    // setParameters(this.gl, oldValues, this.state);
```
ibgreen: Remove commented code?

src/webgl-utils/parameter-access.js
```diff
@@ -253,10 +268,27 @@ export function glGetParameter(gl, key) {
...
+  for (const key of parameterKeys) {
+    state[key] = getParameter(gl, key);
   }
   return state;
 }
+
+export function getDefaultParameters(gl) {
+  // TODO - Query GL.VIEWPORT and GL.SCISSOR_BOX since these are dynamic
+  return Object.assign({}, GL_PARAMETER_DEFAULTS, {
+    // [GL.VIEWPORT]: gl.constructor.prototype.getParameter.call(gl, GL.VIEWPORT),
```
ibgreen: When commenting out code, good to add an additional comment explaining why we leave it in (rather than cut it).

PR #213, opened 2017-06-15
Fix  'blendFunc' setter, update examples to use setParameters

1chandu:


examples/core/cubemap/app.js
```diff
@@ -1,14 +1,20 @@
...
 
 const animationLoop = new AnimationLoop({
   onInitialize: ({gl, canvas}) => {
     addControls(canvas);
 
-    gl.clearColor(0, 0, 0, 1);
-    gl.clearDepth(1);
-    gl.enable(GL.DEPTH_TEST);
-    gl.depthFunc(GL.LEQUAL);
+    resetParameters(gl);
```
ibgreen: I think the `resetParameters` calls here could be confusing to a new user, trying to learn the API from the examples.

My understanding:
1. we don't currently share context, so `resetParameters` is not currently needed.
2. once we share context, we should call `resetParameters` in `AnimationLoop`

What do you think, should we just remove it?

src/webgl/context-state.js
```diff
@@ -17,7 +17,7 @@ export const LUMA_SETTERS = {
     gl.blendEquationSeparate(...args);
   },
   blendFunc: (gl, args) => {
-    args = isArray(args) && args.length === 3 ? [...args, ...args] : args;
+    args = isArray(args) && args.length === 2 ? [...args, ...args] : args;
```
ibgreen: Good catch!

test/webgl-utils/custom-parameter-settings.js
```diff
@@ -140,66 +140,61 @@ export const GL_PARAMETER_SETTINGS_TWO = {
   [GL.UNPACK_SKIP_IMAGES]: 32
 };
 
-// Note: These settings are same as GL_PARAMETER_SETTINGS_TWO
-// When changing change objects.
-export const GL_PARAMETER_SETTINGS_TWO_ENUM_FUNCTION = {
```
ibgreen: Since these test cases are related to the `webgl` layer, I think they should be in `tests/webgl` rather than `tests/webgl-utils`.

Essentially split this into two files.

test/webgl-utils/custom-parameter-settings.js
```diff
@@ -59,7 +59,7 @@ export const GL_PARAMETER_SETTINGS_ONE = {
   // WEBGL2 / EXTENSIONS
   // gl1: 'OES_standard_derivatives'
   [GL.FRAGMENT_SHADER_DERIVATIVE_HINT]: GL.FASTEST,
-  // RASTERIZER_DISCARD ...
```
ibgreen: Maybe name this file `sample-parameters.js` (and `sample-settings.js` if splitting it, see below)?

test/webgl/context-state.spec.js
```diff
@@ -2,8 +2,8 @@ import {GL} from 'luma.gl';
 import {getParameter, getParameters, setParameters, withParameters, resetParameters} from 'luma.gl';
 import {GL_PARAMETER_DEFAULTS as GL_PARAMETERS} from '../../src/webgl-utils/parameter-access';
 import {
-  GL_PARAMETER_SETTINGS_TWO,
-  GL_PARAMETER_SETTINGS_TWO_ENUM_FUNCTION
+  GL_PARAMETER_SETTINGS_ONE,
```
ibgreen: Maybe call these constants `PARAMETER_STYLE_SETTINGS` and `FUNCTION_STYLE_SETTINGS`?

We may also need to describe these concepts in the docs, so good to agree on a terminology.

ibgreen: Maybe call them `_SET1` and `_SET2` rather than `_ONE` etc?

test/webgl/context-state.spec.js
```diff
@@ -51,14 +57,57 @@ test('WebGLState#getParameter (WebGL2)', t => {
...
+
+  for (const key in GL_PARAMETER_SETTINGS_ONE) {
+    const value = getParameter(gl, key);
+    t.deepEqual(value, GL_PARAMETER_SETTINGS_ONE[key],
+      `got expected value ${stringifyTypedArray(value)} for key: ${key}`);
+  }
+  t.end();
+});
+
+test('WebGLState#setParameters function with \'seperate\' versions', t => {
```
ibgreen: Clarify a bit for the non-settings expert
" - changing parameters with `***Separate` WebGL setter functions"?

test/webgl/context-state.spec.js
```diff
@@ -51,14 +57,57 @@ test('WebGLState#getParameter (WebGL2)', t => {
   t.end();
 });
 
-test('WebGLState#setParameters', t => {
+test('WebGLState#setParameters with enum and functions keys', t => {
```
ibgreen: "...mixing parameter and function style keys"?

examples/core/cubemap/app.js
1chandu: @ibgreen @shaojingli, I dug little more into this, for each example, on animation loop's onInitialize, we try to create new gl context, but it is being served from the canvas object, and I think all examples have same canvas and hence are sharing same gl context. To verify, I enabled blending in `Instancing` , and switched to `Cube` example and blending is enabled there too.

So we do need, `resetParameters` on every example , I am going to add it to all remaining examples, I will leave comment explaining why we need it. Let me know what you guys think.

ibgreen: I am still confused by this. 

Does this mean that calling getContext on a canvas several times returns the **same** context (https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/getContext).

If this is true it would be surprising, and bad, but also  extremely useful, it would give us a sneaky way to get access to the contexts from other libraries.

And if this is true, I think there is even more reason we should call resetParameters in the `AnimationLoop`.

test/webgl/context-state.spec.js
1chandu: Agree, `PARAMETER` might be too generic, how about `ENUM_STYLE_SETTINGS` and `FUNCTION_STYLE_SETTINGS`.

examples/core/cubemap/app.js
shaojingli: @ibgreen yes.
Calling getContext with the same id ('2d' or 'webgl') on the same canvas would return the same context.
https://dev.w3.org/html5/spec-preview/the-canvas-element.html

test/webgl/context-state.spec.js
ibgreen: Sounds good.

examples/core/cubemap/app.js
ibgreen: @shaojingli @1chandu TIL!

In light of this, it might be interesting to give a `reset` parameter to `AnimationLoop.start()`. If that parameter is true we would call `resetParameters` on the context before calling `onInitailize`.

The discussion is mostly whether `reset` should be `true` or `false` by default. Probably what the user would expect, but open to opinions.

1chandu: My opinion is `reset` should be `true` by default, some one writing new example would like to start with default state, moreover one should not depend on existing state given user might click on examples in any order.

test/webgl-utils/custom-parameter-settings.js
1chandu: I moved into `test/webgl`, but still keeping these maps in same file, all three maps are used to set state and also retrieve and verify, didn't see a reason to split, happy to change if you think otherwise.

src/core/animation-loop.js
```diff
@@ -89,6 +90,12 @@ export default class AnimationLoop {
         if (!isWebGLContext(this.gl)) {
           throw new Error('AnimationLoop.onCreateContext - illegal context returned');
         }
+
+        // Unless reset is set to false, reset the context.
+        if ((contextParams.reset !== false)) {
+          resetParameters(this.gl);
```
1chandu: @ibgreen is this ok?

PR #215, opened 2017-06-17
Parameter handling polish

ibgreen:
Various improvements to parameter handling.

* Adds logging of changed values. 

* Fixes a few test cases, and makes set-parameters tests independent of track-context-state.


src/webgl-utils/set-parameters.js
```diff
@@ -244,22 +244,25 @@ export function setParameters(gl, values) {
...
 
-  for (const key in compositeSetters) {
-    assert(gl.state.cache);
-    // TODO - avoid calling composite setters if values have not changed.
-    const compositeSetter = GL_PARAMETER_COMPOSITE_SETTERS[key];
-    // Note - if `trackContextState` has been called,
-    // the setter will automatically update this.state.cache
-    compositeSetter(gl, mergedValues);
+    for (const key in compositeSetters) {
+      assert(gl.state.cache);
```
1chandu: Can be removed.

src/webgl-utils/set-parameters.js
```diff
@@ -244,22 +244,25 @@ export function setParameters(gl, values) {
   // This depends on `trackContextState`, which is technically a "circular" dependency.
   // But it is too inconvenient to always require a cache parameter here.
   // This is the ONLY external dependency in this module/
-  const mergedValues = Object.assign({}, gl.state.cache, values);
+  const cache = gl.state && gl.state.cache;
+  if (cache) {
```
1chandu: We should add log for else case.

src/webgl-utils/track-context-state.js
```diff
@@ -261,27 +255,39 @@ class GLState {
 
   pop() {
     assert(this.stateStack.length > 0);
-    const oldValues = this.stateStack.pop();
+    // Use the saved values in the state stack to restore parameters
+    const oldValues = this.stateStack[this.stateStack.length - 1];
```
1chandu: I see you found it, this is the reason for some of layers not working with my deck.gl diff.

PR #216, opened 2017-06-19
Shadertools module no longer experimental

ibgreen:
Goal is to make `deck.gl` import `assembleShaders` from luma.gl in v4.1, and make some of the luma.gl v4.0 examples use shader modules.


ibgreen:
78 files. looks like a monster diff, but it is mostly file moves.

1chandu:
Lot of changes, did a quick pass, LGTM.

PR #219, opened 2017-06-20
Add withParameters unit test + fix typo in GL constants

1chandu:
- Add a unit test for withParameters to test push/pop context with stack size > 1.
- Fix a typo in GL constants, FUNC_SUBSTRACT -> FUNC_SUBTRACT.

### DIFF
--- a/test/webgl/context-state.spec.js
+++ b/test/webgl/context-state.spec.js
@@ -151,3 +151,90 @@ test('WebGLState#withParameters', t => {
 
   t.end();
 });
+
+test('WebGLState#withParameters: recursive', t => {
+  const {gl} = fixture;
+
+  resetParameters(gl);
+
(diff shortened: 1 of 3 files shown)


test/webgl/context-state.spec.js
```diff
@@ -151,3 +151,90 @@ test('WebGLState#withParameters', t => {
```
ibgreen: should this file be called `track-context-state.spec.js`

1chandu: `setParameters` and `withParameters` are defined in context-state.js, so keeping the same name, context-state.spec.js, any other specific reason you want to change it?

ibgreen: I see now, you are correct, of course.

PR #225, opened 2017-06-23
Add NPOT workaround for texture2d object.

1chandu:
- Remove NPOT (non power of two) texture workaround from framebuffer object and add it to texure2d object. Make it applicable for only webgl1 context and when texture is NPOT.
- Fix assert failures in few lessons.
- Note: Merge is pending until deck.gl picking is verified.

### DIFF
--- a/src/utils/utils.js
+++ b/src/utils/utils.js
@@ -1,3 +1,5 @@
+import assert from 'assert';
+
 /**
  * Wraps the argument in an array if it is not one.
  * @param {object} a - The object to wrap.
@@ -75,4 +77,14 @@ function detach(elem) {
   }
 
   return ans;
+}
+
+/**
+ * Verifies if a given number is power of two or not.
+ * @param {object} n - The number to check.
+ * @return {Array} Returns true if the given number is power of 2, false otherwise.
+ **/
+export function isPowerOfTwo(n) {
+  assert((typeof n === 'number'), 'Input must be a number');
+  return n && ((n & (n - 1)) === 0);
 }
--- a/src/webgl/framebuffer.js
+++ b/src/webgl/framebuffer.js
@@ -433,7 +433,19 @@ export default class Framebuffer extends Resource {
         format: GL.RGBA,
         type: GL.UNSIGNED_BYTE,
         width,
-        height
+        height,
+        // Note: Mipmapping can be disabled by texture resource when we resize the texture
+        // to a non-power-of-two dimenstion (NPOT texture) under WebGL1. To have consistant
+        // behavior we always disable mipmaps.
+        mipmaps: false,
+        // Set MIN and MAG filtering parameters so mipmaps are not used in sampling.
+        // Set WRAP modes that support NPOT textures too.
+        parameters: {
+          [GL.TEXTURE_MIN_FILTER]: GL.NEAREST,
+          [GL.TEXTURE_MAG_FILTER]: GL.NEAREST,
+          [GL.TEXTURE_WRAP_S]: GL.CLAMP_TO_EDGE,
+          [GL.TEXTURE_WRAP_T]: GL.CLAMP_TO_EDGE
+        }
       });
     }
 
--- a/src/webgl/framebuffer.js
+++ b/src/webgl/framebuffer.js
@@ -151,8 +151,7 @@ export default class Framebuffer extends Resource {
 
   // Attach from a map of attachments
   attach(attachments, {
-    clearAttachments = false,
-    setFilteringParameters = true
+    clearAttachments = false
   } = {}) {
     const newAttachments = {};
 
@@ -185,15 +184,6 @@ export default class Framebuffer extends Resource {
         this._attachTexture({attachment, texture, layer, level});
       } else {
         this._attachTexture({attachment, texture: object, layer: 0, level: 0});
-      }
-
-      if (setFilteringParameters && object instanceof Texture2D) {
-        object.setParameters({
-          [GL.TEXTURE_MIN_FILTER]: GL.NEAREST,
-          [GL.TEXTURE_MAG_FILTER]: GL.NEAREST,
-          [GL.TEXTURE_WRAP_S]: GL.CLAMP_TO_EDGE,
-          [GL.TEXTURE_WRAP_T]: GL.CLAMP_TO_EDGE
-        });
       }
 
       // Resize objects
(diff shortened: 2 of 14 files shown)


src/utils/utils.js
```diff
@@ -76,3 +78,13 @@ function detach(elem) {
 
   return ans;
 }
+
+/**
+ * Verifies if a given number is power of two or not.
+ * @param {object} n - The number to check.
+ * @return {Array} Returns true if the given number is power of 2, false otherwise.
+ **/
+export function isPowerOfTwo(n) {
```
ibgreen: Thought - if we don't need this in multiple places, then maybe put inside texture.js?

src/webgl/framebuffer.js
```diff
@@ -187,15 +186,6 @@ export default class Framebuffer extends Resource {
         this._attachTexture({attachment, texture: object, layer: 0, level: 0});
       }
 
-      if (setFilteringParameters && object instanceof Texture2D) {
```
ibgreen: Two questions:
- Are there any requirements on filtering parameters in a texture when adding it to a framebuffer? I.e. could attachment fail if the filtering parameters are not correct? My assumption was that filtering has to be set to `GL.NEAREST`...

Note that framebuffer creates default attachments if none are specified. They need the right params.

src/webgl/texture.js
```diff
@@ -210,6 +221,25 @@ export default class Texture extends Resource {
     };
     const glSettings = Object.assign({}, DEFAULT_TEXTURE_SETTINGS, pixelStore);
 
+    // NOTE: NPOT workaround: Force update default settings.
+    if (this._isNPOTWorkaroundApplicable()) {
+
+      log.warn(0, 'NPOT (non power of two) texture found, disabling mipmaping');
+      mipmaps = false;
+      if (parameters[this.gl.TEXTURE_MIN_FILTER] === undefined) {
```
ibgreen: The `initialize()` function is a bit long, we should define helper functions and break it down.

src/webgl/texture.js
```diff
@@ -680,12 +722,15 @@ export default class Texture extends Resource {
...
-      this.gl.bindTexture(this.target, this.handle);
       this.gl.texParameteri(this.target, pname, param);
       break;
     }
 
     this.gl.bindTexture(this.target, null);
     return this;
   }
+
+  _isNPOTWorkaroundApplicable() {
```
ibgreen: Maybe a public function `isNPOT()` or `isPOT()` would be more useful?

src/webgl/texture.js
```diff
@@ -666,6 +687,27 @@ export default class Texture extends Resource {
   _setParameter(pname, param) {
     this.gl.bindTexture(this.target, this.handle);
 
+    // NOTE: Apply NPOT workaround
```
ibgreen: This looks like it could be broken out into a helper function.

src/webgl/texture.js
```diff
@@ -210,6 +221,25 @@ export default class Texture extends Resource {
...
+      if (parameters[this.gl.TEXTURE_MIN_FILTER] === undefined) {
+        log.warn(0, 'NPOT (non power of two) texture found, forcing TEXTURE_MIN_FILTER to LINEAR');
+        parameters[this.gl.TEXTURE_MIN_FILTER] = this.gl.LINEAR;
+      }
+      if (parameters[this.gl.TEXTURE_WRAP_S] === undefined) {
+        log.warn(0, 'NPOT (non power of two) texture found, forcing TEXTURE_WRAP_S to CLAMP_TO_EDGE');
+        parameters[this.gl.TEXTURE_WRAP_S] = this.gl.CLAMP_TO_EDGE;
+      }
+      if (parameters[this.gl.TEXTURE_WRAP_T] === undefined) {
+        log.warn(0, 'NPOT (non power of two) texture found, forcing TEXTURE_WRAP_T to CLAMP_TO_EDGE');
```
ibgreen: Use `log.once` to avoid error cascades.

src/webgl/texture.js
```diff
@@ -666,6 +687,27 @@ export default class Texture extends Resource {
   _setParameter(pname, param) {
     this.gl.bindTexture(this.target, this.handle);
 
+    // NOTE: Apply NPOT workaround
+    if (this._isNPOTWorkaroundApplicable()) {
+      const supportedMinFilters = [GL.LINEAR, GL.NEAREST];
```
ibgreen: Define a constant at top of file
`const NPOT_MIN_FILTERS`
you could make it an object, then you don't need to use indexOf.

src/utils/utils.js
ibgreen: The util function `detach` above is it even used? If not, delete it.

test/utils/utils.spec.js
```diff
@@ -30,3 +30,13 @@ test('Utils#uid', t => {
     'uid("prefix") starts with prefix');
   t.end();
 });
+
+test('Utils#isPowerOfTwo', t => {
```
ibgreen: Nice!

test/webgl/texture.spec.js
```diff
@@ -214,3 +214,66 @@ test('WebGL2#Texture2D setParameters', t => {
 
   t.end();
 });
+
+test('WebGL#Texture2D NPOT Workaround: texture creation', t => {
+  const {gl} = fixture;
```
ibgreen: Test cases should be run both on WebGL and WebGL2 context (only available in `test-browser`).

There should be some negative test cases, i.e. that test that the constructor throws exceptions when given NPOT textures under WebGL1.

src/webgl/framebuffer.js
1chandu: I can't find any such filtering restrictions on the texture attachment, also I tested lesson-16 and picking on several deck-gl layers, all looking good.

ibgreen: OK - They should only play a role when rendering to textures. I have had problems with this in the past but perhaps unrelated.

ibgreen: Meant to say : should only play a role when **reading from** textures

src/utils/utils.js
1chandu: `detach` is been used in `merge`, which is used in several other files.

src/webgl/framebuffer.js
1chandu: Yeah, thats is what my understanding too, those settings should affect sampling only and should have no impact when writing to a render target.

src/webgl/texture.js
1chandu: done, moved to a helper method.

test/webgl/texture.spec.js
1chandu: Wrote test cases for WebGL2, there are no exceptions to test, but I am testing parameter settings (for WebGL2, we shouldn't change these settings).

src/webgl/texture.js
1chandu: done, moved to a helper method.

test/setup/index.js
```diff
@@ -14,7 +14,7 @@ export function createTestContext(opts = {}) {
 }
 
 export const fixture = {
-  gl: createTestContext(),
+  gl: createTestContext({webgl2: false, webgl1: true}),
```
1chandu: createTestContext() with no input options using default options and hence creating WebGL2 context, I made this change to create WebGL1 context.

src/utils/utils.js
1chandu: Keeping this function here, as it is pretty generic and nothing to do with texture object.

PR #227, opened 2017-06-26
Update webgl-utils README

ibgreen:


ibgreen:
Landing now but please review, I will update post-facto.

PR #229, opened 2017-06-27
Rename demo to website

ibgreen:


ibgreen:
Aligning with other repos - calling the `website` as `demo` can be confusing to users that try to build it first instead of looking at the much simpler `examples` folder.

PR #230, opened 2017-06-27
Shader module refactor

ibgreen:
PR ready for review, 3 sub PRs broken out to focus this PR on shader module system only. A lot of files since there are a lot of small modules.


src/core/model.js
```diff
@@ -65,30 +77,25 @@ export default class Model extends Object3D {
...
-    }
-    this.getModuleUniforms = getUniforms || (x => {});
+    this.uniforms = {};
 
-    // set a custom program per o3d
-    this.program = program || new Program(gl, {vs, fs});
-    assert(this.program instanceof Program, 'Model needs a program');
+    // Make sure we have some reasonable default uniforms in place
+    uniforms = Object.assign({}, this.program.defaultUniforms, uniforms);
+    this.setUniforms(uniforms);
```
1chandu: nit: should we rename `setUniforms` to 'appendUniforms` ?

src/shadertools/modules/diffuse/diffuse.js
```diff
@@ -0,0 +1,58 @@
+// Minimal texture (diffuse map) support
```
1chandu: nit : why we use `diffuse` when this module isn't doing't any diffuse lighting?

src/shadertools/modules/dirlight/dirlight.js
```diff
@@ -0,0 +1,42 @@
...
+
+const fs = `\
+uniform vec3 dirlight_uLightDirection;
+
+/*
+ * Returns color attenuated by angle from light source
+ */
+vec4 dirlight_filterColor(vec4 color) {
+  vec3 normal = project_getNormal_World();
+  float d = abs(dot(normalize(normal), normalize(dirlight_uLightDirection)));
```
1chandu: nit: Can be fixed later, but shouldn't this be `max(0, dot(normalize(normal), normalize(dirlight_uLightDirection)))` ?

src/core/model.js
ibgreen: Agreed - we have too many `setUniforms` on Program, Model, Layer. We should take a good look at them and their semantics and it could be that we can find a better naming.

src/shadertools/modules/diffuse/diffuse.js
ibgreen: I am open to suggestions. I think the initial idea was to support a texture, which I usually see referred to as a diffuse map.

src/shadertools/modules/dirlight/dirlight.js
ibgreen: Good catch. Let's go through the modules one by one after we land.

PR #232, opened 2017-06-27
Move canvas and canvas context creation to webgl-utils

ibgreen:
Significantly simplifies `webgl/context.js` and `core/animation-loop.js`. I noticed these files were very messy to read and maintain when @1chandu and I walked through them.

Also contains some other small cleanup and few small bug fixes.


ibgreen:
This contains about half the code in the shader modules branch - once this lands I should be able to open a clean branch on the shader-module refactor.

src/core/animation-loop.js
```diff
@@ -122,171 +115,140 @@ export default class AnimationLoop {
...
     }
     return this;
   }
 
   // Resize canvas in "CSS coordinates" (may be different from device coords)
-  // NOTE: No effect on headless contexts
   // @param {Number} width, height - new width and height of canvas in CSS coordinates
   resizeCanvas(width, height) {
-    this._resizeCanvas(width, height);
+    resizeCanvas({width, height});
```
shaojingli: why this function is calling itself recursively

src/core/animation-loop.js
```diff
@@ -30,6 +27,11 @@ export default class AnimationLoop {
     glOptions = {},
     width = null,
     height = null,
+
+    // options
+    framebuffer = true, // Maintain an offscreen framebuffer of same size as screen
```
1chandu: `framebuffer` is later used as `Framebuffer` object, should this supposed to a boolean flag with a different name?

src/core/animation-loop.js
ibgreen: > why this function is calling itself recursively

@shaojingli Got an email with your question but can't see it in github. The answer is that it is actually calling a global function with the same name (since there is no `this.`). 

Agree it is a bit too subtle, the alternative is to give either the global function or the method a less natural name.

src/core/animation-loop.js
ibgreen: Good point. Removed the parameter for now as it wasn't used.

src/core/animation-loop.js
ibgreen: Renamed the function to make things clearer.

PR #238, opened 2017-06-28
Enable Query object for model draw timing and add unit tests

1chandu:
- Replace raw gl timer queries with Query object.
- Add more unit tests for Query class.

### DIFF
--- a/src/core/model.js
+++ b/src/core/model.js
@@ -10,6 +10,7 @@ import {MONOLITHIC_SHADERS, MODULAR_SHADERS} from '../shadertools/shaders';
 import {assembleShaders} from '../shadertools';
 
 import {addModel, removeModel, logModel, getOverrides} from '../debug/seer-integration';
+import Query from '../webgl/query';
 import assert from 'assert';
 
 const MSG_INSTANCED_PARAM_DEPRECATED = `\
@@ -137,9 +138,7 @@ export default class Model extends Object3D {
     // assert(program || program instanceof Program);
     assert(this.drawMode !== undefined && Number.isFinite(this.vertexCount), ERR_MODEL_PARAMS);
 
-    // TimerQuery - TODO replace with Query class
-    this.ext = this.program.gl.getExtension('EXT_disjoint_timer_query');
-    this.timerQueryEnabled = timerQueryEnabled && this.ext !== null;
+    this.timerQueryEnabled = timerQueryEnabled && Query.isSupported(this.gl, {timer: true});
     this.timeElapsedQuery = undefined;
     this.lastQueryReturned = true;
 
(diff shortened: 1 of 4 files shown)


src/core/model.js
```diff
@@ -326,54 +325,41 @@ export default class Model extends Object3D {
...
   _timerQueryStart() {
-    if (this.timerQueryEnabled === true && this.lastQueryReturned === true && this.ext) {
-      this.program.gl.getParameter(this.ext.GPU_DISJOINT_EXT);
-      this.timeElapsedQuery = this.ext.createQueryEXT();
-      this.ext.beginQueryEXT(this.ext.TIME_ELAPSED_EXT, this.timeElapsedQuery);
+    if (this.timerQueryEnabled === true) {
+      if (!this.timeElapsedQuery) {
+        this.timeElapsedQuery = new Query(this.gl);
+      }
+      if (this.lastQueryReturned) {
```
ibgreen: Is there a method or member on the query object we could use instead of keeping this extra state here: `lastQueryReturned`?

The point of the `Query` class is to simplify use in apps, but we it seems that we still need to do a lot of book keeping here.

test/webgl/query.spec.js
```diff
@@ -91,4 +117,141 @@ test('WebGL#Query completed/failed queries', t => {
...
+  const {gl2} = fixture;
+  if (!gl2) {
+    t.comment('WebGL2 not available, skipping tests');
+    t.end();
+    return;
+  }
+  testQueryCompleteFail(gl2, t);
+});
+
+function testdQuery(gl, opts, target, t) {
```
ibgreen: testdQuery: typo? If not, please spell out.

test/webgl/query.spec.js
```diff
@@ -91,4 +117,141 @@ test('WebGL#Query completed/failed queries', t => {
...
+  if (!Query.isSupported(gl, opts)) {
+    t.comment('Query API not supported, skipping tests');
+    t.end();
```
ibgreen: Maybe keep the t.end() in the calling function?

test/webgl/query.spec.js
```diff
@@ -91,4 +117,141 @@ test('WebGL#Query completed/failed queries', t => {
...
+  }
+
+  return query.promise.then(finalizer, finalizer);
+}
+
+test('WebGL#getTimestamp', t => {
+  const {gl} = fixture;
+
+  // const ext = gl.getExtension('EXT_disjoint_timer_query');
```
ibgreen: Kill commented code?

test/webgl/query.spec.js
```diff
@@ -91,4 +117,141 @@ test('WebGL#Query completed/failed queries', t => {
...
+
+test('WebGL2#getTimestamp', t => {
+  const {gl2} = fixture;
+  if (!gl2) {
+    t.comment('WebGL2 not available, skipping tests');
+    t.end();
+    return;
+  }
+
+  // const ext = gl2.getExtension('EXT_disjoint_timer_query_webgl2');
```
ibgreen: ditto

PR #239, opened 2017-06-28
Fixes

ibgreen:
Fixes a couple of the breakages in `test-browser`.

### DIFF
--- a/src/webgl/context-features.js
+++ b/src/webgl/context-features.js
@@ -50,26 +50,24 @@ Object.keys(WEBGL_FEATURES).forEach(key => {
 });
 export {FEATURES};
 
+// TODO - cache the value
 function getFeature(gl, cap) {
   const feature = WEBGL_FEATURES[cap];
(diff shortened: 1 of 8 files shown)


1chandu:
@ibgreen fixed NPOT failures, it was due to creating WebGL2 context even when WebGL1 was requeted.

Two more failures to fix, in 'context-feature-spec.js'.

src/webgl/context-features.js
```diff
@@ -80,7 +78,7 @@ export function hasFeature(gl, feature) {
 export function hasFeatures(gl, features) {
   features = Array.isArray(features) ? features : [features];
   return features.every(feature => {
-    return getFeature(gl, feature) || gl.getExtension(feature);
```
1chandu: @ibgreen , this would always return true since , gl.getExtension would return an object. Given we are also checking gl.getExtension in getFeature, removing it here.

PR #240, opened 2017-06-29
Convert any v3 texture parameters to v4

### LINKED ISSUE #197: Ensure compatibility with luma.gl v3 Texture parameters
We need to add test cases in luma.gl to cover the failed size deduction in https://github.com/uber/deck.gl/issues/683.

Landing https://github.com/uber/deck.gl/pull/684 will help debugging in deck.gl

1chandu:
Fixes #197

### DIFF
--- a/src/webgl/texture.js
+++ b/src/webgl/texture.js
@@ -112,6 +112,13 @@ export const TEXTURE_FORMATS = {
   // [GL.COMPRESSED_RGBA_ATC_INTERPOLATED_ALPHA_WEBGL]: {compressed: true, gl1: ETC1}
 };
 
+const V3_TO_V4_OPTIONS_MAP = {
+  magFilter: [GL.TEXTURE_MAG_FILTER],
+  minFilter: [GL.TEXTURE_MIN_FILTER],
+  wrapS: [GL.TEXTURE_WRAP_S],
+  wrapT: [GL.TEXTURE_WRAP_T]
+};
+
 function isFormatSupported(gl, format) {
   assert(isWebGL(gl), ERR_WEBGL);
   const info = TEXTURE_FORMATS[format];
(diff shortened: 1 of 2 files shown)


src/webgl/texture.js
```diff
@@ -112,6 +112,13 @@ export const TEXTURE_FORMATS = {
   // [GL.COMPRESSED_RGBA_ATC_INTERPOLATED_ALPHA_WEBGL]: {compressed: true, gl1: ETC1}
 };
 
+const V3_TO_V4_OPTIONS_MAP = {
```
1chandu: @ibgreen , in v3 branch, setParameters() has following arguments:
```
setParameters({
    magFilter,
    minFilter,
    wrapS,
    wrapT,
    // WEBGL2
    wrapR,
    baseLevel,
    maxLevel,
    minLOD,
    maxLOD,
    compareFunc,
    compareMode
  }) {
```
Assuming I should add all of them, can you please confirm?

src/webgl/texture.js
```diff
@@ -608,6 +626,21 @@ export default class Texture extends Resource {
   }
   */
 
+  // OVERRIDES
+  setParameters(parameters) {
```
1chandu: Are these deprecated options present in `opts.parameters` or in `opts` (in constructor), if it is in `opts`,  I may have to extract them , add it to `parameters` before calling `setParameters` in constructor ?

src/webgl/texture.js
ibgreen: I am pretty sure the WebGL2 parameters have never been used, since there were other issues with the WebGL2 texture support in v3.

So for compatibility reasons, the answer is no.

We could decide to keep this API, it is a little easier to use than GL ENUMS. Although it would be the only class for which we have our own parameters. What do you think?

src/webgl/texture.js
ibgreen: Yes they are in opts, not in opts.parameters. You need to extract them (there is a code example in the issue).

src/webgl/texture.js
1chandu: Either way is fine, not a big difference but I would opt for GL ENUM style, couple of reasons
1. Don't see much advantage other than shorter names.
2. Since we are using enums in other places, lets keep same style every where.

ibgreen: >  I would opt for GL ENUM style 

Sounds good! Let's just add backwards compatibility support for the old opts then, with deprecation warnings.

src/webgl/texture.js
1chandu: Done!

src/webgl/texture.js
1chandu: Done!

PR #251, opened 2017-07-05
Fix picking module & add FB unit tests.

1chandu:
- Fix readPixels in picking module.
- Add FrameBuffer unit tests for readPixels.

### DIFF
--- a/src/core/pick-models.js
+++ b/src/core/pick-models.js
@@ -51,8 +51,13 @@ export default function pickModels(gl, {
       model.setUniforms({picking_uActive: 0});
 
       // Sample Read color in the central pixel, to be mapped as a picking color
-      const color = new Uint8Array(4);
-      gl.readPixels(deviceX, deviceY, 1, 1, gl.RGBA, gl.UNSIGNED_BYTE, color);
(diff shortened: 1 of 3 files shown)


src/utils/typed-array-utils.js
```diff
@@ -35,7 +35,7 @@ export function getGLTypeFromTypedArray(arrayOrType) {
 // Used to auto deduce gl parameter types
 
 /* eslint-disable complexity */
-export function getTypedArrayFromGLType(glType, {clamped = true} = {}) {
+export function getTypedArrayFromGLType(glType, {clamped = false} = {}) {
```
1chandu: When type is 'GL_UNSIGNED_BYTE', `readPixels` expects  `Uint8Array` , so setting this to false by default. Tested all examples and lessons, should I test anything else for this change?

ibgreen: Why set this to false by default? Can't we specify the param in readPixels instead? A little worried that we break something else...

1chandu: Yeah agree, made that change.

PR #252, opened 2017-07-06
Fix picking color encoding.

1chandu:
Current encoding is going beyond max value 255, and resulting in wrong picking color. Fix the encoding. (Thanks to @ibgreen ).


ibgreen:
@1chandu solves the scaling problem!

I noticed the picking color is not cleared when moving out of the window, probably the picking modules `getUniforms` needs a fix in handle the `nullPickingColor`.

PR #253, opened 2017-07-06
math library fixes

ibgreen:
A light cleanup pass on the math package.

Includes handling null arguments in `MathArray.equals` and `exactEquals`.


src/packages/math/src/matrix4.js
```diff
@@ -28,6 +28,18 @@ import vec4_transformMat4 from 'gl-vec4/transformMat4';
 
 const IDENTITY = [1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1];
 
+export function validateMatrix4(m) {
+  return m.length === 16 &&
+    Number.isFinite(m[0]) && Number.isFinite(m[1]) &&
```
Pessimistress: Why not just `m.every(Number.isFinite)`?

src/packages/math/src/common.js
```diff
@@ -96,12 +96,24 @@ export function atan(radians) {
...
+      return false;
+    }
+    for (let i = 0; i < a.length; ++i) {
+      if (!equals(a[i], b[i])) {
+        return false;
+      }
+    }
+    return true;
+  }
+  return Math.abs(a - b) <= config.EPSILON * Math.max(1.0, Math.abs(a), Math.abs(b));
```
Pessimistress: Technically you could just do `Math.max(1, Math.abs(a))` as the extra margin of error is only `EPSILON * delta`

shaojingli: I don't really understand this function... what is the reason that we turn to absolute error margin when abs(a) and abs(b) is small? Reference: http://floating-point-gui.de/errors/comparison/

src/packages/math/src/vector4.js
```diff
@@ -17,12 +16,11 @@ import vec4_dot from 'gl-vec4/dot';
...
-// import vec4_divide from 'gl-vec4/divide';
-// import vec4_ceil from 'gl-vec4/ceil';
-// import vec4_floor from 'gl-vec4/floor';
-// import vec4_min from 'gl-vec4/min';
-// import vec4_max from 'gl-vec4/max';
+export function validateVector4(v) {
+  return v.length === 4 &&
+    Number.isFinite(v[0]) && Number.isFinite(v[1]) &&
+    Number.isFinite(v[2]) && Number.isFinite(v[3]);
+}
```
shaojingli: ditto

src/packages/math/src/matrix4.js
ibgreen: > Why not just m.every(Number.isFinite)?

This is old code that I just moved around. Pretty sure this is due to a premature assumption about speed, that inline comparisons are than a loop of function calls,.

We already know that this math library has a performance issue with class creation and it really needs a benchmark suite - out of scope for this PR.

src/packages/math/src/common.js
ibgreen: > Technically you could just do Math.max(1, Math.abs(a))

This is the way gl-matrix does comparisons, since this library replaces gl-matrix based code it should be safe to compare this way.

The paper is interesting. I am open to changing the comparison method, though I'd prefer to see such a change accompanied by tests.

PR #266, opened 2017-07-12
Fix Shader module dependency ordering

1chandu:
- When parsing module dependency graph, update level only if the existing level is lower.
- Add unit tests.

### DIFF
--- a/src/shadertools/lib/shader-modules.js
+++ b/src/shadertools/lib/shader-modules.js
@@ -84,20 +84,16 @@ export function getShaderDependencies(modules) {
   return Object.keys(result).sort((a, b) => result[b] - result[a]);
 }
 
-// PRIVATE API
-
-function registerShaderModule(shaderModule, {ignoreMultipleRegistrations = false}) {
-  assert(shaderModule.name, 'shader module has no name');
-  if (!ignoreMultipleRegistrations && shaderModules[shaderModule.name]) {
-    // TODO - instead verify that definition is not changing...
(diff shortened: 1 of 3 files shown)


src/shadertools/lib/shader-modules.js
```diff
@@ -84,28 +84,26 @@ export function getShaderDependencies(modules) {
...
+ * Recursively checks module dpendencies to calculate dependency
+ * level of each module.
+ *
+ * @param {String[]} modules - Array of modules
+ * @param {Number} level - Current level
+ * @return {result} - Map of module name to its level
+ */
 // Adds another level of dependencies to the result map
-function getDependencyGraph({modules, level, result}) {
+export function getDependencyGraph({modules, level, result}) {
```
1chandu: @ibgreen @shaojingli , is there a better way instead of exporting it, so I can unit test this method?

src/shadertools/lib/shader-modules.js
```diff
@@ -123,3 +121,15 @@ function getDependencyGraph({modules, level, result}) {
 
   return result;
 }
+
+// PRIVATE API
+
+function registerShaderModule(shaderModule, {ignoreMultipleRegistrations = false}) {
```
1chandu: No change, here , just moved it.

src/shadertools/lib/shader-modules.js
shaojingli: exporting it is okay

PR #278, opened 2017-07-25
Website fix

shaojingli:
Still decide to pull up a PR for the website fix
@Pessimistress


website/src/components/markdown-page.js
```diff
@@ -26,7 +25,6 @@ const urlRewrites = {
  */
 const imageRewrites = {};
 
-@pureRender
 export default class MarkdownPage extends Component {
```
Pessimistress: `PureComponent`?

PR #280, opened 2017-07-26
Website styling

Pessimistress:
Did I break the following?
- The transform feedback demo doesn't seem to do anything
- Examples don't work as standalone


1chandu:
@Pessimistress 
TF demo should just renders RGB gradient square.
I never tested examples stand alone, do they work before your change?

Pessimistress:
@1chandu I don't think so. We should fix them though.

PR #288, opened 2017-08-02
Picking shader module: Handle invalid/null picking color

1chandu:
- Instead of overloading `pickingSelectedColor` with a unique null color value (`pickingNullColor`), use a boolean enum, `pickingValid`. This way it is more clear and within the shader we only do boolean check instead of vec3 compare.
- Do not highlight models when `pickingValid` is false.
- Update picking/instancing apps to listen to `mouseleave` events and set pickPosition to null.
- Tested with picking/instancing core examples.
@shaojingli we need this to be ported to release branch.

### DIFF
--- a/examples/core/instancing/app.js
+++ b/examples/core/instancing/app.js
@@ -50,7 +50,12 @@ const animationLoop = new AnimationLoop({
       uModel: new Matrix4().rotateX(tick * 0.01).rotateY(tick * 0.013)
     });
 
-    const pickInfo = pickModels(gl, {models: [cube], position: pickPosition, framebuffer});
+    const pickInfo = pickPosition && pickModels(gl, {
+      models: [cube],
+      position: pickPosition,
+      framebuffer
+    });
+
     cube.updateModuleSettings({
       pickingSelectedColor: pickInfo && pickInfo.color,
       pickingValid: pickInfo !== null
--- a/examples/core/instancing/app.js
+++ b/examples/core/instancing/app.js
@@ -8,6 +8,9 @@ let pickPosition = [0, 0];
 function mousemove(e) {
   pickPosition = [e.offsetX, e.offsetY];
 }
+function mouseleave(e) {
+  pickPosition = null;
+}
 
 const animationLoop = new AnimationLoop({
   onInitialize({gl}) {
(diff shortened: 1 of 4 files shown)


examples/core/instancing/app.js
```diff
@@ -47,21 +51,13 @@ const animationLoop = new AnimationLoop({
     });
 
     const pickInfo = pickModels(gl, {models: [cube], position: pickPosition, framebuffer});
-    // console.log(pickInfo ? pickInfo.color : 'picked background');
     cube.updateModuleSettings({
-      pickingSelectedColor: pickInfo && pickInfo.color
+      pickingSelectedColor: pickInfo && pickInfo.color,
+      pickingValid: pickInfo !== null
```
ibgreen: Do we really need this separate variable? Isn't `pickingSelectedColor` set to `null` when `pickInfo` is `null`?

examples/core/instancing/app.js
```diff
@@ -47,21 +51,13 @@ const animationLoop = new AnimationLoop({
...
     });
 
     gl.clear(GL.COLOR_BUFFER_BIT | GL.DEPTH_BUFFER_BIT);
     cube.render();
-  },
-  onMouseMove({gl, tick, cube, framebuffer}) {
```
ibgreen: It would be neat to integrate with luma's `addEvents` here. Or even align with the new event handling in deck.gl. 

Will require some discussion, maybe open an issue...

src/core/pick-models.js
```diff
@@ -21,6 +21,10 @@ export default function pickModels(gl, {
   assert(isWebGL(gl), ILLEGAL_ARG);
   assert(framebuffer, ILLEGAL_ARG);
 
+  if (position === null) {
```
ibgreen: 1) `if (!position)` is usually better, also handles `position === undefined`.

2) Should the `position === null` case be handled inside this function, or outside, by the caller? 

I find it a little surprising that it is handled here, I wouldn't expect this behavior from this function.

src/shadertools/modules/picking/picking.js
```diff
@@ -3,18 +3,18 @@ const DEFAULT_HIGHLIGHT_COLOR = new Uint8Array([0, 64, 128, 64]);
 const DEFAULT_MODULE_OPTIONS = {
   pickingSelectedColor: null, //  Set to a picking color to visually highlight that item
   pickingHighlightColor: DEFAULT_HIGHLIGHT_COLOR, // Color of visual highlight of "selected" item
-  pickingNullColor: [-1, 0, 0],
```
ibgreen: Don't we need to let the user define the `nullColor`?

PR #290, opened 2017-08-03
Picking shader module: Handle invalid/null picking color

1chandu:
cherry-pick #288 from master.
CC: @shaojingli


1chandu:
This is wrongly created against master, intention was to create this against 4.0-release branch.

Issue #291, opened 2017-08-09
Scatterplot layer crashes in Safari


1chandu:
This is due to 'gl.bufferData' call crashes on safari when size is 0 (second argument), but works fine on Chrome. Copy pasting email contents describing the issue :

CC: @ibgreen @shaojingli 

> After I upgrade deck.gl version from 4.0.3 to 4.1.0, Safari does not work.
> Here is a situation details.
> 
> deck.gl@4.0.3 and 4.1.0 on Chrome is work well
> deck.gl@4.0.3 on Safari was work well
> Here is a code to use deck.gl
> https://github.com/sideroad/floras/blob/master/src/components/WorldMap.jsx
> Working example
> 
> https://floras.herokuapp.com/en
> 
> Error details
> 
> Safari throw error in scatterplot-layer
> https://github.com/uber/deck.gl/blob/master/src/layers/core/scatterplot-layer/scatterplot-layer.js#L104
> because of this.state.model is null.
> 
> But when I open site on Safari below, no error happen...
> https://uber.github.io/deck.gl/#/examples/core-layers/scatterplot-layer
> https://uber.github.io/deck.gl/#/documentation/layer-catalog/scatterplot-layer
> 
> If there is any good way to debug my code, plz tell me.

ibgreen:
> This is due to 'gl.bufferData' call crashes on safari when size is 0 (second argument)

Can we just force size to 1 when 0? Allocating too much should not break any apps. We can add a comment that this is needed on MacOS.

1chandu:
With reference to conformance test, by creating zero sized array fixes the issue , I will create PR soon.

https://www.khronos.org/registry/webgl/sdk/tests/conformance/more/functions/bufferData.html?webglVersion=1&quiet=0
https://www.khronos.org/registry/webgl/sdk/tests/conformance/more/functions/bufferDataBadArgs.html?webglVersion=1&quiet=0

1chandu:
Resolved by #292.

PR #292, opened 2017-08-09
bufferData workaround to fix Safari crash.

1chandu:
Fix for #291 
Tested with deck.gl/website layer examples and deck.gl/example/layer-browser.

### DIFF
--- a/src/webgl/buffer.js
+++ b/src/webgl/buffer.js
@@ -100,7 +100,14 @@ export default class Buffer extends Resource {
 
     if (!data) {
       type = type || GL.FLOAT;
-      bytes = bytes || 0;
+
+      // Workaround needed for Safari (#291):
+      // gl.bufferData with size (second argument) equal to 0 crashes.
+      // hence create zero sized array.
+      if (!bytes || bytes === 0) {
+        bytes = 0;
+        data = new Float32Array(0);
+      }
     } else {
       type = type || getGLTypeFromTypedArray(data);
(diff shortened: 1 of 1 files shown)


src/webgl/buffer.js
```diff
@@ -100,7 +100,14 @@ export default class Buffer extends Resource {
...
     if (!data) {
       type = type || GL.FLOAT;
-      bytes = bytes || 0;
+
+      // Workaround needed for Safari (#291):
+      // gl.bufferData with size (second argument) equal to 0 crashes.
+      // hence create zero sized array.
+      if (!bytes || bytes === 0) {
+        bytes = 0;
+        data = new Float32Array(0);
```
1chandu: Maybe I should not change incoming 'data' object, but create and set one to `this.data` and leave `data` as is?

ibgreen: > Maybe I should not change incoming 'data' object, but create new array and set it to this.data and leave data as is?

Well `data` is just a reference, setting it to a new `Array` doesn't change what was passed in.

The more important think is what we set `this.data` to. Should we still set it to `null`? It will be `null` if `data=null` and `bytes>0` so probably should be `null` if `bytes===0` as well?

1chandu: this.data will hold same value that we have passed to `bufferData()` (if any). Don't see any where else it is used, so I think it should be fine.

PR #297, opened 2017-08-28
Debug functionality polish

### LINKED ISSUE #296: Debug improvements
The debug logging is in need of some polish:
- [x] wrap uniform/attribute tables in group (more compact log level 2)
- [x] remove rendering model end log (unnecessary line)
- [x] sort uniforms in table - first local uniforms, then module uniforms
- [x] title of model (layer) in table - more prominent, don't waste row
- [x] unify middle columns in attribute table (location and type/size/vert column)
- [x] reduce #lines of log from queryTimer (`model.js`) - enabled for all when seer is active

ibgreen:
Fixes #296

### DIFF
--- a/src/core/model.js
+++ b/src/core/model.js
@@ -321,11 +321,11 @@ export default class Model extends Object3D {
 
     this.setUniforms(resolvedUniforms);
 
-    log.log(2, `>>> RENDERING MODEL ${this.id}`, this);
+    log.group(2, `>>> RENDERING MODEL ${this.id}`, {collapsed: log.priority <= 2});
 
     this.setProgramState();
 
-    this._logAttributesAndUniforms(3, resolvedUniforms);
+    this._logAttributesAndUniforms(2, resolvedUniforms);
 
     this.onBeforeRender();
 
@@ -355,7 +355,7 @@ export default class Model extends Object3D {
 
     this.setNeedsRedraw(false);
 
-    log.log(2, `<<< RENDERING MODEL ${this.id} - complete`);
+    log.groupEnd(2, `>>> RENDERING MODEL ${this.id}`);
 
     return this;
   }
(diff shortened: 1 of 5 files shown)


src/core/model.js
```diff
@@ -405,12 +405,12 @@ export default class Model extends Object3D {
...
-        log.log(2, 'program.id: ', this.program.id);
-        log.log(2, `last frame time: ${this.stats.lastFrameTime}ms`);
-        log.log(2, `average frame time ${this.stats.averageFrameTime}ms`);
-        log.log(2, `accumulated frame time: ${this.stats.accumulatedFrameTime}ms`);
-        log.log(2, `profile frame count: ${this.stats.profileFrameCount}`);
-
+        log.log(2, `\
+GPU time ${this.program.id}: ${this.stats.lastFrameTime}ms \
+avg ${this.stats.averageFrameTime}ms \
+acc: ${this.stats.accumulatedFrameTime}ms \
```
1chandu: Nit: should we use full names `average` and `accumulated` to be clear?

src/webgl/uniforms.js
```diff
@@ -213,42 +224,38 @@ export function getUniformsTable({
...
-      table[uniformName] = {
-        Type: uniform,
-        Value: formatValue(uniform)
-      };
+  // Add program's provided uniforms (in alphabetical order)
+  const uniformNames = Object.keys(uniformLocations).sort();
+
+  // First add non-underscored uniforms (assumed not coming from shader modules)
+  for (const uniformName of uniformNames) {
+    if (!uniformName.match(SHADER_MODULE_UNIFORM_REGEXP)) {
```
1chandu: Doesn't have to be this PR, but displaying these uniforms in alphabetically sorted order will be very useful, especially when debugging shaders that have many uniforms, you can easily point to the one you need.
Able to display only a list user provided ones (using something like luma.log.uniformsList = ['uniform1', 'uniform2']) will be even great.

src/core/model.js
ibgreen: > Nit: should we use full names average and accumulated to be clear?

We could. I was trying to keep this to a short line in the log. But probably the longer words could fit.

src/webgl/uniforms.js
ibgreen: >  but displaying these uniforms in alphabetically sorted order will be very useful, especially when debugging shaders that have many uniforms

That is part of this PR, see like 233. It first sorts alphabetically, and then it sorts so that any uniforms that have underscored in them come last (so the shaders own uniforms first, then shader module uniforms).

Unfortunately this didn't fix the corresponding issue in the seer integration which seems to need separate sorting.

PR #298, opened 2017-08-29
Fix fp64 test

shaojingli:
@ibgreen @1chandu

### DIFF
--- a/src/shadertools/modules/fp64/test/fp64-shader.spec.js
+++ b/src/shadertools/modules/fp64/test/fp64-shader.spec.js
@@ -24,8 +24,7 @@
 import 'babel-polyfill';
 
 import {document, window} from 'global';
-import {Buffer, createGLContext, Program, setParameters} from 'luma.gl';
-import {assembleShaders} from 'deck.gl';
+import {Buffer, createGLContext, Program, setParameters, assembleShaders, registerShaderModules, fp64} from 'luma.gl';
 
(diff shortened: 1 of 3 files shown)


src/shadertools/modules/fp64/test/fp64-shader.spec.js
```diff
@@ -139,11 +138,11 @@ function initializeGL(canvas)
   });
   gl.clear(gl.COLOR_BUFFER_BIT | gl.DEPTH_BUFFER_BIT);
 
-  var fp_texture_support = gl.getExtension('OES_texture_float');
-  if (!fp_texture_support)
-  {
-    console.error("no floating point texture support!");
-  }
+  // var fp_texture_support = gl.getExtension('OES_texture_float');
```
ibgreen: Was the check removed due to WebGL2 not supporting this extension?

If so, an option is using luma.gl capability system to do this check, it will do the right thing both on WebGL 1 & 2.

src/shadertools/modules/fp64/test/fp64-shader.spec.js
```diff
@@ -251,9 +250,9 @@ function test_float_add(gl, testName) {
   const program = new Program(gl, assembleShaders(gl, {
         vs: require('./vs_float_add.glsl'),
         fs: require('./fs.glsl'),
-        fp64: true,
-        project64: true
-      }));
+        modules: ['fp64']
```
ibgreen: Hmm I thought `modules` was only supported by `Model` not` Program`. But if this works, then of course fine!

PR #301, opened 2017-08-31
Use shaderCache to avoid re-compilaiton of same shader.

### LINKED ISSUE #299: `shaderCache` param does not seem to be used by `Model`
All deck.gl layers are sending a `shaderCache` parameter to the `Model` constructor, assuming they will get "pre-compiled" `Shader`s from the cache if a second instance of a layer is created.

But I can not see any code in `Model` that handles the `shaderCache` parameter. 

Somehow this must have regressed when we moved `assembleShaders` from deck to luma.

Caching is important for startup perf on some GPU drivers.

@Pessimistress @shaojingli

1chandu:
Verified using examples, lessons and test-browser.
Update picking example to verify cache hit.
Update webpack.config files of examples, so `npm run start-local` works.
Fixes #299

### DIFF
--- a/src/core/model.js
+++ b/src/core/model.js
@@ -1,3 +1,4 @@
+/* eslint quotes: ["error", "single", { "allowTemplateLiterals": true }]*/
 // A scenegraph object node
 import {GL, Buffer, Program, withParameters, checkUniformValues, isWebGL} from '../webgl';
 // import {withParameters} from '../webgl/context-state';
@@ -55,7 +56,9 @@ export default class Model extends Object3D {
     defaultUniforms,
 
     // 3: Pre-created program
-    program,
+    program = null,
+
+    shaderCache = null,
 
     isInstanced = false, // Enables instanced rendering
     instanced, // deprecated
@@ -85,7 +88,8 @@ export default class Model extends Object3D {
       defines,
       moduleSettings,
       defaultUniforms,
-      program
+      program,
+      shaderCache
     });
 
     this.uniforms = {};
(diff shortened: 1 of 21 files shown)


examples/lessons/03/app.js
```diff
@@ -1,7 +1,6 @@
 /* eslint-disable array-bracket-spacing, no-multi-spaces */
 import {
-  GL, AnimationLoop, Program, Model, Geometry, Matrix4,
-  resetParameters, setParameters
+  GL, AnimationLoop, Model, Geometry, Matrix4, setParameters, ShaderCache
```
ibgreen: Very good idea: to make sure shader cache is used by one of the examples. This will help us catch regressions!

Although, I would suggest that we do this in Lesson 9 instead of this lesson. The reason is that lesson 9 creates 50 copies of a `Star` model and I can see an almost second long delay when I start it, so the introduction of the shaderCache in that example would be justified (I would expect it to completely eliminate the startup delay), and could be accompanied with a comment explaining the optimization technique. Could even deserve a bullet in the little HTML panel.

Lesson 3 is supposed to be a minimal example introducing only small incremental concepts, so I think it would be better to leave the shader cache out of this example.

src/core/model.js
```diff
@@ -157,28 +160,42 @@ export default class Model extends Object3D {
...
+      }
 
-    const assembleResult = assembleShaders(this.gl, {vs, fs, modules, defines});
-    ({vs, fs} = assembleResult);
-    const {getUniforms} = assembleResult;
+      const assembleResult = assembleShaders(this.gl, {vs, fs, modules, defines});
+      ({vs, fs} = assembleResult);
 
-    this.getModuleUniforms = getUniforms || (x => {});
+      // Retrive compiled shaders from cache if exist.
```
ibgreen: Note: shader cache creates them if the don't exist, adding them to the cache.

src/core/model.js
```diff
@@ -157,28 +160,42 @@ export default class Model extends Object3D {
...
+      if (!vs) {
+        vs = MODULAR_SHADERS.vs;
+      }
+      if (!fs) {
+        fs = MODULAR_SHADERS.fs;
+      }
+
+      // Assign default uniforms (if any default shaders are being used)
+      if (vs === MONOLITHIC_SHADERS.vs || fs === MONOLITHIC_SHADERS.fs) {
+        defaultUniforms = defaultUniforms || MONOLITHIC_SHADERS.defaultUniforms;
```
ibgreen: Unrelated, but a good time to do it: We should have a deprecation warning here. We want to phase out the monolithic shaders completely in next release.

src/core/model.js
```diff
@@ -157,28 +160,42 @@ export default class Model extends Object3D {
...
-    const assembleResult = assembleShaders(this.gl, {vs, fs, modules, defines});
-    ({vs, fs} = assembleResult);
-    const {getUniforms} = assembleResult;
+      const assembleResult = assembleShaders(this.gl, {vs, fs, modules, defines});
+      ({vs, fs} = assembleResult);
 
-    this.getModuleUniforms = getUniforms || (x => {});
+      // Retrive compiled shaders from cache if exist.
+      if (shaderCache) {
+        vs = shaderCache.getVertexShader(this.gl, vs);
```
ibgreen: What happens if the shaders are already `Shader` instances (i.e. not strings). Will ShaderCache handle that case?  Should we add a test case?

Also, not sure if supporting both Strings and Shaders is documented behavior, might be worth double checking the docs and make sure things are correctly described, at the same time as making sure the `shaderCache` parameter is added to `Model`.

examples/lessons/03/app.js
1chandu: Agree, we should keep lesson 03 to show minimal increments, will remove it from there. In lesson 09 we create `Program` object once and use it all copies of `Star` model, so shaders are compiled only once. Only other better place I see is `picking` example, which creates multiple `Sphere` models, will add `shaderCache` there.

src/core/model.js
1chandu: ShaderCache only handles 'strings', and this is already documented in shader-cache section.

ibgreen: > ShaderCache only handles 'strings', and this is already documented in shader-cache section.

Yes that sounds right. To be clear, then I was probably correct in assuming that this code won't work when `Shader`s are passed in right. That is why it would be good to have a test case for `Model` being instantiated with `Shader`s instead of strings.

PR #305, opened 2017-09-12
Lesson 11 example for creating textured and lighted sphere

jianhuang01:
New example of creating a sphere with moon texture and basic lighting, demonstrates sphere API, model matrix operation and mouse event handling.


examples/lessons/11/app.js
```diff
@@ -0,0 +1,206 @@
...
+      uPMatrix: new Matrix4().perspective({fov: 45 * Math.PI / 180, aspect, near: 0.1, far: 100})
+    });
+  }
+});
+
+animationLoop.getInfo = () => {
+  return `
+  <p>
+    <a href="http://learningwebgl.com/blog/?p=1253" target="_blank">
+      Improving the code structure with lots of moving objects
```
ibgreen: Same comment as last PR: please update the title of the lesson.

In this case if you follow the link: "Spheres, Rotation Matrices, and Mouse Events" (instead of "Improving the code structure with lots of moving objects", which is Lesson 9 title)

examples/lessons/11/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "contributors": [
+    "Jian Huang <jianh@uber.com>"
+  ],
+  "scripts": {
+    "start": "webpack-dev-server --progress --hot --open -d",
+    "start-local": "webpack-dev-server --env.local --progress --hot --open -d"
+  },
+  "dependencies": {
+    "luma.gl": ">=4.0.0",
+    "math.gl": "1.0.0-alpha.7"
```
ibgreen: `>=1.0.0-alpha.7`, will pick up improved versions. Current form will pin it.

examples/lessons/11/app.js
```diff
@@ -0,0 +1,206 @@
...
+      clearDepth: 1,
+      depthTest: true
+    });
+
+    return loadTextures(gl, {
+      urls: ['moon.gif']
+    })
+    .then(textures => {
+      let program = new Program(gl, {
+        fs: FRAGMENT_SHADER,
```
ibgreen: I think you can just add `fs` and `vs` as parameter to `Sphere` and it will create a `Program` internally.

examples/lessons/11/app.js
```diff
@@ -0,0 +1,206 @@
+/* eslint-disable no-var, max-statements */
```
ibgreen: Just a thought: Is linting set up for the examples? Long term we want to remove as many of these linter exceptions as possible. Especially the `no-var` would be good to get rid off. Try to remove the others and do `npm run lint` to see if they are still needed.

examples/lessons/11/app.js
```diff
@@ -0,0 +1,206 @@
...
+    onDragEnd(e) {
+      mouseMovement.mouseDown = false;
+    }
+  });
+}
+
+export default animationLoop;
+
+// expose on Window for standalone example
+window.animationLoop = animationLoop; // eslint-disable-lie
```
ibgreen: Type: `eslint-disable-line`

examples/lessons/11/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "luma.gl-examples-lesson11",
+  "description": "An es6 example for luma.gl.",
```
ibgreen: Drop the `name` field, we are not going to publish this as a module, and we want these files to be simple.
We could keep the description but maybe give it better "attribution" (we are copying some other examples after all):
`description: The classic WebGL Lesson 11 ported to luma.gl`

examples/lessons/11/app.js
jianhuang01: Lint still doesn't work for me, now I manually make sure there is no vars in my example code, so shouldn't keep these comments here.

examples/lessons/11/app.js
jianhuang01: cool, good to know that.

examples/lessons/11/app.js
jianhuang01: thanks, fixed both examples.

examples/lessons/11/package.json
jianhuang01: done.

examples/lessons/11/package.json
jianhuang01: done.

PR #306, opened 2017-09-13
Keep version directive at the beginning of the shader during assembly.

### LINKED ISSUE #304: `assembleShaders` prepends code to shader defeating GLSL ES 3.00 version directive
I'm working on a deck.gl shader with bit-wise and modulus operators. I am getting errors:

    ^^^ ERROR: '%' : integer modulus operator supported in GLSL ES 3.00 and above only

I have been browsing documentation, but couldn't find anything on specifying shader versions?

1chandu:
Fixes #304
Tested with test-browser and lesson04 update to use GLSL 3.0 shaders.

### DIFF
--- a/src/shadertools/lib/assemble-shaders.js
+++ b/src/shadertools/lib/assemble-shaders.js
@@ -116,7 +116,18 @@ ${type === FRAGMENT_SHADER ? FRAGMENT_SHADER_PROLOGUE : ''}
   }
 
   // Add the actual source of this shader
-  assembledSource += source;
+  if (source.indexOf('#version ') === 0) {
+    // Keep any version directive at the begining of the shader
+    // TODO : keep all pre-processor statements at the begining of the shader.
+    const lines = source.split('\n');
+    const assembledLines = [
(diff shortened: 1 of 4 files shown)


src/shadertools/lib/assemble-shaders.js
```diff
@@ -116,7 +116,18 @@ ${type === FRAGMENT_SHADER ? FRAGMENT_SHADER_PROLOGUE : ''}
   }
 
   // Add the actual source of this shader
-  assembledSource += source;
+  if (source.indexOf('#version ') === 0) {
```
ibgreen: Nit: while this is functionally fine, to me the logic would be cleaner by separating this into two steps:
1) Early on split the `source` into `versionLine` & `coreSource`, where `versionLine` can be the empty string.
2) In the end, just append `versionLine` and `coreSource` in the appropriate positions, without any if statements.

Long term this would also allow us to deduce the version of the shader early on to do version dependent injection as well as do further breakdown of the core source to find additional injection points.

Issue #300, opened 2017-08-30
GLSL errors are reported both as errors and as multiple warnings


ibgreen:
<img width="514" alt="screen shot 2017-08-30 at 12 24 26 pm" src="https://user-images.githubusercontent.com/7025232/29890946-454905c2-8d7e-11e7-902f-a92a42e37491.png">

Issue #273, opened 2017-07-20
Allow cross origin image load in luma.gl loadTextures()


shaojingli:

ibgreen:
This should be a one-line change. We just need for forward `opts` to `loadImage` here: https://github.com/uber/luma.gl/blob/master/src/io/load-files.js#L44

Issue #259, opened 2017-07-09
Verify GL.MIN, GL.MAX blendEquation settings under WebGL1 context.


ibgreen:
Add test case for GL.MIN, GL.MAX under WebGL1 context, just to make sure the extension is used in that case.

1chandu:
>  just to make sure the extension is used in that case

@ibgreen , what is the extension here, `GL_EXT_blend_minmax` ?

ibgreen:
Yes the idea is that in luma.gl the application can just use 
```
setParameter({
  blendEquation: GL.MAX
})
```
And under the hood, luma.gl will either use WebGL2 or the extension if available.

There is a also a feature 
`
hasFeature(BLEND_EQUATION_MINMAX)
`
That should be true if WebGL2 or the extension is available.

1chandu:
#308 added a unit test, remaining for this issue:
 For WebGL1 this new unit test passing could be false positive, verify if state set is successful, we could be just returning the value from cache.

1chandu:
I verified the calls running the unit test under WebGL1 context, when setting `Min` or `Max` values, we do detect change in cache and call the original `blendEquationSeparate/blendEquation` methods to update the state. So the only remaining thing here is to actual rendering.

ibgreen:
> So the only remaining thing here is to actual rendering.

I am not sure we should. We need to stop somewhere, our goal is not to create a complete validation suite for WebGL. We should test our classes, not WebGL itself. I suspect that we have many other more important holes in our test suite.

ibgreen:
> I verified the calls running the unit test under WebGL1 context, when setting Min or Max values, we do detect change in cache and call the original blendEquationSeparate/blendEquation methods to update the state.

Is that captured by a test case or through manual testing?

1chandu:
> Is that captured by a test case or through manual testing?

unit test that was added in #308

PR #308, opened 2017-09-14
Miscellaneous fixes: GLSL error reporting, BlendMinMax test, 'cross origin' image load

1chandu:
Fixes for #259 #273 #300 
Verified with test-browser, examples.

### DIFF
--- a/src/io/browser-load.js
+++ b/src/io/browser-load.js
@@ -26,9 +26,7 @@ export function loadFile(url, opts) {
  * image.crossOrigin can be set via opts.crossOrigin, default to 'anonymous'
  * returns a promise tracking the load
  */
-export function loadImage(url, opts = {
-  crossOrigin: 'anonymous'
-}) {
+export function loadImage(url, opts) {
   url = pathPrefix ? pathPrefix + url : url;
 
   return new Promise((resolve, reject) => {
@@ -36,7 +34,7 @@ export function loadImage(url, opts = {
       const image = new Image();
       image.onload = () => resolve(image);
       image.onerror = () => reject(new Error(`Could not load image ${url}.`));
-      image.crossOrigin = opts.crossOrigin;
+      image.crossOrigin = (opts && opts.crossOrigin) || 'anonymous';
       image.src = url;
     } catch (error) {
       reject(error);
--- a/src/webgl/shader.js
+++ b/src/webgl/shader.js
@@ -86,7 +86,7 @@ export class Shader extends Resource {
         parseGLSLCompilerError(infoLog, this.source, this.shaderType);
       log.error(0, `GLSL compilation errors in ${shaderName}\n${errors}`);
       log.warn(0, `GLSL compilation warnings in ${shaderName}\n${warnings}`);
-      throw new Error(`GLSL compilation errors in ${shaderName}\n${errors}\n${warnings}`);
+      throw new Error(`GLSL compilation errors in ${shaderName}\n${errors}`);
     }
   }
 
--- a/test/webgl/context-state.spec.js
+++ b/test/webgl/context-state.spec.js
@@ -1,3 +1,5 @@
+/* eslint max-depth: ["error", 4] */
+
 import test from 'tape-catch';
 
 import {GL} from 'luma.gl';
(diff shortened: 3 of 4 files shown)


src/io/browser-load.js
```diff
@@ -26,17 +26,15 @@ export function loadFile(url, opts) {
...
+export function loadImage(url, opts) {
   url = pathPrefix ? pathPrefix + url : url;
 
   return new Promise((resolve, reject) => {
     try {
       const image = new Image();
       image.onload = () => resolve(image);
       image.onerror = () => reject(new Error(`Could not load image ${url}.`));
-      image.crossOrigin = opts.crossOrigin;
+      image.crossOrigin = opts && opts.crossOrigin || 'anonymous';
```
ibgreen: Nit: I usually add parenthesis when using multiple operators of similar precedence to make it extra clear what order is intended:
```
image.crossOrigin = (opts && opts.crossOrigin) || 'anonymous';
```

src/webgl/shader.js
```diff
@@ -85,7 +85,6 @@ export class Shader extends Resource {
       const {shaderName, errors, warnings} =
         parseGLSLCompilerError(infoLog, this.source, this.shaderType);
       log.error(0, `GLSL compilation errors in ${shaderName}\n${errors}`);
-      log.warn(0, `GLSL compilation warnings in ${shaderName}\n${warnings}`);
```
ibgreen: This is not a fix, but rather a removal of the v4 feature of presenting warnings and errors with different "colors" in the console.

I believe that the real issue is that `parseGLSLCompilerError` includes too much text into one of the `errors` and `warnings` strings. If we could fix that problem there, then we could still show errors and warnings here with different "colors" as intended.

If you want to try to fix the error parser, the most easy way for me to reproduce this is when compiling a shader using the `fp64` module (just introduce any error to the main shader), fp64 module generates 
a number of warnings.

If you decide not to fix the parse, just comment this line out (instead of deleting it) and leave a comment that it should be restored when the parser is fixed.

test/webgl/context-state.spec.js
```diff
@@ -238,3 +240,54 @@ test('WebGLState#withParameters: recursive', t => {
...
+      [GL.BLEND_EQUATION_RGB]: GL.MAX,
+      [GL.BLEND_EQUATION_ALPHA]: GL.MIN
+    },
+    {
+      [GL.BLEND_EQUATION_RGB]: GL.MAX,
+      [GL.BLEND_EQUATION_ALPHA]: GL.MAX
+    }
+  ];
+
+  for (const contextName in contexts) {
```
ibgreen: I really like this test setup! 

I am however a little surprised that this works under Node.js as I suspect that the WebGL1 context provided by headless-gl does not support the extension. 

This makes me think we are getting a false positive. I suppose internally we just set the parameter without checking if it fails, and then just serve back the value from our cache, even though the true value is something else. 

To deal with this we probably would need an override of `gl.blendEquation` in the context polyfill.

Not sure that this complication is worth extra effort right now, we could address it in a future PR. Maybe just leave a comment in the test case about this.

src/webgl/shader.js
```diff
@@ -86,7 +86,7 @@ export class Shader extends Resource {
         parseGLSLCompilerError(infoLog, this.source, this.shaderType);
       log.error(0, `GLSL compilation errors in ${shaderName}\n${errors}`);
       log.warn(0, `GLSL compilation warnings in ${shaderName}\n${warnings}`);
-      throw new Error(`GLSL compilation errors in ${shaderName}\n${errors}\n${warnings}`);
```
1chandu: @ibgreen ,  warnings and errors are properly identified, but the`Error` thrown here includes both warnings and errors hence we are seeing warnings as errors. 
(Note: filtered to only errors)
Before:
screen shot 2017-09-14 at 4 07 53 pm
After:
screen shot 2017-09-14 at 4 07 29 pm

test/webgl/context-state.spec.js
1chandu: I will add a TODO, and keep #259 open.

src/webgl/shader.js
ibgreen: > warnings and errors are properly identified, but theError thrown here includes both warnings and errors hence we are seeing warnings as errors.

OK that makes sense. Yes definitely let's leave the warnings out of the Exception message. Thanks for clearing it up!

PR #309, opened 2017-09-14
Add "math.gl" dependency for website

jianhuang01:
After cleaning up all the dependencies under examples/lessons/10/node-modules, found the website testing failed for lesson10 because of missing "math.gl", figured out we need add "math.gl" dependency for website project.

### DIFF
--- a/examples/lessons/10/app.js
+++ b/examples/lessons/10/app.js
@@ -3,9 +3,7 @@ import {
   resetParameters, setParameters
 } from 'luma.gl';
 
-import {
-  Matrix4
-} from 'math.gl';
+import { Matrix4 } from 'math.gl';
 
 import {loadWorldGeometry, World} from './world';
 
--- a/examples/lessons/10/app.js
+++ b/examples/lessons/10/app.js
@@ -5,7 +5,7 @@ import {
 
 import {
   Matrix4
-} from 'math.gl'
+} from 'math.gl';
 
 import {loadWorldGeometry, World} from './world';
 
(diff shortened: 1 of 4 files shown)


website/webpack.config.js
```diff
@@ -46,7 +46,8 @@ const COMMON_CONFIG = {
 
   resolve: {
     alias: {
-      'luma.gl': libSources
+      'luma.gl': libSources,
+      'math.gl': join(__dirname, 'node_modules/math.gl')
```
jianhuang01: Not sure if this is the right way to expose math.gl to our examples, adding dependency in website/package.json only doesn't work, not super clear about the reason why it doesn't work.

examples/lessons/10/app.js
```diff
@@ -5,7 +5,7 @@ import {
 
 import {
   Matrix4
-} from 'math.gl'
+} from 'math.gl';
```
ibgreen: Nit: Let's keep this as a one line import. We want our examples to be fairly compact. The multiline style is suitable when importing lots of symbols from a module.
```
import {Matrix4} from 'math.gl'
```

examples/lessons/11/app.js
```diff
@@ -4,7 +4,7 @@ import {
 
 import {
   Matrix4
-} from 'math.gl'
+} from 'math.gl';
```
ibgreen: ditto

website/webpack.config.js
ibgreen: Hard to say what the problem is without testing. Maybe you could just add a TODO comment here about this, we can come back to it later.

PR #311, opened 2017-09-15
Add "lesson 13" example about fragment lighting and multiple programs

jianhuang01:
New "lesson 13" example to demonstrate  fragment based lighting vs vertex based lighting, and how to switch between multiple programs.


jianhuang01:
@ibgreen I used "git mv" keep history of crate.gif this time, just found moon.gif under old example folder has smaller size, so I downloaded new one to replace it and lost the track of history for it.

ibgreen:
> I used "git mv" keep history of crate.gif this time, just found moon.gif under old example folder has smaller size, so I downloaded new one to replace it and lost the track of history for it

OK, the history of bitmaps is not so important, and git will only store one copy if the files are identical. I do think this example needs a bigger texture, it shows the moon much more zoomed in.

examples/lessons/13/app.js
```diff
@@ -0,0 +1,294 @@
...
+  if (uUseTextures) {
+    fragmentColor = texture2D(uSampler, vec2(1.0 - vTextureCoord.s, 1.0 - vTextureCoord.t));
+  } else {
+    fragmentColor = vec4(1.0, 1.0, 1.0, 1.0);
+  }
+  gl_FragColor = vec4(fragmentColor.rgb * lightWeighting, fragmentColor.a);
+}
+`;
+
+const moonRotation = {
```
ibgreen: Why wrap this in an object? Can't we just do
```
const moonRotation = new Matrix4()....
```

examples/lessons/13/app.js
```diff
@@ -0,0 +1,294 @@
...
+  matrix: new Matrix4().rotateY(radians(180)).translate([2, 0, 0])
+};
+
+const cubeRotation = {
```
ibgreen: ditto

examples/lessons/13/app.js
```diff
@@ -0,0 +1,294 @@
...
+  },
+  onRender: ({
+    gl, tick, aspect, moon, cube, vertexLightingProgram, fragmentLightingProgram
+  }) => {
+    // set camera position
+    const eyePos = new Matrix4()
+    .rotateX(radians(-30))
+    .transformVector3([0, 0, 5]);
+
+    gl.clear(GL.COLOR_BUFFER_BIT | GL.DEPTH_BUFFER_BIT);
```
ibgreen: Nit; I'd move this line `gl.clear` up or down to keep eyePos and uVMatrix calculations together.

examples/lessons/13/app.js
jianhuang01: it doesn't work in Safari by using "const moonRotation = new Matrix4()....", cause global functions can't access const or let in Safari even I pass const or let as parameters to them. After wrapping it with an object, it works well.

examples/lessons/13/app.js
jianhuang01: same.

examples/lessons/13/app.js
jianhuang01: Done.

examples/lessons/13/app.js
ibgreen: > After wrapping it with an object, it works well.

Got it. In that case I would suggest using a single wrapper 
```
const appState = {
  moonRotation: new ...,
  cubeRotation: 
};
```
It would serve the purpose of holding all app state together rather than just being a confusing work-around for a browser issue.

Maybe update all examples with this approach?

jianhuang01: No problem. I will do it.

PR #312, opened 2017-09-15
Hook up website links for example "lesson 12" and "lesson 13"

jianhuang01:
Enable example links of "lesson 12" and "lesson 13", tested locally. Also updated thumb image of "lesson 13" to the right one.


jianhuang01:
@ibgreen the old thumb png for "lesson 13" is wrong, that's why I replaced it with the right one. The old thumb png is actually from "lesson 12". So no way to avoid that.

Issue #314, opened 2017-09-20
Problems in buffer.getData


cmsdkfz:
https://github.com/uber/luma.gl/blob/d98642a20a6a8237e89980662cd28542eeec4eae/src/webgl/buffer.js#L196

There seems to be two problems in this function:
1) The default ArrayBuffer is not a ArrayBufferView, and it will throw an Exception

2) This one is more subtle: it binds the buffer to COPY_READ_WRITE, but when calling getBufferSubData, the target is using "this.target", this might be different target! And it just silently read out the wrong data!

ibgreen:
@cmsdkfz Looks like you have found two valid bugs, thanks for reporting them! 

These will need to be fixed asap. We usually get a patch release published within a week, hopefully you can work around till then?

cmsdkfz:
Not blocking me, thanks for prompt feedback.

PR #317, opened 2017-09-21
Fix buffer.getData default parameters and target setting.

1chandu:
Fix for #314
- Optimize ArrayBufferView allocation when it is not provided.
- Add additional asserts to fire on invalid combination of parameters to `getBufferSubData` (WebGL documentation is very confusing, this should help our users.)
- Update docs for `copyData` and `getData` with simple use cases.
- Add unit tests.

### DIFF
--- a/docs/api-reference/webgl/buffer.md
+++ b/docs/api-reference/webgl/buffer.md
@@ -14,7 +14,7 @@ For additional information, see [OpenGL Wiki](https://www.khronos.org/opengl/wik
 | `delete` | Destroys buffer |
 | `subData` | Updates a subset of a buffer object's data store. |
 | `copyData` (WebGL2) | Copies part of the data of another buffer into this buffer |
-| `getData` (WebGL2) | Reads data from buffer (GPU) into an ArrayBuffer or SharedArrayBuffer. When no destination ArrayBufferView is provided, it optimally allocates enough memory to copy all eligible data.|
+| `getData` (WebGL2) | Reads data from buffer (GPU) into an `ArrayBufferView` or `SharedArrayBuffer`.|
 
 
 ## Usage
@@ -57,14 +57,38 @@ buffer.subData({})
 
 Copying data between buffers (WebGL2)
 ```js
-const buffer = new Buffer(gl, {bytes: 200})
-buffer.subData({offset: 20, data: new Float32Array([1, 2, 3])});
+const sourceBuffer = ...
+const destinationBuffer = ...
+
+// To copy 32 bytes from sourceBuffer to destinationBuffer
+destinationBuffer.copyData({sourceBuffer, size: 32});
+
+// To copy 32 bytes from sourceBuffer at 8 byte offset into
+// destinationBuffer at 16 byte offset.
+destinationBuffer.copyData({sourceBuffer,
+  readOffset: 8,
+  writeOffset: 16,
+  size: 32});
 ```
 
 Getting data from a buffer (WebGL2)
 ```js
 const buffer = ...;
-const data = buffer.getData({srcByteOffset: 32, length: 5});
+
+// To get all the data from buffer
+const data = buffer.getData();
+
+// To get all the data from buffer starting from byteOffset 8
+// into existing ArrayBufferView.
+const existingArray = ...
+const data = buffer.getData({dstData: existingArray, srcByteOffset: 8});
+// Maximum possible elements will be copied based buffer and dstData size.
+
+// To get 5 elements from source buffer starting from byteOffset 8
+// into existing ArrayBufferView starting from 3rd element position.
+const existingArray = ...
+const data = buffer.getData({dstData: existingArray, srcByteOffset: 8, dstOffset: 3, length: 5});
+
 ```
 
 Binding and unbinding a buffer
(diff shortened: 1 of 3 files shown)


docs/api-reference/webgl/buffer.md
```diff
@@ -13,8 +13,8 @@ For additional information, see [OpenGL Wiki](https://www.khronos.org/opengl/wik
 | `initialize` | Allocates and optionally initializes the buffer object's data store on GPU. |
 | `delete` | Destroys buffer |
 | `subData` | Updates a subset of a buffer object's data store. |
-| `copySubData` (WebGL2) | Copies part of the data of another buffer into this buffer |
-| `getSubData` (WebGL2) | Reads data from buffer (GPU) into an ArrayBuffer or SharedArrayBuffer. |
+| `copyData` (WebGL2) | Copies part of the data of another buffer into this buffer |
+| `getData` (WebGL2) | Reads data from buffer (GPU) into an ArrayBuffer or SharedArrayBuffer. When no destination ArrayBufferView is provided, it optimally allocates enough memory to copy all eligible data.|
```
ibgreen: Nit: Add back ticks around the three type names.

src/webgl/buffer.js
```diff
@@ -185,17 +185,56 @@ export default class Buffer extends Resource {
     return this;
   }
 
-  // WEBGL2 ONLY: Reads data from buffer into an ArrayBuffer or SharedArrayBuffer.
+  //
+  /**
+   * WEBGL2 ONLY: Reads data from buffer into an ArrayBufferView or SharedArrayBuffer.
+   *
+   * @param {ArrayBufferView} dstData= -ArrayBufferView to copy elements into,
```
ibgreen: I personally don't think it makes sense to document a method twice. This is a public method and we have explicit docs in our markdown. We should focus on improving those public docs instead of duplicating text here. 

That's why there in general is only a one line comment on public methods. That's been the intention, obviously some code is older so the code may not be 100% consistent.

src/webgl/buffer.js
```diff
@@ -185,17 +185,56 @@ export default class Buffer extends Resource {
...
     srcByteOffset = 0,
-    dstOffset = 0,
-    length = 0
+    dstOffset = 0
   }) {
-    // TODO optimize dstData according to offset and length
-    dstData = dstData || new ArrayBuffer(this.bytes);
+    assertWebGL2Context(this.gl);
+
+    // Use GL_COPY_READ_BUFFER to avoid disturbing other targets and locking type
```
ibgreen: This comment got detached from its code.

ibgreen: This is a lot of preliminary calculations, should we create a helper function to perform the work?

test/webgl/buffer.spec.js
```diff
@@ -97,3 +97,50 @@ test('WebGL#Buffer setData/subData', t => {
 
   t.end();
 });
+
+test('WebGL#Buffer getData', t => {
```
ibgreen: Can we add a case for `copyData` as well?

test/webgl/buffer.spec.js
```diff
@@ -97,3 +97,50 @@ test('WebGL#Buffer setData/subData', t => {
...
+  if (!gl2) {
+    t.comment('WebGL2 not available, skipping tests');
+    t.end();
+    return;
+  }
+
+  let data = new Float32Array([1, 2, 3]);
+  let buffer = new Buffer(gl2, {data});
+
+  let receivedData = buffer.getData({});
```
ibgreen: Can we add default `({...}={})` parameter to `getData` so that we don't need to use empty braces here?

test/webgl/buffer.spec.js
```diff
@@ -97,3 +97,50 @@ test('WebGL#Buffer setData/subData', t => {
...
+    length: 1
+  });
+  expectedData = new Float32Array([0, 3]);
+  t.deepEqual(expectedData, receivedData, 'Buffer.getData: with src/dst offsets and length successful');
+
+  data = new Uint8Array([128, 255, 1]);
+  buffer = new Buffer(gl2, {data});
+
+  receivedData = buffer.getData({});
+  t.deepEqual(data, data, 'Buffer.getData: Uint8Array + default parameters successful');
```
ibgreen: Ideally, we should add a test case that triggers on the `buffer.target` bug.

docs/api-reference/webgl/buffer.md
```diff
@@ -64,7 +64,7 @@ buffer.subData({offset: 20, data: new Float32Array([1, 2, 3])});
 Getting data from a buffer (WebGL2)
 ```js
 const buffer = ...;
-const data = buffer.getSubData({offset: 20, size: 10});
+const data = buffer.getData({srcByteOffset: 32, length: 5});
```
ibgreen: Maybe add a couple of `getData` parameter variations and add `// comments` in front of each `getData` example explaining what effect these "cryptic" parameters have.

1chandu:
> Did we want to address not saving buffer.data as well, or should that be a separate PR?

Created #318, lets do this in a different PR, addressed all other comments.

PR #319, opened 2017-09-26
Add framebuffer bindings to WebGL state manager.

### LINKED ISSUE #316: Add framebuffer bindings to WebGL state manager.
Using `withParameters` to recursively override framebuffers, but on exit it doesn't reset to previous, it always resets to null (the default framebuffer) - this breaks rendering badly.

We should integrate WebGL bindings (at least frame buffer bindings) into state management.

1chandu:
Fixes #316


src/webgl-utils/track-context-state.js
```diff
@@ -45,9 +45,10 @@ export const GL_STATE_SETTERS = {
 
   // SPECIFIC SETTERS
 
-  clearStencil: (setter, s) => setter({
-    [GL.STENCIL_CLEAR_VALUE]: s}
-  ),
+  bindFramebuffer: (setter, target, fb) => setter({
```
ibgreen: Per my read of the article I forwarded, it should be 
```
bindFramebuffer: (setter, target, fb) => {
   switch (target) {
   case GL.FRAMEBUFFER:  
      setter({
         [GL.READ_FRAMEBUFFER_BINDING]: fb,
         [GL.DRAW_FRAMEBUFFER_BINDING]: fb
      });
     break;
   case GL.DRAW_FRAMEBUFFER:  
      setter({[GL.DRAW_FRAMEBUFFER_BINDING]: fb});
     break;
  case GL.READ_FRAMEBUFFER:
    setter({[GL.DRAW_FRAMEBUFFER_BINDING]: fb});
    break;
  }
})
```

src/webgl/context-state.js
```diff
@@ -120,7 +120,7 @@ export function setParameters(gl, parameters) {
 export function withParameters(gl, parameters, func) {
   // assertWebGLContext(gl);
 
-  const {frameBuffer, nocatch = true} = parameters;
+  const {frameBuffer, readFramebuffer, nocatch = true} = parameters;
```
ibgreen: I think we can skip this.
The `readFramebuffer` is only used internally by framebuffer.blit and framebuffer.readPixels. 
I do not expect that the application would need to bind it like this.
Also for a patch release, we should avoid adding new functionality. We could leave a comment or open an issue if we are uncertain.

src/webgl/context-state.js
```diff
@@ -130,21 +130,20 @@ export function withParameters(gl, parameters, func) {
...
   pushContextState(gl);
 
   setParameters(gl, parameters);
 
   if (framebuffer) {
-    // TODO - was there any previously set frame buffer we need to remember?
     framebuffer.bind();
   }
 
+  if (readFramebuffer) {
```
ibgreen: Ditto, suggest we drop.

src/webgl-utils/track-context-state.js
1chandu: I moved this logic into `context-state.js`, which actually handles function style settings.

PR #321, opened 2017-09-27
Fix Framebuffer.clear

### LINKED ISSUE #315: framebuffer.clear broken
It calls `clear` with wrong parameters, no webgl context. 

We need a test case.

1chandu:
Fixes #315

### DIFF
--- a/src/webgl/framebuffer.js
+++ b/src/webgl/framebuffer.js
@@ -222,7 +222,7 @@ export default class Framebuffer extends Resource {
     this.gl.bindFramebuffer(GL_FRAMEBUFFER, this.handle);
 
     if (color || depth || stencil) {
-      clear({color, depth, stencil});
+      clear(this.gl, {color, depth, stencil});
     }
 
     drawBuffers.forEach((value, drawBuffer) => {
--- a/test/webgl/framebuffer.spec.js
+++ b/test/webgl/framebuffer.spec.js
@@ -1,7 +1,7 @@
 /* eslint-disable max-len */
 import test from 'tape-catch';
 import 'luma.gl/headless';
-import {GL, Framebuffer, Renderbuffer, Texture2D, clear} from 'luma.gl';
+import {GL, Framebuffer, Renderbuffer, Texture2D} from 'luma.gl';
 
 import {fixture} from '../setup';
 
@@ -183,7 +183,7 @@ function testFramebufferReadPixels(t, gl) {
   const clearColor = [1, 0.5, 0.25, 0.125];
   const expectedColor = [255, 128, 64, 32];
 
-  clear(gl, {framebuffer, color: clearColor});
+  framebuffer.clear({color: clearColor});
 
   const color = framebuffer.readPixels({
     x: 0,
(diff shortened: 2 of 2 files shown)


test/webgl/framebuffer.spec.js
```diff
@@ -183,7 +183,7 @@ function testFramebufferReadPixels(t, gl) {
   const clearColor = [1, 0.5, 0.25, 0.125];
   const expectedColor = [255, 128, 64, 32];
 
-  clear(gl, {framebuffer, color: clearColor});
+  framebuffer.clear({color: clearColor});
```
1chandu: This unit test is testing both, `clear` and `readPixels`, no new unit test needed.

src/webgl/framebuffer.js
```diff
@@ -222,7 +222,7 @@ export default class Framebuffer extends Resource {
     this.gl.bindFramebuffer(GL_FRAMEBUFFER, this.handle);
 
     if (color || depth || stencil) {
-      clear({color, depth, stencil});
+      clear(this.gl, {color, depth, stencil});
```
ibgreen: This handles the crash which was the primary objective of this PR.

The line on 232 worries me though `this.gl.bindFramebuffer(GL_FRAMEBUFFER, null);` - what if another framebuffer was bound? 

It will be a separate PR but we absolutely need a solution for this to shore up our framebuffer API. One way would be to wrap all framebuffer methods in a `withParameters` call, although that is a little heavy (e.g. on the call stack). As an alternative, we could make a bindFramebuffer wrapper that just returns previous value from cache, without querying the context.

test/webgl/framebuffer.spec.js
ibgreen: > This unit test is testing both, clear and readPixels, no new unit test needed.

One could argue that for clarity and completeness, it would be good to have a "named test case" for each Framebuffer method. While both method calls are indeed being exercised here, they are inside a big test case that focuses on another thing, and it is not clear to me that future maintainers will understand that these calls must be kept here lest we lose coverage.

Nit: Is there still a separate test for the global function `clear`, or was that lost in this change?

test/webgl/framebuffer.spec.js
```diff
@@ -290,3 +290,33 @@ test('WebGL2#Framebuffer texture attach and read', t => {
   t.end();
 });
 */
+
+test('WebGL2#Framebuffer blit', t => {
```
ibgreen: Awesome. 

Note for future PR: We should add a texture comparison helper method to validate the output.

test/webgl/framebuffer.spec.js
1chandu: I changed `clear` to `framebuffer.clear` in this unit test, but `clear` is being tested also given it is called from `framebuffer.clear`, we should also have `clear-spec.js`.

ibgreen: > I changed clear to framebuffer.clear in this unit test, but clear is being tested also given it is called from framebuffer.clear, we should also have clear-spec.js.

Exactly, the implementation of `framebuffer.clear` could change, so we shouldn't rely on that for coverage. 

We still have a lot of untested functions, and there is no way we can address it all in a single PR, but your current approach of adding a little bit more coverage than we need in each PR really helps narrow the gap!

src/webgl/framebuffer.js
1chandu: > As an alternative, we could make a bindFramebuffer wrapper that just returns previous value from cache, without querying the context.

I like this idea. Also I was trying to add a unit tests to cover `_setDrawBuffers` and ran into issues, some of the code in this method needs to be handled by `polyfill`, requires more changes and a different PR. I will create an umbrella task to document all these.

1chandu: #322

PR #325, opened 2017-10-02
Clean up shadowmap example (delete unused files)

ibgreen:


examples/core/shadowmap/app.js
```diff
@@ -149,7 +149,7 @@ const animationLoop = new AnimationLoop({
       uProjection: camProj,
       uShadowView: shadowView,
       uShadowProj: shadowProj,
-      uShadowMap: fbShadow.texture
+      uShadowMap: fbShadow
```
shaojingli: fbShadow.texture?

ibgreen: I am updating the uniforms system to accept framebuffer's directly, using the default color attachment. 

I decided to break out these files to keep the size of that PR down. I suppose if this lands first, this example will be briefly broken on master.

shaojingli: Got it. Stamped.

PR #333, opened 2017-10-05
Fix `TransformFeedback.isSupported` bug

Pessimistress:
The returned value (WebGL extension) is consistent with the behavior of other classes' `isSupported` methods, though we really should convert it to `Boolean` for a cleaner interface.

### DIFF
--- a/src/webgl/transform-feedback.js
+++ b/src/webgl/transform-feedback.js
@@ -9,7 +9,7 @@ const GL_TRANSFORM_FEEDBACK = 0x8E22;
 export default class TranformFeedback extends Resource {
 
   static isSupported(gl) {
-    return isWebGL2(gl) || gl.getExtension('OES_vertex_array_object');
+    return isWebGL2(gl) && gl.getExtension('OES_vertex_array_object');
   }
 
   static isHandle(handle) {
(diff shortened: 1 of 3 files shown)


src/webgl/transform-feedback.js
```diff
@@ -9,7 +9,7 @@ const GL_TRANSFORM_FEEDBACK = 0x8E22;
 export default class TranformFeedback extends Resource {
 
   static isSupported(gl) {
-    return isWebGL2(gl) || gl.getExtension('OES_vertex_array_object');
+    return isWebGL2(gl) && gl.getExtension('OES_vertex_array_object');
```
ibgreen: Yes this is incorrectly copy-pasted from vertex-array.js. But the change is not correct. TF has nothing to do with VAOs, it just depends on WebGL2 so the condition should just be:
```
static isSupported(gl) {
    return isWebGL2(gl);
```
In fact, the expression as written in the PR will never be true, since WebGL2 implementations are not allowed to support WebGL1 extensions for features that are available by default in WebGL2.
```
return isWebGL2(gl) && gl.getExtension('OES_vertex_array_object'); // Always false
```

PR #336, opened 2017-10-09
Fix browser test for TransformFeedback

Pessimistress:


test/webgl/transform-feedback.spec.js
```diff
@@ -22,7 +22,7 @@ test('WebGL#TransformFeedback constructor/delete', t => {
 
   t.throws(
     () => new TransformFeedback(),
-    /.*WebGLRenderingContext.*/,
+    (/.*Requires WebGL2.*/),
```
1chandu: So this needs to match the assert message, good to know.

PR #338, opened 2017-10-10
Picking module from deck.gl

1chandu:
- Copy enhanced picking module from deck.gl to luma.gl
- Fix documentation.

Fix for #890.
Next step:  Remove picking module from deck.gl. (Once this change is published in an alpha release)

Verified with, `test-browser`, `examples`.

### DIFF
--- a/examples/core/instancing/app.js
+++ b/examples/core/instancing/app.js
@@ -59,7 +59,7 @@ const animationLoop = new AnimationLoop({
 
     cube.updateModuleSettings({
       pickingSelectedColor: pickInfo && pickInfo.color,
-      pickingValid: pickInfo !== null
+      pickingSelectedColorValid: pickInfo !== null
     });
 
     gl.clear(GL.COLOR_BUFFER_BIT | GL.DEPTH_BUFFER_BIT);
--- a/src/core/model.js
+++ b/src/core/model.js
@@ -22,6 +22,10 @@ const ERR_MODEL_PARAMS = 'Model needs drawMode and vertexCount';
 
 const LOG_DRAW_PRIORITY = 2;
 
+// These old picking uniforms should be avoided and we should use picking module
+// and set uniforms using Model class 'updateModuleSettings()'
+const DEPRECATED_PICKING_UNIFORMS = ['renderPickingBuffer', 'pickingEnabled'];
+
 // Model abstract O3D Class
 export default class Model extends Object3D {
   constructor(gl, opts = {}) {
@@ -287,6 +291,7 @@ in a future version. Use shader modules instead.`);
 
   // TODO - should actually set the uniforms
   setUniforms(uniforms = {}) {
+    this._checkForDeprecatedUniforms(uniforms);
     checkUniformValues(uniforms, this.id);
     Object.assign(this.uniforms, uniforms);
     this.setNeedsRedraw();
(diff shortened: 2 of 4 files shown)


src/core/model.js
```diff
@@ -405,6 +410,17 @@ in a future version. Use shader modules instead.`);
     return this;
   }
 
+  // HELPER METHODS
+
+  _checkForDeprecatedUniforms(uniforms) {
```
1chandu: @ibgreen 
I am wondering do we have to do this, what if an app uses these uniforms for any other purpose?

examples/core/instancing/app.js
```diff
@@ -59,7 +59,7 @@ const animationLoop = new AnimationLoop({
 
     cube.updateModuleSettings({
       pickingSelectedColor: pickInfo && pickInfo.color,
-      pickingValid: pickInfo !== null
+      pickingSelectedColorValid: pickInfo !== null
```
ibgreen: I somewhat strongly feel that the `getUniforms` method should do this `null` check and set the uniforms accordingly, and save the application the trouble. The application would just set `pickingSelectedColor` to `null`.

It's kind of the purpose of `getUniforms` to help the app do the mapping from JavaScript friendly names to uniform names.

src/shadertools/modules/picking/picking.js
```diff
@@ -1,17 +1,25 @@
-const DEFAULT_HIGHLIGHT_COLOR = new Uint8Array([0, 64, 128, 64]);
+import {log} from '../../../utils';
```
1chandu: This is a copy paste of this file from deck.gl, which has been used for picking/highlighting for few months on master.

src/core/model.js
ibgreen: I think it is OK. It should just generate a single warning. Helps us catch any stragglers and we can remove in next release.

src/shadertools/modules/picking/README.md
```diff
@@ -32,22 +32,23 @@ main() {
...
 
 ### getUniforms
 
-`getUniforms` returns an object with key/value pairs representing the uniforms that the `picking` module shaders need.
+`getUniforms` takes an object takes a set of key/value pairs, returns an object with key/value pairs representing the uniforms that the `picking` module shaders need.
 
-`getUniforms({enabled, })`
+`getUniforms(opts)`
+opts can contain following keys:
+* `pickingValid` (*boolean*) - When true current instance picking color is ignored, hence no instance is highlighted.
```
ibgreen: API Audit mode on:
* Most modules have an `enabled` flag, allowing them to be turned off.
* Is this semantically different? If not, should we call it `pickingEnabled`?
* Why is picking ignored when valid is `true`?

examples/core/instancing/app.js
1chandu: In case of `autoHighlighting`, we may have non-null `pickInfo` (i.e. non numm pickingSelectedColor), but we should set `pickingSelectedColorValid` to false if highlighting is disabled, this can still be overloaded in JS `getUniforms`, but for this PR given I moving this module from luma.gl would like to keep the change minimal, to avoid any regressions in deck.gl. I will make a note in under API audits.

src/shadertools/modules/picking/README.md
1chandu: `pickingValid` is renamed to `pickingSelectedColorValid`, I just updated this doc too. Also it implied `pickingColor` is ignored when this flag (highlighting) is off.

Issue #340, opened 2017-10-12
Overwhelming console warnings


Pessimistress:
Deprecation warnings flood the console as they are issued every animation frame.
deck.gl issues unique warnings only once.

1chandu:
What are these warning?
Are you saying these warning getting issued for each frame or only once?

PR #341, opened 2017-10-12
Do not issue duplicate warnings

Pessimistress:
Warning messages were cached, but `log.warn` did not check cache before printing to console.

Fixes https://github.com/uber/luma.gl/issues/340.

### DIFF
--- a/src/utils/log.js
+++ b/src/utils/log.js
@@ -23,14 +23,14 @@ const log = {
   once(priority, arg, ...args) {
     if (!cache[arg]) {
       log.log(priority, arg, ...args);
+      cache[arg] = true;
     }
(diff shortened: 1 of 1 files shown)


src/utils/log.js
```diff
@@ -27,7 +27,9 @@ const log = {
     cache[arg] = true;
   },
   warn(arg, ...args) {
-    console.warn(`luma.gl: ${arg}`, ...args);
+    if (!cache[arg]) {
+      console.warn(`luma.gl: ${arg}`, ...args);
+    }
     cache[arg] = true;
```
1chandu: >cache[arg] = true;

Nit: should we move it inside the loop?

PR #356, opened 2017-11-18
Switch to buildkite

balthazar:


.buildkite/pipeline.yml
```diff
@@ -0,0 +1,26 @@
+steps:
+  - name: ":docker:"
+    plugins:
+      docker-compose#v1.5.2:
+        build: fusion-core
```
KevinGrandon: This should be `build: luma-gl`

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,26 @@
...
+  - wait
+
+  - name: ":eslint:"
+    command: "yarn lint"
+    plugins:
+      docker-compose#v1.5.2:
+        run: luma-gl
+
+  - name: ":hourglass:"
+    command: "sh /etc/init.d/xvfb && yarn test"
```
KevinGrandon: Consider making this a file so it's easier to reproduce locally. E.g., I want to be able to run the same command with something like: 

```
docker-compose run luma-gl .buildkite/test.sh
```

test/bench/node.js
```diff
@@ -21,6 +21,9 @@
 // Enables ES2015 import/export in Node.js
 require('reify');
 
+/* global global */
```
KevinGrandon: Is this change needed as part of this?

1chandu:
>This is great, nice work! Need to finish the process to get Buildkite fully enabled and we should be able to land this soon. (We could potentially land this earlier if we want to leave the .travis.yml file)

@balthazar , I am not an expert of this area, but as long as it doesn't block landing PRs I am fine landing it before Buildkite process is finished.

.buildkite/pipeline.yml
balthazar: ahhh failed again

test/bench/node.js
balthazar: yep

balthazar:
I would say my biggest concern is that people have to signup to buildkite in order to see the build status, which should be something public. I guess we could bring this up to them if they are open to the idea.

KevinGrandon:
> biggest concern is that people have to signup

This is actually something they're working on, and we should be able to have open build pages soon.

balthazar:
That's good to know, let's see next week how it goes about the security stuff, if it doesn't look like it will happen next week I'll add back the `travis.yml` and merge

PR #368, opened 2017-12-06
Enable Intel Tan shader WA for default GPU

1chandu:
* When GPU/Render info not avilable, apply Intel Tan shader workaround.
* I would also port this to 4.0-release branch.

Possible fix for https://github.com/uber/deck.gl/issues/1216

Verified with test-browser and examples.


1chandu:
Ok, will keep this open to track, created https://github.com/uber/luma.gl/pull/369 for 4.0-release branch.

1chandu:
We have confirmed reports that #369 fixed offset issues from several users, merging this one.

PR #369, opened 2017-12-06
Add Intel Tan Shader WA for default GPU

1chandu:
* When GPU/Render info not avilable, apply Intel Tan shader workaround.

Possible fix for https://github.com/uber/deck.gl/issues/1216

Verified with test-browser and examples.

### DIFF
--- a/src/shadertools/lib/platform-defines.js
+++ b/src/shadertools/lib/platform-defines.js
@@ -37,7 +37,7 @@ export function getPlatformShaderDefines(gl) {
     platformDefines += `\
 #define INTEL_GPU
 #define INTEL_FP64_WORKAROUND 1
-#define NVIDIA_EQUATION_WORKAROUND 1\n \
+#define NVIDIA_EQUATION_WORKAROUND 1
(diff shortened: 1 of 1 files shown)


src/shadertools/lib/platform-defines.js
```diff
@@ -37,7 +37,7 @@ export function getPlatformShaderDefines(gl) {
     platformDefines += `\
 #define INTEL_GPU
 #define INTEL_FP64_WORKAROUND 1
-#define NVIDIA_EQUATION_WORKAROUND 1\n \
```
1chandu: @ibgreen @shaojingli , not sure why we have this extra `\n \` at end of the line, could be causing any issues? removed.

1chandu:
>So we are also turning on the tan workaround for unknown gpu in default
Yeah, thats the plan, in cases when browser doesn't provide correct GPU string (WEBGL_debug_renderer_info is not supported) we will apply the WA, hopefully it fixes issues like https://github.com/uber/deck.gl/issues/1216

Issue #269, opened 2017-07-14
Model: render() and draw() have unused parameters.


1chandu:
Model.draw() and Model.render() take, attributes and samples, but these are never used. We either need to append them to existing ones or remove to avoid confusion.

CC: @ibgreen @shaojingli

PR #375, opened 2017-12-19
Fix `model.render` inconsistencies

Pessimistress:
- `draw` calls `this.render(uniforms, attributes, samplers)` without the `parameters` argument
- `render` never uses `attributes` and `samplers`
- `setProgramState` uses `this.samplers` which is never defined

https://github.com/uber/luma.gl/issues/269

Tested: node, browser, website examples

### DIFF
--- a/src/core/model.js
+++ b/src/core/model.js
@@ -105,6 +105,7 @@ export default class Model extends Object3D {
 
     // TODO - remove?
     this.buffers = {};
+    this.samplers = {};
     this.userData = {};
     this.drawParams = {};
     this.dynamic = false;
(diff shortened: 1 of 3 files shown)


src/core/model.js
```diff
@@ -334,13 +335,15 @@ export default class Model extends Object3D {
     return this;
   }
 
-  render(uniforms = {}, attributes = {}, parameters = {}, samplers = {}) {
+  render(uniforms = {}, attributes = {}, samplers = {}) {
```
1chandu: `parameters` was never part of documented API, so removing here is fine. We only documented `uniforms`.

Can you also update the doc, to include `attributes` and `samplers`. Also in `whats-new.md`.

Issue #372, opened 2017-12-07
Allow shader module's `getUniforms` callback access dependency uniforms


Pessimistress:
https://github.com/uber/deck.gl/pull/1219

1chandu:
Closing since uber/deck.gl#1219 got merged.

1chandu:
I think this is tracking to have more improved solution for #1219, re-opening it.

PR #379, opened 2018-01-05
Allow shaderModule.getUniforms access dependency uniforms

Pessimistress:
https://github.com/uber/luma.gl/issues/372


Pessimistress:
Similar to how `getUniforms` handle different `opts` input, they should expect the shape of the dependency uniforms to vary based on what's passed in `opts`. This behavior is actually intended because we only want to recalculate a uniform if the uniforms that it depends on are changed.

An example is the `project` and `project64` modules in deck.gl. When `opts.viewport` is provided, `project64` will receive the updated projection matrix generated by the `project` module. If `opts.viewport` is empty, then the `project` module generates nothing and so should `project64`.

PR #386, opened 2018-01-17
Bump webgl-debug module to 2.0.0

1chandu:


1chandu:
> QQ: Are we able to test debug mode at all? Is it turned on in all or part of our test suite?

We have a mix, some tests are using debug context when they use the fixture defined in `test/setup/index.js`, but some other tests are creating gl context without debug flag.

Issue #401, opened 2018-02-08
New Transform/GPGPU API


1chandu:
- [X] Add label and assign to milestone
  - [X] Coding
  - [X] Unit tests
  - [X] Documenation
  - [x] Whats new, link to website documentation.
- [x] Update luma.gl TF-instancing demo to use new API.
- [x] Update deck.gl wind-demo and attribute-transitions to use new API.
- [x] Update luma.gl `Model` documentation to specify `geometry` is optional.

PR #418, opened 2018-02-22
Transform: Add updateBuffers method

1chandu:
For #401

Add a new method, `updateBuffers` to handle the use cases , where one or more buffer bindings can be updated while keeping the others as is.

Updated unit tests and documentation.

Verified with unit tests, TF examples, and deck.gl attribute transitions.

### DIFF
--- a/docs/api-reference/core/experimental/transform.md
+++ b/docs/api-reference/core/experimental/transform.md
@@ -7,7 +7,7 @@ The `Transform` class provides easy interface to perform Transform Feedback oper
 | **Method**      | **Description** |
 | ---             | --- |
 | `constructor`   | creates a `Transform` object |
-| `bindBuffers`   | Sets up source and destination buffers |
+| `updateBuffers` | Update some or all buffer bindings |
 | `run`           | Performs one iteration of TransformFeedback |
 | `swapBuffers`   | Swaps source and destination buffers |
 | `getBuffer`     | Returns current destination buffer of given varying |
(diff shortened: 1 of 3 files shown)


docs/api-reference/core/transform.md
```diff
@@ -79,27 +79,47 @@ When `sourceDestinationMap` is specified buffers can be swapped using a single c
...
+
+Once `Transform` object is constructed and used, one or more source or destination buffers can be updated using `updateBuffers`.
+
+```js
+// transform is set up as above
+...
+
+// update buffer binding for 'inValue' attribute
+const newSourceBuffer = new Buffer(gl, {data: newSourceData});
+transform.updateBuffers({
```
1chandu: -> update(), chaining.

src/core/transform.js
```diff
@@ -60,63 +64,105 @@ export default class Transform {
...
 
-    // Append matching version string to FS.
-    let fs = PASS_THROUGH_FS;
-    const vsLines = vs.split('\n');
-    if (vsLines[0].indexOf('#version ') === 0) {
-      fs = `\
-${vsLines[0]}
-${PASS_THROUGH_FS}
-`;
+    for (const bufferName in sourceBuffers) {
```
1chandu: Object.assign instead of looping.

src/core/transform.js
```diff
@@ -60,63 +64,105 @@ export default class Transform {
...
 
+  // Swap source and destination buffers.
+  swapBuffers() {
+    assert(this._buffersSwapable);
+    this.currentIndex = (this.currentIndex + 1) % 2;
+  }
+
+  // Return Buffer object for given varying name.
+  getBuffer(varyingName = null) {
+    assert(varyingName && this.destinationBuffers[this.currentIndex][varyingName]);
```
1chandu: const destinationBuffers = this.destinationBuffers[this.currentIndex]

PR #419, opened 2018-02-24
Fix a crash when Model object created without geometry.

1chandu:
Fix crash, that happens when a `Model` object is created without a `Geometry` object.

More context:
All of our layers use geometry object, but for GPGPU features I am creating `Model` object without geometry (attributes, and vertex counts are updated after `Model` object is created).

Verified running unit tests.

### DIFF
--- a/src/core/model.js
+++ b/src/core/model.js
@@ -175,7 +175,9 @@ export default class Model extends Object3D {
     let redraw = false;
     redraw = redraw || this.needsRedraw;
     this.needsRedraw = this.needsRedraw && !clearRedrawFlags;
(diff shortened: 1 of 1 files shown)


src/core/model.js
```diff
@@ -532,7 +534,11 @@ count: ${this.stats.profileFrameCount}`
       const attributeTable = this._getAttributesTable({
         header: `${this.id} attributes`,
         program: this.program,
-        attributes: Object.assign({}, this.geometry.attributes, this.attributes)
+        attributes: Object.assign(
+          {},
+          this.geometry ? this.geometry.attributes : {},
```
Pessimistress: `this.geometry && this.geometrt.attributes` would work. `Object.assign` handles empty arguments.

PR #464, opened 2018-03-24
PickingModule: Add picking_filterColor method

1chandu:
#### Background
Add a convenience method in picking module , that can be used when highlighting enabled with no custom filters. 

#### Change List
- PickingModule: Add picking_filterColor method


1chandu:
Should I add an entry in whats-new, given it is one convenience method (no new functionality) ?

PR #491, opened 2018-04-09
Add verticalAxis opt to TruncatedConeGeometry

Pessimistress:
#### Background
`CylindarGeometry` and `ConeGeometry` are currently aligned with the y axis and cannot be rotated.
For example, deck.gl's HexagonLayer must manually rotate the cylindar geometry in vertex shader so that vertical is aligned with the z axis.

#### Change List
- Add `verticalAxis` option to `TruncatedConGeometry` constructor
- Update docs


Pessimistress:
A quick audit of existing geometry classes:
- `PlaneGeometry` has a `type` option that is one of `x,y`, `x,z` or `y,z`.
- `CubeGeometry` and `SphereGeometry` do not have this issue.
- `ConeGeometry` and `CylinderGeometry` both inherit this new option from `TruncatedConGeometry`.

Issue #318, opened 2017-09-22
buffer cleanup : Verify usage and remove buffer.data 


1chandu:
in `initialize` we are saving `data` to `this.data`,  seems this is not used anywhere, verify and remove.

1chandu:
@ibgreen @Pessimistress 
With this change we need to fix https://github.com/uber/deck.gl/issues/1974 before deck.gl can upgrade to version with this PR.

PR #588, opened 2018-06-30
Buffer: Remove reference to data

1chandu:
For #318

#### Change List
-  Buffer: Remove reference to data


1chandu:
@ibgreen @Pessimistress
With this change we need to fix uber/deck.gl#1974 before deck.gl can upgrade to version with this PR.

I tested all core-layers/experimental-layers using layer-browser and render-tests, only attribute transitions need this api.

PR #594, opened 2018-07-03
Fix: Uniform setting

1chandu:
For #

#### Background
Misc webgl fixes.
Create an array when setting single uniform.

#### Change List
- Fixes: Webgl and uniform setting.

### DIFF
--- a/src/webgl-context/context-state.js
+++ b/src/webgl-context/context-state.js
@@ -23,7 +23,7 @@ import assert from '../utils/assert';
 //   in an array, in a form that can be accepted by the setter.
 export const LUMA_SETTERS = {
   framebuffer: (gl, framebuffer) => {
-    const handle = framebuffer && framebuffer.handle ? framebuffer.handle : framebuffer;
+    const handle = framebuffer && 'handle' in framebuffer ? framebuffer.handle : framebuffer;
     return gl.bindFramebuffer(GL.FRAMEBUFFER, handle);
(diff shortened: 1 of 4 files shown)


1chandu:
Created this PR from @ibgreen private branch. 
Fixed an issue when setting single boolean uniform.

Over to @ibgreen to verify and land, I am not sure why we should convert single uniform values to arrays, is there test/demo failing without these changes?

1chandu:
Synced with @ibgreen offline, these changes are need when setting array type uniforms that are not vectors (like `uniform float fArray[5]). Added unit test that fail without this change.

src/webgl-context/context-state.js
```diff
@@ -23,7 +23,7 @@ import assert from '../utils/assert';
 //   in an array, in a form that can be accepted by the setter.
 export const LUMA_SETTERS = {
   framebuffer: (gl, framebuffer) => {
-    const handle = framebuffer && framebuffer.handle ? framebuffer.handle : framebuffer;
+    const handle = framebuffer && 'handle' in framebuffer ? framebuffer.handle : framebuffer;
```
Pessimistress: Add a comment on when `handle` would not exist?

1chandu: I think when `handle` itself is passed to the function as `framebuffer`, backward compatibility? CC: @ibgreen .

PR #728, opened 2018-09-14
Part-1: Transform texture support: add shader and texture utility methods

1chandu:
For #671

#### Background
Set of shader/texture utility methods needed for Transform texture support.

#### Change List
- Part-1: Transform texture support: add shader and texture utility methods

### DIFF
--- a/modules/core/src/webgl-utils/texture-utils.js
+++ b/modules/core/src/webgl-utils/texture-utils.js
@@ -1,0 +1,23 @@
+import Texture2D from '../webgl/texture-2d';
+import TextureCube from '../webgl/texture-cube';
+import Texture3D from '../webgl/texture-3d';
+import assert from 'assert';
+
+// Clone a new texture object from a reference texture object.
+export function cloneTextureFrom(refTexture, overrides) {
+  assert(
+    refTexture instanceof Texture2D ||
+    refTexture instanceof TextureCube ||
+    refTexture instanceof Texture3D
+  );
+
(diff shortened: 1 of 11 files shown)


modules/core/src/webgl-utils/texture-utils.js
```diff
@@ -0,0 +1,23 @@
...
+export function cloneTextureFrom(refTexture, overrides) {
+  assert(
+    refTexture instanceof Texture2D ||
+    refTexture instanceof TextureCube ||
+    refTexture instanceof Texture3D
+  );
+
+  const TextureType = refTexture.constructor;
+
+  const {gl, width, height, format, type, dataFormat, border, mipmaps} = refTexture;
```
1chandu: We can use this.constructor and move it to Texture class.

modules/core/src/webgl/framebuffer.js
```diff
@@ -272,14 +272,17 @@ export default class Framebuffer extends Resource {
...
+  } = {}) {
     const {gl} = this;
 
     // TODO - Set and unset gl.readBuffer
     if (attachment === GL.COLOR_ATTACHMENT0 && this.handle === null) {
       attachment = GL.FRONT;
     }
 
+    // Deduce the type from color attachment if not provided.
+    type = type || this.attachments[GL.COLOR_ATTACHMENT0].type;
```
1chandu: use `attachment` , add assert for valid attachments[]

Issue #743, opened 2018-10-02
Add messages for Transform Feedback examples when running under WebGL1


1chandu:
check: https://github.com/uber/deck.gl/issues/2252

PR #744, opened 2018-10-02
Add helpful string on canvas, for WebGL2 only demos

1chandu:
For #743

#### Background
For WebGL2 only demos: (Transform and Transform-feedback examples)
Remove `webgl1: false` to avoid app crash.
When running on webstie: add message to canvas
Standalone app: Add a useful log to console.

#### Change List
- Add helpful string on canvas, for WebGL2 only demos

### DIFF
--- a/website/src/react-demo-runner.js
+++ b/website/src/react-demo-runner.js
@@ -27,6 +27,16 @@ export default class DemoRunner extends Component {
     const {width, height, name, demo, noPanel, sourceLink} = this.props;
     const controls = demo.getInfo && demo.getInfo();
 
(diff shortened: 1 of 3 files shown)


website/src/react-demo-runner.js
```diff
@@ -27,6 +27,16 @@ export default class DemoRunner extends Component {
     const {width, height, name, demo, noPanel, sourceLink} = this.props;
     const controls = demo.getInfo && demo.getInfo();
 
+    const notSupported = demo.isNotSupported && demo.isNotSupported();
```
Pessimistress: Nit: `is*` functions usually are expected to return boolean. Make it two flags `webGL2Only` and `altText`?

PR #761, opened 2018-10-25
Perf: only set changed uniforms

Pessimistress:
#### Change List
- Only set changed uniforms

### DIFF
--- a/modules/core/src/webgl/program.js
+++ b/modules/core/src/webgl/program.js
@@ -145,6 +145,8 @@ export default class Program extends Resource {
         this.setUniforms(uniforms, samplers);
       }
 
+      this._bindTextures();
+
       if (framebuffer !== undefined) {
         parameters = Object.assign({}, parameters, {framebuffer});
       }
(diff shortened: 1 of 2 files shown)


modules/core/src/webgl/program.js
```diff
@@ -236,15 +260,8 @@ export default class Program extends Resource {
           const texture = uniform;
           const {textureIndex} = uniformSetter;
 
-          // TODO - this should be separated out from uniform setting, since it needs to be done
-          // before every draw even if uniforms have not changed
           texture.bind(textureIndex);
```
1chandu: Shouldn't we also remove `texture.bind(textureIndex);` , this is already done during `_bindTextures()` ?

Pessimistress: I tried it, looks like WebGL still needs *some* texture bound to the index to be able to set the uniform.

1chandu: Oh, that probably because `_bindTextures()` called afterwards, texture.bind() is not expensive so should be fine. Otherwise, we need 3 steps, extract changed uniforms, bind all textures, set changed uniforms.

Issue #764, opened 2018-11-05
Module split work


jianhuang01:
#### Background 
Current luma.gl repo has several submodules under thee module folder, but most of them have no correct package.json or webpack.config setup, can't be compiled with lerna. Here we need improve it by adding the right setup.
#### To Do List 
  - [x] Add label and assign to milestone 
  - [x] Coding
  - [x] Test

PR #765, opened 2018-11-05
module split work

jianhuang01:
For #764 

#### Change List
- Add package.json/webpack.config for each submodule
- Enable lerna build for each submodule
- Change sub-module names to follow our pattern
- Make folder structure the same for each submodule


modules/glfx/package.json
```diff
@@ -1,36 +1,26 @@
...
+  "main": "dist/es5/index.js",
+  "module": "dist/esm/index.js",
+  "esnext": "dist/es6/index.js",
   "files": [
-    "dist",
-    "lib",
-    "es",
-    "src"
+    "src",
+    "dist"
```
Pessimistress: Include `README.md`

PR #767, opened 2018-11-07
Fix offscreen rendering for Chrome 70

Pessimistress:
#### Change List
- Use the latest `AnimationLoop` API - `AnimationLoopProxy.createWorker` now takes a single `AnimationLoop` instance as argument
- Support mouse events

#### Core Example Test Results
- x cubemap  (accesses `document`)
- ✓ fragment
- ✓ instancing
- ✓ mandelbrot
- x picking  (`loadTextures` accesses `Image`)
- ✓ shadowmap
- ✓ transform
- ✓ transform-feedback


examples/core/cubemap/app.js
```diff
@@ -221,6 +221,6 @@ function drawTexture({ctx, sign, axis, size}) {
 export default animationLoop;
 
 /* global window */
-if (!window.website) {
+if (typeof window !== 'undefined' && !window.website) {
```
1chandu: Super nit: Create a util method given this is used in multiple places?

modules/core/src/core/animation-loop-proxy.js
```diff
@@ -106,12 +139,49 @@ export default class AnimationLoopProxy {
...
+      case 'removeEventListener':
+        this.canvas.removeEventListener(evt.data.type, this._onEvent);
+        break;
+
+      default:
+    }
+  }
+
+  _onEvent(evt) {
+    const devicePixelRatio = this.props.useDevicePixels ? (window.devicePixelRatio || 1) : 1;
```
1chandu: Is this code common to on-screen and off-screen? If it is only for offscreen, isn't `useDevicePixels` always false?

modules/core/src/core/animation-loop.js
```diff
@@ -213,7 +218,7 @@ export default class AnimationLoop {
     this.onRender(this.animationProps);
     // end callback
 
-    if (this.offScreen) {
+    if (this.offScreen && this.gl.commit) {
```
1chandu: Who sets this `this.gl.commit`, may be add a comment?

examples/core/cubemap/app.js
Pessimistress: This is per example as they need to work independently.

modules/core/src/core/animation-loop-proxy.js
Pessimistress: `useDevicePixels` is false on the worker thread but defined by user on the main thread.

modules/core/src/core/animation-loop.js
Pessimistress: https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/commit

Calling `gl.commit` is required in FF but not in Chrome 70.

PR #769, opened 2018-11-08
Dof example

tsherif:
#### Background
Created a Depth of Field example, which is a port of my PicoGL.js example: https://tsherif.github.io/picogl.js/examples/dof.html
Based on the original algorithm from: http://www.nutty.ca/?page_id=352&link=depth_of_field

I also added links to it through the example page of the website. Note that the example won't work through the website until it's merged because it's using the path to raw.githubusercontent.com for the textures, as is done in the other examples. If you want to run it in the website locally before merging, you can change the path to this one in `website/content/pages.js`: https://tsherif.github.io/picogl.js/examples/img/

I'll also note that some parts of implementing this were perhaps more difficult than they had to be and suggest ways in which Luma's API could be developed:
- A way to create a framebuffer with depth/stencil texture attachments, without explicitly creating the textures.
- A way to set up instanced matrix attributes without having to create 4 individual vec4 attributes. You can see I did this in PicoGL [here](https://github.com/tsherif/picogl.js/blob/master/src/vertex-array.js#L238-L263).

#### Change List
- Depth of Field example
- Links added to example page of website

luma-dof


examples/core/dof/app.js
```diff
@@ -0,0 +1,433 @@
...
+
+    ////////////////////////////////////////////////
+    // Set up frambuffers. 
+    ////////////////////////////////////////////////
+
+    // Need to ensure both color and depth targets can be sampled.
+    const mainFramebuffer = new Framebuffer(gl, {
+      width: gl.drawingBufferWidth,
+      height: gl.drawingBufferHeight,
+      attachments: {
```
1chandu: Nit: Looks like both color and depth attachments have standard options, `Framebuffer` object can create these when not specified, makes code cleaner.

examples/core/dof/app.js
```diff
@@ -0,0 +1,433 @@
...
+        [gl.TEXTURE_MAG_FILTER]: gl.LINEAR,
+        [gl.TEXTURE_MIN_FILTER]: gl.LINEAR_MIPMAP_NEAREST
+      }
+    }).then((textures) => {
+
+      ////////////////////////////////////////////
+      // Set up instanced model to draw boxes.
+      ////////////////////////////////////////////
+
+      const cubeGeo = new CubeGeometry().getAttributes();
```
1chandu: Instead of extracting all these details, can you subclass from `Cube` and add your customizations there? Like this : https://github.com/uber/luma.gl/blob/05263ca5245cffced569bc7e9c8d63ad1e3c78a2/examples/core/instancing/app.js#L16

examples/core/dof/app.js
```diff
@@ -0,0 +1,433 @@
...
+        attributes: {
+          positions: positionBuffer,
+          normals: normalBuffer,
+          uvs: uvBuffer,
+
+          // Attributes are limited to 4 components,
+          // So we have to split the matrices across
+          // 4 attributes. They're reconstructed in 
+          // the vertex shader.
+          modelMatCol1: {
```
1chandu: Looks like complicated setup. Can this be improved? Looks like all you need is position offset so each instance can be placed at a different offset, rest of the transformations can be constant (like rotation etc), so can you make a new `instancePositions` attribute and set rest of the matrix as uniform?

examples/core/dof/app.js
```diff
@@ -0,0 +1,433 @@
...
+      box.rotate[0] += 0.01;
+      box.rotate[1] += 0.02;
+      box.matrix.identity().translate(box.translate).rotateXYZ(box.rotate).scale(box.scale);
+      boxMatrices.set(box.matrix, i * 16);
+    }
+
+    matrixBuffer.setData(boxMatrices);
+
+    ////////////////////////////////////
+    // Draw boxes to main framebuffer
```
1chandu: Nit: `main framebuffer` comments is confusing, you are actully rendering to offscreen frame buffer here, so may be change this comment and names to `offscreenFramebuffer` or `sceneFramebuffer`.  `dofProgram` is actually rendering to main framebuffer.

examples/core/dof/app.js
```diff
@@ -0,0 +1,433 @@
...
+    texelOffset[0] = 0;
+    texelOffset[1] = 1;
+
+    dofProgram.setUniforms({
+        uTexelOffset: texelOffset,
+        uColor: mainFramebuffer.color,
+        uDepth: mainFramebuffer.depth
+    });
+
+    dofProgram.draw({
```
1chandu: Nit: wrap this into a model object?

tsherif:
Thanks for your comments, @1chandu. I'm made several updates based on your suggestions, and I'll respond to your individual comments with my reasoning in the places where I didn't.

examples/core/dof/app.js
```diff
@@ -0,0 +1,433 @@
...
+      vs: SCENE_VERTEX,
+      fs: SCENE_FRAGMENT
+    });
+
+    const dofProgram = new Program(gl, {
+      id: "DOF_PROGRAM",
+      vs: QUAD_VERTEX,
+      fs: DOF_FRAGMENT,
+      uniforms: {
+        uDepthRange: [0.1, 30],
```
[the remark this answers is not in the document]
tsherif: Done, but I did it as input sliders, since there are a few different parameters to manipulate.

examples/core/dof/app.js
tsherif: The default is to create a [depth Renderbuffer](https://github.com/uber/luma.gl/blob/master/modules/core/src/webgl/framebuffer.js#L654-L661) attachment, which wouldn't work for this because I need to sample depth.

examples/core/dof/app.js
tsherif: There's a translation as well as animating rotations about the X and Z axes, so that would be a different animation. More generally, the idea here is to show how to send an arbitrary transform matrix per instance. I also added a note to the PR description about how it might be worth it for Luma to support doing this directly through the API.

examples/core/dof/app.js
tsherif: Conceptually, I find it a bit awkward to think of a post processing screen-space pass as a "model". If it's really a break from Luma programming conventions, though, I can change it.

1chandu:
@tsherif , this is good to merge into master, may be you want to rebase first on current master and then merge?

tsherif:
@1chandu sure thing. I'm not familiar with a rebase workflow. I just tried to rebase dof-example off of master, but then git considers the branches to have diverged and I can't push. Do you have some documentation on how I should prep the branch for merging?

1chandu:
@tsherif, thanks for the rebase, merged.

Issue #750, opened 2018-10-08
[Shader Injection] declaration injection position


Pessimistress:
#### Actual Result
Currently, when injecting declarations (`vs:#decl`, `fs:#decl`) into shaders, the code is inserted at the beginning of the shader, see [this line](https://github.com/uber/luma.gl/blob/master/modules/core/src/shadertools/src/lib/inject-shader.js#L11).

If the user wants to inject a function definition, this new function cannot utilize any existing functions that are defined in the target shader, because the new function will be defined before everything else.

#### Expected Result

User injected declarations should be injected before the `main` function.

#### To Do List
  - [ ] Add label and assign to milestone
  - [ ] Coding
  - [ ] Test

1chandu:
#774 resolved this.

PR #774, opened 2018-11-15
Fixed injection location for function declaration in inject shader

XiaokaiUber:
For #750 

#### Background
Functions are injected at the beginning of the shader, which cannot utilize the functions already defined in the target shader. We need to move the injected function before the main function.

#### Change List
- Changed the location of the injected functions
- modified the test cases

### DIFF
--- a/modules/core/src/shadertools/src/lib/inject-shader.js
+++ b/modules/core/src/shadertools/src/lib/inject-shader.js
@@ -8,8 +8,8 @@ const MODULE_INJECTORS = {
   [FRAGMENT_SHADER]: MODULE_INJECTORS_FS
 };
 
-const REGEX_DECLARATIONS = /^(#version[^\n]*\n)?/; // Beginning of file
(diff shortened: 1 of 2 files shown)


modules/core/src/shadertools/src/lib/inject-shader.js
```diff
@@ -21,10 +20,10 @@ export default function injectShader(source, type, inject, injectStandardStubs)
   for (const key in inject) {
     const fragment = inject[key];
     switch (key) {
-    // declarations are injected at beginning of shader
+      // declarations are injected before the main function
     case 'vs:#decl':
       if (isVertex) {
-        source = source.replace(REGEX_DECLARATIONS, match => match + fragment);
+        source = source.replace(REGEX_START_OF_MAIN, match => fragment + match);
```
Pessimistress: Insert a line break after fragment for safety.

XiaokaiUber: @Pessimistress it's updated

Issue #755, opened 2018-10-12
AnimationLoop.setNeedsRedraw is broken


tuner:
`AnimationLoop.setNeedsRedraw(reason)` has no effect, as the reason does not reach the `onRender` callback. Only *drawing buffer resized* events work.

The reason appears to be that `animationProps.needsRedraw` gets overwritten with null in `_updateCallbackData`. See:

```
_updateCallbackData() {
    // Update redraw reason
    this.animationProps.needsRedraw = this.needsRedraw; <-- This is later overwritten
    this.needsRedraw = null; <-- with this

    ... height/width/aspect is checked here

    this.animationProps.width = width;
    this.animationProps.height = height;
    this.animationProps.aspect = aspect;

    this.animationProps.needsRedraw = this.needsRedraw; <-- oops

    ...
}
```

1chandu:
@tuner , thanks for reporting it. Can you please explain your use case in more detail, so this can be fixed.

PR #779, opened 2018-11-16
Fix needsRedraw

1chandu:
For #755

#### Background
Make `needsRedraw` flag setting behavior match with documentation.
Clear the flag after every render frame. If the flag was set by the application between the render frames, pass it to the `onRender` callback.

#### Change List
- Fix `needsRedraw` flag setting.

### DIFF
--- a/modules/core/src/core/animation-loop.js
+++ b/modules/core/src/core/animation-loop.js
@@ -182,6 +182,10 @@ export default class AnimationLoop {
 
   // PRIVATE METHODS
 
+  _clearNeedsRedraw() {
+    this.needsRedraw = null;
+  }
+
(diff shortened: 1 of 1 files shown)


modules/core/src/core/animation-loop.js
```diff
@@ -211,6 +215,9 @@ export default class AnimationLoop {
     this.onRender(this.animationProps);
     // end callback
 
+    // clear needsRedraw flag
+    this._clearNeedsRedraw();
```
Pessimistress: Nit: use `getNeedsRedraw(clearRedrawFlag = false)` to align with the rest of this codebase?

Issue #786, opened 2018-11-18
Texture.setImageData and setSubImageData error when using WebGL2 texture types


Pessimistress:
#### To Do List 
  - [x] Add label and assign to milestone
  - [x] Coding
  - [x] Test

PR #787, opened 2018-11-19
Fix Texture.setImageData and setSubImageData

Pessimistress:
For https://github.com/uber/luma.gl/issues/786

#### Change List
- Fix gl calls
- Add tests

### DIFF
--- a/modules/core/src/webgl/texture.js
+++ b/modules/core/src/webgl/texture.js
@@ -8,7 +8,6 @@ import {withParameters} from '../webgl-context/context-state';
 
 import {WebGLBuffer} from '../webgl-utils';
 import {isWebGL2, assertWebGL2Context} from '../webgl-utils';
-import {getTypedArrayFromGLType} from '../webgl-utils/typed-array-utils';
 
 import {log, uid, isPowerOfTwo} from '../utils';
 import assert from '../utils/assert';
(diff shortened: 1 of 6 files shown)


modules/core/src/webgl/texture.js
```diff
@@ -315,14 +315,14 @@ export default class Texture extends Resource {
...
-    height,
+    width = this.width,
+    height = this.height,
     level = 0,
-    format = GL.RGBA,
-    type,
-    dataFormat,
+    format = this.format,
+    type = this.type,
+    dataFormat = this.dataFormat,
```
gnavvy: Shall we align the naming with the [API spec](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/texImage2D) to prevent from mis-assignment?

`format` => `internalformat`
`dataFormat` => `format`

And update the jsDocs accordingly.

modules/core/src/webgl/texture.js
```diff
@@ -404,15 +409,15 @@ export default class Texture extends Resource {
...
```
gnavvy: same as above.

modules/core/src/webgl/texture.js
Pessimistress: That will be a breaking change. I will open an issue for v7.

Pessimistress: https://github.com/uber/luma.gl/issues/788

gnavvy:
ref: https://github.com/uber/luma.gl/issues/788

Issue #795, opened 2018-11-28
struct array uniform binding not work


jianhuang01:
#### Actual Result
struct PointLight {
 vec3 color;
 float intensity;
 vec3 position;
};
uniform PointLight lights[5]

can't set uniform data by specifying lights[0].color, light[0].intensity, etc.
#### Expected Result
should work fine
#### To Do List 
  - [ ] Add label and assign to milestone
  - [ ] Coding
  - [ ] Test

PR #796, opened 2018-11-28
add shortcut for uniform parse

jianhuang01:
For #795

#### Change List
- add shortcut when parsing uniforms

### DIFF
--- a/modules/core/src/webgl/uniforms.js
+++ b/modules/core/src/webgl/uniforms.js
@@ -118,7 +118,7 @@ export function parseUniformName(name) {
   if (name[name.length - 1] !== ']') {
     return {
       name,
-      length: name.length,
+      length: 1,
       isArray: false
     }
   }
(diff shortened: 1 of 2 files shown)


modules/core/src/webgl/uniforms.js
```diff
@@ -114,8 +114,14 @@ function toUIntArray(value, uniformLength) {
 }
 
 export function parseUniformName(name) {
-  // name = name[name.length - 1] === ']' ?
-  // name.substr(0, name.length - 3) : name;
+  // Shortcut to avoid redundant or bad matches
+  if (name[name.length - 1] !== ']') {
+    return {
+      name,
+      length: name.length,
```
Pessimistress: I believe `length` refers to the array length.

- What use case is this change trying to address?
- Add test cases to cover this scenario?

jianhuang01: - We doesn't use the length, don't know the initial purpose, for now it is not array length, for regular uniform like "position", the length is 1, but for array uniform like "position[0]", the length is "[0]".
- As mentioned in the issue ticket #795 , we need support struct array uniform like "lights[0].color", the current code change it to be "lights" which is not correct. Also for regular uniform, we don't need to do the regular expression match which is slower.
- Unit test is added

Issue #789, opened 2018-11-19
[picking module] Mix auto-highlight color with object color


Pessimistress:
When `pickingHighlightColor` is translucent, the highlight color should be mixed with the original object color rather than replacing it.

#### Actual Result
![image](https://user-images.githubusercontent.com/2059298/48740456-63a01b80-ec0c-11e8-9b25-54f3fb36b471.png)

#### Expected Result
![image](https://user-images.githubusercontent.com/2059298/48740566-cb566680-ec0c-11e8-84ef-939d512e7968.png)

#### To Do List 
  - [ ] Add label and assign to milestone
  - [ ] Coding
  - [ ] Test

PR #799, opened 2018-12-03
blend auto-highlight color and original color when an object is picked

XiaokaiUber:
For #789

#### Background
The original color of the object is replace by the auto-highlight color when the object is picked. We need to blend these two colors.

#### Change List
- blended the auto-highlight color with the original color of the object

picking_bug

### DIFF
--- a/modules/core/src/shadertools/src/modules/picking/picking.js
+++ b/modules/core/src/shadertools/src/modules/picking/picking.js
@@ -77,8 +77,18 @@ const float COLOR_SCALE = 1. / 255.;
  */
 vec4 picking_filterHighlightColor(vec4 color) {
   bool selected = bool(picking_vRGBcolor_Aselected.a);
-  vec4 highlightColor = selected ? (picking_uHighlightColor * COLOR_SCALE) : color;
-  return highlightColor * 0.3 + color * 0.7;
+
+  if (selected) {
+    vec4 highLightColor = picking_uHighlightColor * COLOR_SCALE;
+
+    float resultAlpha = highLightColor.a + color.a * (1.0 - highLightColor.a);
+
+    vec3 resultRGB = (highLightColor.rgb * highLightColor.a
+     + color.rgb * color.a * (1.0 - highLightColor.a)) / resultAlpha;
+    return vec4(resultRGB, color.a);
+  } else {
+    return color;
+  }
 }
 
 /*
--- a/modules/core/src/shadertools/src/modules/picking/picking.js
+++ b/modules/core/src/shadertools/src/modules/picking/picking.js
@@ -77,7 +77,8 @@ const float COLOR_SCALE = 1. / 255.;
  */
 vec4 picking_filterHighlightColor(vec4 color) {
   bool selected = bool(picking_vRGBcolor_Aselected.a);
-  return selected ? (picking_uHighlightColor * COLOR_SCALE) : color;
+  vec4 highlightColor = selected ? (picking_uHighlightColor * COLOR_SCALE) : color;
+  return highlightColor * 0.3 + color * 0.7;
 }
 
 /*
--- a/modules/core/src/shadertools/src/modules/picking/picking.js
+++ b/modules/core/src/shadertools/src/modules/picking/picking.js
@@ -81,10 +81,10 @@ vec4 picking_filterHighlightColor(vec4 color) {
   if (selected) {
     vec4 highLightColor = picking_uHighlightColor * COLOR_SCALE;
 
-    float resultAlpha = highLightColor.a + color.a * (1.0 - highLightColor.a);
+    float highLightAlpha = highLightColor.a;
+    float highLightRatio = highLightAlpha / (highLightAlpha + color.a * (1.0 - highLightAlpha));
 
-    vec3 resultRGB = (highLightColor.rgb * highLightColor.a
-     + color.rgb * color.a * (1.0 - highLightColor.a)) / resultAlpha;
+    vec3 resultRGB = mix(color.rgb, highLightColor.rgb, highLightRatio);
     return vec4(resultRGB, color.a);
   } else {
     return color;


modules/core/src/shadertools/src/modules/picking/picking.js
```diff
@@ -77,7 +77,8 @@ const float COLOR_SCALE = 1. / 255.;
  */
 vec4 picking_filterHighlightColor(vec4 color) {
   bool selected = bool(picking_vRGBcolor_Aselected.a);
-  return selected ? (picking_uHighlightColor * COLOR_SCALE) : color;
+  vec4 highlightColor = selected ? (picking_uHighlightColor * COLOR_SCALE) : color;
+  return highlightColor * 0.3 + color * 0.7;
```
jianhuang01: need use picking_uHighlightColor.a to blend this final color, do not hardcode the ratio like this.

1chandu: Also 1. Older behavior is as per design, if we are doing this, we should document this behavior, 2. I think we should only blend RGB and leave alpha as is, otherwise we are changing blending behavior with other layers.

jianhuang01: You may need use https://en.wikipedia.org/wiki/Alpha_compositing to blend the colors.

XiaokaiUber: The highlight RGB and and original RGB are blended with alpha compositing. The resulting alpha value is set the same as the original alpha value.

modules/core/src/shadertools/src/modules/picking/picking.js
```diff
@@ -77,7 +77,18 @@ const float COLOR_SCALE = 1. / 255.;
  */
 vec4 picking_filterHighlightColor(vec4 color) {
   bool selected = bool(picking_vRGBcolor_Aselected.a);
-  return selected ? (picking_uHighlightColor * COLOR_SCALE) : color;
+
+  if (selected) {
+    vec4 highLightColor = picking_uHighlightColor * COLOR_SCALE;
+
+    float resultAlpha = highLightColor.a + color.a * (1.0 - highLightColor.a);
```
1chandu: Nit : save `1.0 - highLightColor.a` in var and reuse.

modules/core/src/shadertools/src/modules/picking/picking.js
```diff
@@ -77,7 +77,18 @@ const float COLOR_SCALE = 1. / 255.;
...
+
+    vec3 resultRGB = (highLightColor.rgb * highLightColor.a
+     + color.rgb * color.a * (1.0 - highLightColor.a)) / resultAlpha;
+    return vec4(resultRGB, color.a);
```
1chandu: Should we use `highlightColor.a` as final alpha instead of `color.a`. Reason being, this only affects the object that is currently picked, so we let highlightColor define what its alpha should be. For instance, if user sets highlightColor.a = 1.0, we make highlighted object opaque ?   Other than this concern, all look good to me.

ibgreen: > Should we use highlightColor.a as final alpha instead of color.a?

Makes sense to me. Either way I think we should document the chosen behavior, in the `picking` shader module docs and probably also in the deck.gl docs.

A test case would also be really cool, not sure if that is feasible with new the shader test setups, need to take a closer look at that...

Pessimistress: I don't think `highlightColor.a` should be used. Say I only want to "slightly tint" the object when it's highlighted, `highlightColor.a` is the only way I can control that, but I don't want it to become translucent.

This is how kepler is implementing its own picking module:
https://github.com/uber/kepler.gl/blob/master/src/shaderlib/picking-module.js#L41

XiaokaiUber: In the following two images, I set the highlight color to [0, 255, 255, 50]. The first image uses the alpha value of the original object. When the planet is picked, the blending result can show the original object with slight highlight color.  The second image uses the alpha value of the highlight color. After blending, the planet becomes almost transparent because its alpha value is 50 now.

object_samll_alpha

highlight_samll_alpha

XiaokaiUber: highlight_samll_alpha

jianhuang01: return vec4(resultRGB, resultAlpha);

Pessimistress: `color.a` actually works better. In one use case we create arrows along a path by manipulating alpha:

image

The hollow part is still pickable, and on highlight it should not be filled.

jianhuang01: Good point, then better to change vec3 resultRGB = (highLightColor.rgb * highLightColor.a + color.rgb * color.a * (1.0 - highLightColor.a)) / resultAlpha;
to 
vec3 resultRGB = (highLightColor.rgb * highLightColor.a + color.rgb * color.a * (1.0 - highLightColor.a)) / color.a; and handle color.a case

Pessimistress: Mathematically this commit is correct. The way it is written is very confusing. It is equivalent to:

```glsl
float highLightAlpha = highLightColor.a;
float highLightRatio = highLightAlpha / (highLightAlpha + color.a * (1.0 - highLightAlpha));
vec3 resultRGB = mix(color.rgb, highLightColor.rgb, blendRatio);
```

Pessimistress: To be honest I think this is totally fine too:

```glsl
vec3 resultRGB = mix(color.rgb, highLightColor.rgb, highLightColor.a);
```

1chandu: > vec3 resultRGB = mix(color.rgb, highLightColor.rgb, highLightColor.a)

Sounds good.

On whether to use actual alpha or highlight alpha, each has its side effects. When using highlight.alpha we might end up making things opaque that are supposed to transparent (like arrow), but when using actual alpha, and if it is very low, you might not notice it is highlighted (defeats purpose of highlighting). At least using highlight.alpha you can change the behavior by changing highlight color prop, but when using actual alpha you don't have that control. I would still vote for hightlight.alpha :) , but using actual alpha is not a blocker.

jianhuang01: That's why better to calculate resultRGB by dividing color.a, which can guarantee hightlight color's RGB component is preserved in final rendering color. Since our rendering system is not based on premultiplied alpha, need handle divided by 0 case with this solution, which is not ideal.

danmarshall:
Hello, I just wanted to make everyone aware that this change has broken my use case, which I believe is compelling enough for consideration. My app has invisible polygons that I use as interactive hotspots. When the user mouses around, they can get a highlight over a certain region. The polygons are invisible so that they do not occlude any part of the rendering, or the background of the page.

Can we add a configuration to re-enable this kind of ability? Perhaps a `.getHighlightColor()` method or a `.mixHighlightColor = true` property on the PolygonLayer? Just suggestions.

I've been using this capability since v5.

PR #856, opened 2019-01-18
add new lighting glsl API and better check for lighting_uEnabled.

jianhuang01:
For #852 

#### Background
- Add better check for lighting_uEnabled, so we can turn off lighting when there is available light sources or material.
- When integrating with mesh layer, found that surface color is only available in fragment shader while our current lighting has to be in vertex shader, so adding a new API to handle that problem, in long term we need move lighting to fragment shader for better results.

#### Change List
- Better check for lighting_uEnabled
- New GLSL API(lighting_getSpecularLightColor)

### DIFF
--- a/modules/shadertools/src/modules/phong-lighting/phong-lighting.glsl.js
+++ b/modules/shadertools/src/modules/phong-lighting/phong-lighting.glsl.js
@@ -43,13 +43,40 @@ vec3 lighting_getLightColor(vec3 surfaceColor, vec3 light_direction, vec3 view_d
     return (lambertian * lighting_uDiffuse * surfaceColor + specular * lighting_uSpecularColor) * intensity;
 }
 
-vec3 lighting_getLightColor(vec3 surfaceColor, vec3 position_worldspace, vec3 normal_worldspace) {
+vec3 lighting_getLightColor(vec3 surfaceColor, vec3 cameraPosition, vec3 position_worldspace, vec3 normal_worldspace) {
   vec3 lightColor = surfaceColor;
 
   if (lighting_uEnabled) {
-    vec3 camera_pos_worldspace = project_uCameraPosition;
-    vec3 view_direction = normalize(camera_pos_worldspace - position_worldspace);
+    vec3 view_direction = normalize(cameraPosition - position_worldspace);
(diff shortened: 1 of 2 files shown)


modules/shadertools/src/modules/phong-lighting/phong-lighting.js
```diff
@@ -51,10 +54,21 @@ function getMaterialUniforms(material) {
...
+    )
+  ) {
+    return {};
+  }
+
   const {ambientLight, pointLights, directionalLights, material} = opts;
 
-  if (!(ambientLight || pointLights || directionalLights) || !material) {
-    return {};
+  if (!(ambientLight || (pointLights && pointLights.length > 0) || (directionalLights && directionalLights.length > 0)) || !material) {
```
Pessimistress: Too many parenthesis and negates. Make a variable e.g.

```js
const hasLights = ambientLight || (pointLights && pointLights.length > 0) || (directionalLights && directionalLights.length > 0);
```

modules/shadertools/src/modules/phong-lighting/phong-lighting.glsl.js
```diff
@@ -43,12 +43,11 @@ vec3 lighting_getLightColor(vec3 surfaceColor, vec3 light_direction, vec3 view_d
     return (lambertian * lighting_uDiffuse * surfaceColor + specular * lighting_uSpecularColor) * intensity;
 }
 
-vec3 lighting_getLightColor(vec3 surfaceColor, vec3 position_worldspace, vec3 normal_worldspace) {
+vec3 lighting_getLightColor(vec3 surfaceColor, vec3 cameraPosition, vec3 position_worldspace, vec3 normal_worldspace) {
   vec3 lightColor = surfaceColor;
 
   if (lighting_uEnabled) {
-    vec3 camera_pos_worldspace = project_uCameraPosition;
```
Pessimistress: How did this uniform get in there? Double check we don't have other dependencies on deck.gl's `project` module?

jianhuang01: The base code is from our old lighting code, didn't notice this uniform before. No other dependencies, only this uniform now.

PR #857, opened 2019-01-19
Add GLTF example with GLTF-loader working

georgios-uber:


examples/gltf/app.js
```diff
@@ -0,0 +1,80 @@
...
+export const animationLoopOptions = {
+
+  onInitialize: ({gl}) => {
+
+    const gltfParser = new GLTFParser();
+    fetch(GLTF_URL).then(res => res.arrayBuffer()).then(data => {
+      gltfParser.parse(data, {createImages: false}); // see https://github.com/uber-web/loaders.gl/issues/46
+      const instantiator = new GLTFInstantiator(gl, {
+        getBuffer: accessor => {
+          // I hope Ib can save us from this terrible hack!
```
georgios-uber: @ibgreen do you have a better way to access the unpackedBuffers?

modules/core/src/scenegraph/scenegraph-node.js
```diff
@@ -40,6 +40,10 @@ export default class ScenegraphNode {
     return this;
   }
 
+  setMatrix(matrixArray) {
+    this.matrix.copy(matrixArray);
```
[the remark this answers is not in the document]
georgios-uber: I prefer it that way because the **Matrix** is mutable and can result in hard-to-debug issues.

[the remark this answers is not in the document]
ibgreen: > I prefer it that way because the Matrix is mutable and can result in hard-to-debug issues.

True, though the perf hit can be very significant, and not giving the user the option to do it performantly is not good.

One way out is to add a second argument that controls whether to copy or set, or just add a separate method `copyMatrix`.

examples/gltf/app.js
```diff
@@ -0,0 +1,86 @@
...
+    const models = [];
+
+    const gltfParser = new GLTFParser();
+    fetch(GLTF_URL).then(res => res.arrayBuffer()).then(data => {
+      gltfParser.parse(data);
+      const instantiator = new GLTFInstantiator(gl, {
+        getImage: index => {
+          const img = gltfParser.getImage(index).image;
+          const promise = new Promise(resolve => {
+            img.onload = () => {
```
ibgreen: Nits: skip variable and braces: `return new Promise(resolve => resolve(...`

examples/gltf/app.js
```diff
@@ -0,0 +1,86 @@
...
+export const animationLoopOptions = {
+
+  onInitialize: ({gl}) => {
+    const models = [];
+
+    const gltfParser = new GLTFParser();
+    fetch(GLTF_URL).then(res => res.arrayBuffer()).then(data => {
+      gltfParser.parse(data);
+      const instantiator = new GLTFInstantiator(gl, {
+        getImage: index => {
```
ibgreen: Shouldn't this image handling be built in to the GLTFInstantiator, rather than in the example?

examples/gltf/app.js
```diff
@@ -0,0 +1,86 @@
...
+              resolve(new Texture2D(gl, { data: img, format: GL.RGB }));
+            };
+          });
+
+          return promise;
+        }
+      });
+      const scene = gltfParser.getScene(0);
+      const lumaScene = instantiator.createScene(gltfParser.getScene(0));
+      console.log("gltfParser.getScene(0) = ", scene);
```
ibgreen: I think we can comment out or drop this logging?

examples/gltf/app.js
```diff
@@ -0,0 +1,86 @@
...
+      console.log("gltfParser = ", gltfParser);
+      console.log("lumaScene = ", lumaScene);
+
+      lumaScene.traverse(node => {
+        console.log("N=", node);
```
ibgreen: Ditto

examples/gltf/app.js
```diff
@@ -0,0 +1,86 @@
...
+
+  onRender: ({gl, tick, width, height, aspect, models}) => {
+    gl.viewport(0, 0, width, height);
+    clear(gl, {color: [0, 0, 0, 1], depth: true});
+
+    const camView = new Matrix4().lookAt({eye: [0, 0.8, 1], center: [0, 0, 0], up: [0, 1, 0]}).rotateXYZ([0, tick * 0.01, 0]);
+    const camProj = new Matrix4().perspective({fov: radians(5), aspect, near: 0.01, far: 5000});
+
+    models.forEach(model => {
+      // console.log("RENDER", model);
```
ibgreen: Cut

examples/gltf/app.js
```diff
@@ -0,0 +1,86 @@
...
+      models
+    };
+  },
+
+  onRender: ({gl, tick, width, height, aspect, models}) => {
+    gl.viewport(0, 0, width, height);
+    clear(gl, {color: [0, 0, 0, 1], depth: true});
+
+    const camView = new Matrix4().lookAt({eye: [0, 0.8, 1], center: [0, 0, 0], up: [0, 1, 0]}).rotateXYZ([0, tick * 0.01, 0]);
+    const camProj = new Matrix4().perspective({fov: radians(5), aspect, near: 0.01, far: 5000});
```
ibgreen: Nit: We don't use abbreviated names in the code base.

examples/pbr/controls.js
```diff
@@ -21,7 +21,7 @@ class Controls {
   resetCamera() {
     this.roll = Math.PI;
     this.pitch = 0.0;
-    this.translate = 4.0;
+    this.translate = 2.5;
```
ibgreen: Should this be a different PR?

modules/core/src/scenegraph/base-model.js
```diff
@@ -329,6 +329,40 @@ export default class BaseModel extends ScenegraphNode {
       vs = vs || MODULAR_SHADERS.vs;
       fs = fs || MODULAR_SHADERS.fs;
 
+      vs = `
```
ibgreen: Can this go into mesh-model for now?

modules/core/src/scenegraph/gltf-instantiator/gltf-instantiator.js
```diff
@@ -1,13 +1,13 @@
...
+import Model from '../model';
 
 // GLTF instantiator for luma.gl
 // Walks the parsed and resolved glTF structure and builds a luma.gl scenegraph
 export default class GLTFInstantiator {
-  constructor(gl) {
+  constructor(gl, options = {}) {
     this.gl = gl;
-    this.options = {
-      removeEmptyGroups: true
```
ibgreen: Let's define a top-level constant with default options and merge them here.

```
this.options = Object.assign({}, DEFAULT_OPTIONS, options);
```

This way we can list all options in one place.

BTW, I think avoiding building unnecessary groups makes sense, not sure why we'd cut that. Otherwise the generated scenegraph can be unnecessarily deep and "verbose".

modules/core/src/scenegraph/gltf-instantiator/gltf-instantiator.js
```diff
@@ -27,75 +27,95 @@ export default class GLTFInstantiator {
...
   createPrimitive(gltfPrimitive, i, gltfMesh) {
     const attributes = this.createAttributes(gltfPrimitive.attributes, gltfPrimitive.indices);
     // TODO - handle gltfPrimitive.material
-    return new Model(this.gl, {
+    const model = new Model(this.gl, {
       id: gltfPrimitive.name || `${gltfMesh.name || gltfMesh.id}=${i}`,
-      attributes,
-      drawMode: gltfPrimitive.mode
+      drawMode: gltfPrimitive.mode || 4,
+      vertexCount: 2046 // from indices count
```
ibgreen: Huh? We can't hardcode this...

modules/core/src/scenegraph/gltf-instantiator/gltf-instantiator.js
```diff
@@ -27,75 +27,95 @@ export default class GLTFInstantiator {
...
     });
+    model.setProps({attributes});
+    this.options.getImage(0).then(img => model.setUniforms({ tex: img }));
```
ibgreen: I think it would be nice to have a single function that resolves all promises in a loaded glTF, rather than having to wait for them one by one like this.

That way a loaded file could be processed without dealing with promises. Of course if we want to support incremental loading, we'd need to do it like this.

The promise resolving function could be extended to handle URI loading etc.

modules/core/src/scenegraph/gltf-instantiator/gltf-instantiator.js
```diff
@@ -27,75 +27,95 @@ export default class GLTFInstantiator {
...
+    }
+    console.log("attributes>>", attributes, result);
+    return result;
   }
 
   createBuffer(accessor) {
-    accessor = this.createAccessor(accessor);
-    return new Buffer(this.gl, accessor).setData();
+    const opts = this.createAccessor(accessor);
+    return new Buffer(this.gl, opts);
```
ibgreen: Some TODO here as we want to create only one `Buffer` per `bufferView`?

modules/core/src/scenegraph/gltf-instantiator/gltf-instantiator.js
```diff
@@ -27,75 +27,95 @@ export default class GLTFInstantiator {
...
   }
 
   createAccessor(accessor) {
-    return null;
-    /*
-    return new Accessor({
+    console.log("AAAA",accessor)
```
ibgreen: Cut

modules/core/src/scenegraph/gltf-instantiator/gltf-instantiator.js
```diff
@@ -27,75 +27,95 @@ export default class GLTFInstantiator {
...
+    if (accessor.bufferView.byteStride) {
+      console.log("bufferView has byteStride, see https://github.com/uber-web/loaders.gl/issues/45");
+    }
+
+    return {
       type: accessor.componentType,
-      size: accessor.components,
+      // size: accessor.count, ????
```
ibgreen: Cut?

modules/core/src/scenegraph/group.js
```diff
@@ -47,7 +47,7 @@ export default class Group extends Node {
       if (child instanceof Group) {
         result = child.traverse(visitor, {modelMatrix});
       } else {
-        // child.setUniforms({modelMatrix});
+        child.setMatrix(modelMatrix); // TODO: can child has his own matrix?
```
ibgreen: Yes every scenegraph node has its own matrix. What we need to do is to multiply them together as we descend and pass the current matrix to `traverse(child, parentMatrix)`

modules/core/src/scenegraph/mesh-model.js
```diff
@@ -426,6 +426,32 @@ export default class MeshModel extends Node {
       vs = vs || MODULAR_SHADERS.vs;
       fs = fs || MODULAR_SHADERS.fs;
 
+      vs = `
```
ibgreen: Add TODO to remove this

modules/core/src/scenegraph/model.js
```diff
@@ -217,7 +217,7 @@ export default class Model extends BaseModel {
 
   // PRIVATE METHODS
 
-  _createProgram({
+  _createProgramXXX({
```
ibgreen: Undo

examples/gltf/app.js
georgios-uber: So what should the parameters of GLTFInstantiator be?

examples/gltf/app.js
georgios-uber: I copied this from the shadow example...

examples/pbr/controls.js
georgios-uber: OK

modules/core/src/scenegraph/base-model.js
georgios-uber: Need to decide if we will have a dedicated shader for glTF.

modules/core/src/scenegraph/gltf-instantiator/gltf-instantiator.js
georgios-uber: I think we will have more edge-cases and will be harder to debug.

examples/gltf/app.js
ibgreen: > So what should the parameters of GLTFInstantiator be?

I guess we'll discover that as we complete the implementation. In general, the fewer options the better.

modules/core/src/scenegraph/gltf-instantiator/gltf-instantiator.js
ibgreen: > I think we will have more edge-cases and will be harder to debug.

Well, just to explain where I am coming from, I noticed is that in many cases we'll just end up wrapping a `Mesh` (or `MeshModel` or ...) in a `Group`. 

```
Group
  | --------|
Group     Group
  |         | 
Mesh       Mesh
```

This looked unnecessary to me, and it seemed to me that simplifying this by dropping (i.e. not generating) the middle layer of groups shouldn't make debugging harder.

My intention was not to do any complex scenegraph transformations but just be a little efficient in avoiding generation of unnecessary nodes.

modules/core/src/scenegraph/gltf-instantiator/gltf-instantiator.js
georgios-uber: But then I have to change the loader.

PR #914, opened 2019-02-17
glTF: Add PBR debug options and ability to drag-and-drop GLB files

georgios-uber:
For #

#### Change List
-


modules/core/src/scenegraph/gltf-instantiator/gltf-instantiator.js
```diff
@@ -17,7 +17,8 @@ const ATTRIBUTE_TYPE_TO_COMPONENTS = {
 };
 
 const DEFAULT_OPTIONS = {
-  modelOptions: {}
+  modelOptions: {},
+  pbrDebug: false
```
Pessimistress: Add to documentation

georgios-uber: Cannot find the proper .md file. Will do later.

PR #918, opened 2019-02-19
Expose program readiness when using async textures

Pessimistress:
#### Background
https://github.com/uber/luma.gl/pull/911#discussion_r257047484

#### Change List
- `Model.draw` and `Program.draw` returns `false` if resources are not fully loaded
- Do not clear `Model.needsRedraw` flag if draw call was not successful
- Add async texture examples to the render test


modules/core/src/scenegraph/base-model.js
```diff
@@ -248,11 +248,13 @@ export default class BaseModel extends ScenegraphNode {
...
     this.onAfterRender();
 
-    this.setNeedsRedraw(false);
+    if (didDraw) {
+      this.setNeedsRedraw(false);
+    }
 
     this._logDrawCallEnd(logPriority, vertexArray, framebuffer);
 
-    return this;
```
tsherif: Not a blocking issue, but returning `this` seems to be a common idiom in Luma. Is method chaining used that often? If not, it could be worth it to stop returning `this` everywhere (since it doesn't seem to happen consistently anyway, e.g. [here](https://github.com/uber/luma.gl/blob/master/modules/core/src/webgl/classes/buffer.js#L121) vs [here](https://github.com/uber/luma.gl/blob/master/modules/core/src/webgl/classes/buffer.js#L135)).

Pessimistress: I think it was the pattern the design was going for, but in practice we are not chaining much.

PR #919, opened 2019-02-20
[test-utils] Support async onInitialized in TestRenderer

Pessimistress:
Align with `AnimationLoop`'s behavior.

#### Change List
- Allow `testCase.onInitialized` to return a promise
- Add more examples to render test


modules/test-utils/src/test-renderer.js
```diff
@@ -158,17 +161,21 @@ export default class TestRenderer {
       this.currentTestCase = testCase;
       this._currentTestCaseStartTime = animationProps.time;
       this._currentTestCaseStartTick = animationProps.tick;
+      this._testCaseData = null;
       // initialize test case
       withParameters(animationProps.gl, DEFAULT_RENDER_PARAMETERS, () => {
-        this._testCaseData =
+        Promise.resolve(
```
georgios-uber: This resolves **immediately**
You can just do `this._testCaseData = ...`

Pessimistress: Not if `onInitialized` returns a promise.

georgios-uber: `(testCase.onInitialize() || Promise.resolve({})).then(...)`

georgios-uber: I think that way is easier to read.

Pessimistress: That won't work if `onInitialized` returns a plain object.

georgios-uber: OK, but can you add a comment?

PR #924, opened 2019-02-22
fix animation loop start/stop bugs

rolyatmax:
I noticed three bugs in the `AnimationLoop`'s `start` and `stop` logic as I was digging into it a bit today. Here's a short explanation of each along with a description of what I assume is the intended functionality:

### `AnimationLoop` instance can have multiple render loops

It looks like [this line](https://github.com/uber/luma.gl/blob/master/modules/core/src/core/animation-loop.js#L103) does not actually protect us from firing up two loops. The `this._animationFrameId` property will be falsy until [here](https://github.com/uber/luma.gl/blob/master/modules/core/src/core/animation-loop.js#L131), several async callbacks later.

**Intended functionality:** call `start()` then `start() again` - there should only be one animation loop running.

### `onInitialize` method & other setup logic can run multiple times

Is `start()` meant to be called only once per webgl context? Do we support calling `start()` then `stop()` then `start()` again? If so, it looks like we might end up calling [these lines](https://github.com/uber/luma.gl/blob/master/modules/core/src/core/animation-loop.js#L112-L114) more than once, which seems problematic. It seems like `start` is being used as both an "initialize" method _and_ a "start-the-loop" method. Perhaps we should decouple these two purposes into separate functions.

**Intended functionality:**  call `start()` then, later, `stop()` then, later, `start()` again - the webgl context should not be created a second time and `onInitialize` should not be called a second time.

### Calling `stop` does not always stop the loop

It seems like the `this._animationFrameId` check [here](https://github.com/uber/luma.gl/blob/master/modules/core/src/core/animation-loop.js#L162) doesn't function correctly because it prevents the logic of `stop()` from running until the first `requestAnimationFrame` is run, several async callbacks after `start()` is called. In other words, `stop()` does not become functional until `start()` and all its async calls have completed.

**Intended functionality:**  call `start()` then `stop()` - loop should not run.

-----

The first commit of this PR adds failing tests for these bugs. I hope to follow up shortly with a fix. Also, I imagine the same fixes need to be implemented in `AnimationLoopProxy`. Does that seem right, @Pessimistress?

### DIFF
--- a/modules/core/src/core/animation-loop.js
+++ b/modules/core/src/core/animation-loop.js
@@ -2,7 +2,7 @@
 import {createGLContext, resizeGLContext, resetParameters} from '../webgl/context';
 import {getPageLoadPromise} from '../webgl/context';
 import {isWebGL, requestAnimationFrame, cancelAnimationFrame} from '../webgl/utils';
-import {log, runOnce} from '../utils';
+import {log} from '../utils';
 import assert from '../utils/assert';
 
 // TODO - remove dependency on webgl classes
@@ -59,7 +59,10 @@ export default class AnimationLoop {
     this.gl = gl;
     this.needsRedraw = null;
 
-    this._stopped = true;
+    this._initialized = false;
+    this._running = false;
+    this._animationFrameId = null;
+    this._startPromise = null;
 
     this.setProps({
       autoResizeViewport,
@@ -74,8 +77,6 @@ export default class AnimationLoop {
 
     this._onMousemove = this._onMousemove.bind(this);
     this._onMouseleave = this._onMouseleave.bind(this);
-
-    this._onFirstStart = runOnce(this._onFirstStart.bind(this));
 
     return this;
   }
@@ -102,26 +103,46 @@ export default class AnimationLoop {
   // Starts a render loop if not already running
   // @param {Object} context - contains frame specific info (E.g. tick, width, height, etc)
   start(opts = {}) {
-    if (this._stopped) {
-      this._stopped = false;
-      // console.debug(`Starting ${this.constructor.name}`);
-      // Wait for start promise before rendering frame
-      this._startPromise = getPageLoadPromise()
-        .then(() => {
-          if (this._stopped) {
-            return null;
-          }
-          return this._onFirstStart(opts);
-        })
-        .then(appContext => {
-          if (!this._stopped) {
-            this._addCallbackData(appContext || {});
-            if (appContext !== false) {
-              this._startLoop();
-            }
-          }
-        });
+    if (this._running) {
+      return this;
     }
+    this._running = true;
+    // console.debug(`Starting ${this.constructor.name}`);
+    // Wait for start promise before rendering frame
+    this._startPromise = getPageLoadPromise()
+      .then(() => {
+        if (!this._running || this._initialized) {
+          return null;
+        }
+
+        this._initialized = true;
+
+        // Create the WebGL context
+        this._createWebGLContext(opts);
+        this._createFramebuffer();
+        this._startEventHandling();
+
+        // Initialize the callback data
+        this._initializeCallbackData();
+        this._updateCallbackData();
+
+        // Default viewport setup, in case onInitialize wants to render
+        this._resizeCanvasDrawingBuffer();
+        this._resizeViewport();
+
+        // Note: onIntialize can return a promise (in case it needs to load resources)
+        return this.onInitialize(this.animationProps);
+      })
+      .then(appContext => {
+        if (this._running) {
+          this._addCallbackData(appContext || {});
+          if (appContext !== false) {
+            // cancel any pending renders to ensure only one loop can ever run
+            cancelAnimationFrame(this._animationFrameId);
+            this._animationFrameId = requestAnimationFrame(this._renderFrame);
+          }
+        }
+      });
     return this;
   }
 
@@ -148,11 +169,11 @@ export default class AnimationLoop {
   // Stops a render loop if already running, finalizing
   stop() {
     // console.debug(`Stopping ${this.constructor.name}`);
-    if (!this._stopped) {
+    if (this._running) {
       this._finalizeCallbackData();
       cancelAnimationFrame(this._animationFrameId);
       this._animationFrameId = null;
-      this._stopped = true;
+      this._running = false;
     }
     return this;
   }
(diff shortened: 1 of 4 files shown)


test/modules/core/core/animation-loop.spec.js
```diff
@@ -71,3 +71,78 @@ test('WebGL#AnimationLoop redraw', t => {
...
+
+  const animationLoop = new AnimationLoop({
+    gl,
+    onInitialize: () => {
+      initializeCalled++;
+    },
+    onRender: () => {
+      animationLoop.stop();
+      // FIXME: this is not directly test that only one loop has been started
+      t.is(initializeCalled, 1, 'onInitialize called');
```
rolyatmax: Note the `FIXME` here. Any ideas on how we could directly test that only one loop is running?

test/modules/core/core/animation-loop.spec.js
```diff
@@ -71,3 +71,78 @@ test('WebGL#AnimationLoop redraw', t => {
...
+      initializeCalled++;
+    }
+  });
+  animationLoop.start();
+  setTimeout(() => animationLoop.stop(), 50);
+  setTimeout(() => animationLoop.start(), 100);
+  setTimeout(() => {
+    t.is(initializeCalled, 1, 'onInitialize called');
+    t.end();
+  }, 150);
```
rolyatmax: These `setTimeout`s are a bit ugly. I'm just trying to get past the async `getPageLoadPromise()` [here](https://github.com/uber/luma.gl/blob/master/modules/core/src/core/animation-loop.js#L105).

rolyatmax:
Updated with my first pass. Still need to apply fixes to `AnimationLoopProxy`. Also would still kinda like to make `start()` synchronous and move the async setup logic somewhere else. 🤔

modules/core/src/core/animation-loop.js
```diff
@@ -73,6 +75,8 @@ export default class AnimationLoop {
     this._onMousemove = this._onMousemove.bind(this);
     this._onMouseleave = this._onMouseleave.bind(this);
 
+    this._onFirstStart = runOnce(this._onFirstStart.bind(this));
```
tsherif: I think this makes it unclear what `onFirstStart` is doing. If you look at the method and where it's used, it looks like it's being called every time `start` is called. Plus, it doesn't look like luma handles context loss right now, but when we do implement it, this will make it harder to set things up again.

I'd suggest just having an `initialized` flag that `_onFirstStart` checks and just returns if it's set.

modules/core/src/core/animation-loop.js
```diff
@@ -227,7 +240,9 @@ export default class AnimationLoop {
 
     this.redraw();
 
-    // Request another render frame now
+    // Request another render frame now but cancel any pending renders
+    // to ensure only one loop can ever run
+    cancelAnimationFrame(this._animationFrameId);
```
tsherif: Is this necessary here?

modules/core/src/core/animation-loop.js
```diff
@@ -59,6 +59,8 @@ export default class AnimationLoop {
     this.gl = gl;
     this.needsRedraw = null;
 
+    this._stopped = true;
```
Pessimistress: I'm not a big fan of double negatives. Can we change this to something like `started`?

modules/core/src/core/animation-loop.js
Pessimistress: +1

modules/core/src/core/animation-loop.js
rolyatmax: It's not necessary as long as `_renderFrame` is only ever called by itself or `_startLoop` (which also contains the check). This is currently the case, but I'm not sure we can guarantee this going forward. I'm generally against defense coding measures like this, but in this case I've found it to be useful for two reasons:

1. It protects against a really easy-to-introduce / hard-to-discover bug. (I only discovered that we had multiple loops running in our application because I happened to be reading and reasoning about the AnimationLoop logic here.)
2. It communicates to readers the (perhaps self-evident) intent that this is meant to be a "singleton" loop.

One option is to move this `_renderFrame` function into `startLoop`, a pattern I've seen quite a bit. That would allow us to drop that extra check and also make sure `_renderFrame` is only called in this one spot. Something like:

```js
_startLoop () {
  const renderFrame = () => {
    if (!this._running) return
    this.redraw()
    this._animationFrameId = requestAnimationFrame(renderFrame)
  }
  cancelAnimationFrame(this._animationFrameId)
  this.animationFrameId = requestAnimationFrame(renderFrame)
}
```

Thoughts?

modules/core/src/core/animation-loop.js
rolyatmax: How about `_running`? To me, `started === true` implies the loop has been started at some point in the past, even if it might not currently be running.

modules/core/src/core/animation-loop.js
rolyatmax: Makes sense to me! How about I rename `_onFirstStart` to `_initialize` and add an `_initialized` flag, as you suggest?

rolyatmax: Actually, I'm going to put that `_onFirstStart` logic back into `start` to reduce the number of changes here.

modules/core/src/core/animation-loop.js
```diff
@@ -98,41 +103,46 @@ export default class AnimationLoop {
...
-        .then(appContext => {
-          if (!this._stopped) {
-            this._addCallbackData(appContext || {});
-            if (appContext !== false && !this._animationFrameId) {
-              this._animationFrameId = requestAnimationFrame(this._renderFrame);
-            }
+    // Wait for start promise before rendering frame
+    this._startPromise = getPageLoadPromise()
+      .then(() => {
+        if (!this._running || this._initialized) {
```
Pessimistress: - Just curious, how can `_running` get set during this time?
- Is `_initialized` the only thing preventing a stopped animation loop from running again? @ibgreen it's not clear from the documentation whether stopping then restarting an animation loop is encouraged, but it was definitely possible before. If we want to keep it that way, there are more issues to be addressed, e.g. every `start` call adds a new canvas.

rolyatmax: These changes make it so that you can stop/restart an animation loop _without_ running this initialization logic again - that's the second of the bugs listed up top.

modules/core/src/core/animation-loop.js
```diff
@@ -98,41 +103,46 @@ export default class AnimationLoop {
...
+          return null;
+        }
+
+        this._initialized = true;
```
Pessimistress: Move setting this flag to *after* all initialization tasks are executed?

modules/core/src/core/animation-loop.js
Pessimistress: Edit: read this again, `_initialized` just skips running this block. No breaking change here.

rolyatmax:
> Once luma.gl build tooling gets upgraded I think we should rewrite the `Promise`-based code in here using `async/await` I suspect that will help make the parts of the code that deal e.g. with async initialization much clearer, which should help avoid this type of tricky bugs.

Happy to help if it's needed when the time comes!

modules/core/src/core/animation-loop.js
rolyatmax: From offline convo with Tarek, let's apply this change as part of this PR.

PR #945, opened 2019-03-03
glTF: Handle alpha mask and blend

georgios-uber:


modules/core/src/scenegraph/gltf-instantiator/gltf-material.js
```diff
@@ -252,6 +254,18 @@ class GLTFMaterialParser {
...
+    if (material.alphaMode === 'MASK') {
+      const {alphaCutoff = 0.5} = material;
+      this.defines.ALPHA_CUTOFF = 1;
+      this.uniforms.u_AlphaCutoff = alphaCutoff;
+    } else if (material.alphaMode === 'BLEND') {
+      log.warn('BLEND alphaMode might not work well because it requires mesh sorting')();
+      Object.assign(this.parameters, {
+        blend: true,
+        blendEquation: this.gl.FUNC_ADD,
+        blendFunc: [this.gl.SRC_ALPHA, this.gl.ONE_MINUS_SRC_ALPHA]
```
Pessimistress: Use natural blending?
https://github.com/uber/deck.gl/blob/master/modules/core/src/lib/deck.js#L493

georgios-uber:
aaaa

modules/core/src/scenegraph/gltf-instantiator/gltf-material.js
```diff
@@ -252,6 +254,18 @@ class GLTFMaterialParser {
       this.parseTexture(material.emissiveTexture, 'u_EmissiveSampler', 'HAS_EMISSIVEMAP');
       this.uniforms.u_EmissiveFactor = material.emissiveFactor || [0, 0, 0];
     }
+    if (material.alphaMode === 'MASK') {
```
ibgreen: * use switch statement (case 'MASK': case 'BLEND)
* Break out a helper function to handle the blend setup?
* Changing to `MASK` does not reset `parameters`?

examples/gltf/app.js
```diff
@@ -99,7 +99,9 @@ export class DemoApp {
 
     this.glOptions = {
       // Use to test gltf with webgl 1.0 and 2.0
-      webgl2: true
+      webgl2: true,
+      // alpha causes issues with some glTF demos
+      alpha: false
```
ibgreen: We might want to add some information about this in the glTF or PBRMaterial docs?

modules/shadertools/src/modules/pbr/pbr-fragment.glsl.js
```diff
@@ -41,6 +41,10 @@ uniform sampler2D u_OcclusionSampler;
 uniform float u_OcclusionStrength;
 #endif
 
+#ifdef ALPHA_CUTOFF
```
ibgreen: Nit: This feature just adds a single if statement, looks rather cheap, is it really justified to require a shader recompile to change it?

tsherif:
If we're doing alpha masking you might want to consider supporting alpha to coverage: https://en.wikipedia.org/wiki/Alpha_to_coverage
More detailed explanation: https://medium.com/@bgolus/anti-aliased-alpha-test-the-esoteric-alpha-to-coverage-8b177335ae4f

Enabling in WebGL is just `gl.enable(gl.ALPHA_TO_COVERAGE)`

georgios-uber:
Thanks @tsherif I'll look into it.

PR #959, opened 2019-03-08
add support for mat attribute types

georgios-uber:
add support for mat attribute types


tsherif:
This is great! I think the last step would be to automate the setting up the attribute pointers, instead of having to to explicitly create `..._LOCATION_1`, etc. Those names could be used internally, but the user could just declare a mat2, mat3 or mat4 attributes, and Luma would set up the 2, 3, or 4 attributes automatically. Would keep the user from having to do this kind of this, which is the more painful part of dealing with matrix attributes: https://github.com/uber/luma.gl/blob/master/examples/core/dof/app.js#L111-L144

PR #987, opened 2019-03-15
Add CubeTexture LOD support and use it for PBR Environment

georgios-uber:


georgios-uber:
Screen Shot 2019-03-15 at 3 43 44 PM

examples/gltf/app.js
```diff
@@ -1,11 +1,20 @@
 import {GLTFParser} from '@loaders.gl/gltf';
 import {DracoDecoder} from '@loaders.gl/draco';
-import {AnimationLoop, setParameters, clear, createGLTFObjects, log} from 'luma.gl';
+import {
+  AnimationLoop,
+  setParameters,
+  clear,
+  createGLTFObjects,
+  GLTFEnvironment,
```
ibgreen: Let's make this an underscore export

modules/core/src/scenegraph/gltf/gltf-material-parser.js
```diff
@@ -93,7 +23,7 @@ export default class GLTFMaterialParser {
     this.parameters = {};
 
     if (ibl) {
-      this.env = new GLTFEnv(gl);
+      this.env = ibl;
```
ibgreen: I feel that `ibl` is too terse, we need a better name. Type out whatever it means or use another name.

modules/core/src/webgl/classes/texture-cube.js
```diff
@@ -81,15 +84,30 @@ export default class TextureCube extends Texture {
     const {gl} = this;
     const imageDataMap = pixels || data;
 
-    return Promise.all(FACES.map(face => imageDataMap[face])).then(resolvedFaces => {
+    return Promise.all(
+      FACES.map(face => {
+        const facePixels = imageDataMap[face];
+        return Promise.all(Array.isArray(facePixels) ? facePixels : [facePixels]);
+      })
+    ).then(resolvedFaces => {
```
ibgreen: Use async await

modules/core/src/webgl/classes/texture-cube.js
```diff
@@ -81,15 +84,30 @@ export default class TextureCube extends Texture {
     const {gl} = this;
     const imageDataMap = pixels || data;
 
-    return Promise.all(FACES.map(face => imageDataMap[face])).then(resolvedFaces => {
+    return Promise.all(
+      FACES.map(face => {
+        const facePixels = imageDataMap[face];
+        return Promise.all(Array.isArray(facePixels) ? facePixels : [facePixels]);
```
ibgreen: Why do we need Promise.all here? Add comment...

modules/core/src/webgl/classes/texture-cube.js
```diff
@@ -81,15 +84,30 @@ export default class TextureCube extends Texture {
...
       this.bind();
 
       FACES.forEach((face, index) => {
-        if (width && height) {
-          gl.texImage2D(face, 0, format, width, height, border, format, type, resolvedFaces[index]);
-        } else {
-          gl.texImage2D(face, 0, format, format, type, resolvedFaces[index]);
+        if (resolvedFaces[index].length > 1 && this.opts.mipmaps !== false) {
+          log.warn(
+            `Texture ${
```
ibgreen: We normally avoid long error messages that add to the bundle size but are very seldom used.

Add as comment and log something minimal, user can click their way back here and read the comment.

modules/core/src/webgl/classes/texture-cube.js
```diff
@@ -81,15 +84,30 @@ export default class TextureCube extends Texture {
...
+      })
+    ).then(resolvedFaces => {
       this.bind();
 
       FACES.forEach((face, index) => {
-        if (width && height) {
-          gl.texImage2D(face, 0, format, width, height, border, format, type, resolvedFaces[index]);
-        } else {
-          gl.texImage2D(face, 0, format, format, type, resolvedFaces[index]);
+        if (resolvedFaces[index].length > 1 && this.opts.mipmaps !== false) {
```
ibgreen: I think we need some comments here, otherwise this will break in next refactor

modules/core/src/webgl/classes/texture-cube.js
```diff
@@ -81,15 +84,30 @@ export default class TextureCube extends Texture {
     const {gl} = this;
     const imageDataMap = pixels || data;
 
-    return Promise.all(FACES.map(face => imageDataMap[face])).then(resolvedFaces => {
+    return Promise.all(
```
ibgreen: I think we need some comments here, otherwise this will break in next refactor

modules/core/src/scenegraph/gltf/gltf-environment.js
```diff
@@ -0,0 +1,97 @@
+import GL from '@luma.gl/constants';
+import {Texture2D, TextureCube} from '../../webgl';
+import {loadImage} from '../../core/load-file';
+
+export default class GLTFEnvironment {
```
ibgreen: I am not quite clear on what this class represents. A number of cube maps that together represents an advanced environment mapping? Could you try to write a doc page for it as a step towards articulating this? 

Also, how do we get additional such maps to use? (Apart from the one example we got through glTF demo)
* Are there libraries of such environments available? 
* Are they part of glTF files?
* Can we generate them? The classic technique is to set up 6 cameras and generate such maps from our scenegraph by rendering in each direction? Then we can create perfect reflections for any object in the scene.

modules/core/src/scenegraph/gltf/gltf-environment.js
```diff
@@ -0,0 +1,97 @@
...
+    return `${this.baseUrl}brdfLUT.png`;
+  }
+
+  makeCube({id, getTextureForFace, parameters}) {
+    return new TextureCube(this.gl, {
+      id,
+      mipmaps: false,
+      parameters,
+      pixels: {
+        [GL.TEXTURE_CUBE_MAP_POSITIVE_X]: getTextureForFace('right'),
```
ibgreen: Are they always called right/left etc or does that just happen to be the case for the example you had? Should this be a parameter to the class?

modules/core/src/scenegraph/gltf/gltf-environment.js
```diff
@@ -0,0 +1,97 @@
...
+        [GL.TEXTURE_CUBE_MAP_NEGATIVE_Z]: getTextureForFace('back')
+      }
+    });
+  }
+
+  getDiffuseEnvSampler() {
+    if (!this._DiffuseEnvSampler) {
+      this._DiffuseEnvSampler = this.makeCube({
+        id: 'DiffuseEnvSampler',
+        getTextureForFace: dir => loadImage(this.getTexUrl('papermill', 'diffuse', dir)),
```
ibgreen: papermill? This sounds specific to one such map? Replace with parameter?

modules/core/src/scenegraph/gltf/gltf-material-parser.js
georgios-uber: `Image Based Rendering or IBL : a technique that uses an image as a light source`

modules/core/src/webgl/classes/texture-cube.js
georgios-uber: Ok

modules/core/src/scenegraph/gltf/gltf-environment.js
georgios-uber: Yes, programs like [this one](https://github.com/derkreature/IBLBaker)

modules/core/src/scenegraph/gltf/gltf-environment.js
georgios-uber: User should subclass if they want different names.

examples/gltf/app.js
georgios-uber: _GLTFEnvironment ok

modules/core/src/scenegraph/gltf/gltf-material-parser.js
georgios-uber: I replaced all of references to `imageBasedLighting`. Should I also rename environment?

modules/core/src/scenegraph/gltf/gltf-environment.js
georgios-uber: Adding it to base url.

modules/core/src/scenegraph/gltf/gltf-environment.js
georgios-uber: I'm gonna refactor this to accept **getTexUrl** as a function parameter. Then the user can do what they want.

modules/core/src/webgl/classes/texture-cube.js
georgios-uber: Probably need unit test for that too.

georgios-uber:
Added doc+unit test.

georgios-uber:
Ping ping

docs/api-reference/webgl/texture-cube.md
```diff
@@ -21,6 +21,20 @@ const textureCube = new TextureCube(gl, {width, height, dataFormat, pixels: {
...
+  [GL.TEXTURE_CUBE_MAP_POSITIVE_X]: [imagePosX_LOD_0, imagePosX_LOD_1, imagePosX_LOD_2],
+  [GL.TEXTURE_CUBE_MAP_POSITIVE_Y]: [imagePosY_LOD_0, imagePosY_LOD_1, imagePosY_LOD_2],
+  [GL.TEXTURE_CUBE_MAP_POSITIVE_Z]: [imagePosZ_LOD_0, imagePosZ_LOD_1, imagePosZ_LOD_2],
+  [GL.TEXTURE_CUBE_MAP_NEGATIVE_X]: [imageNegX_LOD_0, imageNegX_LOD_1, imageNegX_LOD_2],
+  [GL.TEXTURE_CUBE_MAP_NEGATIVE_Y]: [imageNegY_LOD_0, imageNegY_LOD_1, imageNegY_LOD_2],
+  [GL.TEXTURE_CUBE_MAP_NEGATIVE_Z]: [imageNegZ_LOD_0, imageNegZ_LOD_1, imageNegZ_LOD_2]
+}});
+```
+
+This class supports _Async Textures_. You can provide promises instead of images on all methods.
```
ibgreen: Some example of "provide promises instead of images on all methods"? Questions here when reading, can you provide promises for each LOD?, or each array of LOD or does it need to be the `pixels` arg itself that is a promise?

ibgreen: Nit: We should update the `### constructor` header below to show which params it takes...

examples/gltf/app.js
```diff
@@ -331,21 +346,21 @@ export class DemoApp {
     switch (value) {
       case 'exclusive':
         Object.assign(this.loadOptions, {
-          pbrIbl: true,
+          imageBasedLighting: this.environment,
```
ibgreen: Is there any reason not to just call this prop `environment`?

* `imageBasedLighting` sounds like a boolean.
* `imageBadedLightingEnvironment`?

modules/core/src/scenegraph/gltf/gltf-environment.js
ibgreen: > Refactor this to accept getTexUrl as a function parameter

Awesome, thank you!

* Can we send the GL constant to that function, rather than the arbitrary (left/top/bottom...) strings? 
* I have seen a couple of conventions for this (`negative-x`, ...), not convinced we should choose one on the user's behalf, unless you have more evidence that this being standardized.

modules/core/src/scenegraph/gltf/gltf-environment.js
```diff
@@ -0,0 +1,90 @@
...
+          [GL.TEXTURE_MIN_FILTER]: GL.LINEAR_MIPMAP_LINEAR,
+          [GL.TEXTURE_MAG_FILTER]: GL.LINEAR
+        }
+      });
+    }
+
+    return this._SpecularEnvSampler;
+  }
+
+  getBrdfTex() {
```
ibgreen: Please type out `Texture` here and in other places.

If we are going to keep `brdf` as a term:
1. It needs to be mentioned/explained in the docs
2. Type in in all caps?

modules/core/src/scenegraph/gltf/gltf-material-parser.js
ibgreen: > I replaced all of references to imageBasedLighting.

OK thanks, let's go with that for now. 

> Should I also rename environment?

Yes, maybe `ImageBasedLightingEnvironment`, or `IBLEnvironment`.

I think writing a doc about that class, with short explanation, references to papers, references to the tool you mention below would be really helpful.

I assume the PBR module in some sense depends on or "collaborates with" this class, making that relationship crips and clear in the docs (perhaps by updating PBR docs?) would surely cast this PR in a much clearer light.

examples/gltf/app.js
georgios-uber: imageBasedLightingEnvironment

georgios-uber: it is a **specific** type of environment

modules/core/src/scenegraph/gltf/gltf-environment.js
georgios-uber: **Bidirectional reflectance distribution function** 

so `getBidirectionalReflectanceDistributionFunctionLookupTexture()` sounds good?

docs/api-reference/webgl/texture-cube.md
georgios-uber: You can provide multiple individual promises. I'll update.

modules/core/src/scenegraph/gltf/gltf-environment.js
georgios-uber: I'll use the gl constant.

georgios-uber: Done

docs/api-reference/webgl/texture-cube.md
georgios-uber: I updated please take a second look.

PR #1029, opened 2019-04-02
glTF Loader: Allow to async return after all resources are loaded

georgios-uber:


modules/addons/src/gltf/gltf-scenegraph-loader.js
```diff
@@ -1,7 +1,12 @@
+/* global window */
 import {GLTFParser} from '@loaders.gl/gltf';
 import {DracoDecoder} from '@loaders.gl/draco';
 import createGLTFObjects from './create-gltf-objects';
 
+function wait() {
+  return new Promise(resolve => window.setTimeout(resolve, 100));
```
tsherif: `requestAnimationFrame`. Avoid having code running out side an `rAF` callback, since there's a good chance it will include GL code.

modules/addons/src/gltf/gltf-scenegraph-loader.js
```diff
@@ -10,7 +15,27 @@ async function parse(data, options, uri, loader) {
     DracoDecoder
   });
 
-  return Object.assign({gltfParser, gltf}, createGLTFObjects(options.gl, gltf, options));
+  const gltfObjects = createGLTFObjects(options.gl, gltf, options);
+
+  if (options.waitForFullLoad) {
```
tsherif: This should probably be pulled into a separate function similar to `Query.createPoll`: https://github.com/uber/luma.gl/blob/master/modules/webgl/src/classes/query.js#L133

georgios-uber:
```
async function waitWhileCondition(condition) {
  while (condition()) {
    await new Promise(resolve => window.requestAnimationFrame(resolve));
  }
}
```

georgios-uber: Does this work?

PR #1058, opened 2019-04-12
Don't modify input attributes

tsherif:
This was interfering with reuse of deck mesh attributes.

### DIFF
--- a/modules/core/src/geometry/geometry.js
+++ b/modules/core/src/geometry/geometry.js
@@ -91,8 +91,11 @@ export default class Geometry {
       // Move indices to separate field
       if (attributeName === 'indices') {
         assert(!this.indices);
-        delete attribute.isIndexed;
         this.indices = attribute;
+        if (this.indices.isIndexed !== undefined) {
+          this.indices = Object.assign({}, this.indices);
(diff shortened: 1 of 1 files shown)


modules/core/src/geometry/geometry.js
```diff
@@ -92,7 +92,6 @@ export default class Geometry {
       if (attributeName === 'indices') {
         assert(!this.indices);
         delete attribute.isIndexed;
-        delete attributes[attributeName];
```
Pessimistress: Is deleting this going to change the model behavior (creating an extra buffer)? Can we just make a copy of the `attributes` object?

modules/core/src/geometry/geometry.js
```diff
@@ -92,7 +92,6 @@ export default class Geometry {
       if (attributeName === 'indices') {
         assert(!this.indices);
         delete attribute.isIndexed;
```
Pessimistress: This is also mutating the input.

modules/core/src/geometry/geometry.js
tsherif: No, this was just modifying the input object. Buffers are created based on `this.attributes`.

PR #1075, opened 2019-04-18
Add interpolation support to gltf-animator

georgios-uber:


modules/addons/src/gltf/gltf-animator.js
```diff
@@ -64,6 +64,38 @@ function applyTranslationRotationScale(gltfNode, node) {
   }
 }
 
+function linearInterpolate(path, t, a, b) {
+  // TODO: implement SLERP when path is rotation
+  return a.map((v, index) => v * (1 - t) + b[index] * t);
```
tsherif: Would recommend not creating a new array here. This function is likely to be called **a lot** in animation system.

modules/addons/src/gltf/gltf-animator.js
```diff
@@ -64,6 +64,38 @@ function applyTranslationRotationScale(gltfNode, node) {
...
+
+function interpolate(time, {input, interpolation, output}, path) {
+  const maxTime = input[input.length - 1];
+  const animationTime = time % maxTime;
+
+  const nextIndex = input.findIndex(t => t >= animationTime);
+  const previousIndex = Math.max(0, nextIndex - 1);
+
+  switch (interpolation) {
+    case 'STEP':
```
tsherif: This is the same as `default`, so you can just remove this and put a comment under `default` that it's considered `STEP`.

modules/addons/src/gltf/gltf-animator.js
```diff
@@ -64,6 +64,38 @@ function applyTranslationRotationScale(gltfNode, node) {
...
+      return output[previousIndex];
+
+    case 'LINEAR':
+      const previousTime = input[previousIndex];
+      const nextTime = input[nextIndex];
+      const t = (animationTime - previousTime) / (nextTime - previousTime);
```
tsherif: I'd rather see this as:
```
if (nextTime !== previousTime) {
     const t = (animationTime - previousTime) / (nextTime - previousTime);
     return linearInterpolate(path, t, output[previousIndex], output[nextIndex]);
}
```
It's clearer about what the condition is and doesn't perform the calculations when they won't be used.

georgios-uber: `if (nextTime > previousTime)`

modules/addons/src/gltf/gltf-animator.js
georgios-uber: Default is not STEP. Fixed it now to print warning.

modules/addons/src/gltf/gltf-animator.js
georgios-uber: Changed to code to not create any new arrays.

PR #1078, opened 2019-04-18
TextEncoder polyfill for gltf example

tsherif:
To get it working in Edge


tsherif:
I don't know what happened, but this doesn't seem to be working anymore. I'll take a closer look tomorrow.

tsherif:
@ibgreen I'm going to go ahead with this because I encountered the following problems with importing a polyfill:
- `text-encoding` doesn't properly add polyfills to the global namespace when imported.
- `fast-text-encoding` only accepts the string "utf-8" and loaders.gl uses "utf8".

We can find a cleaner solution later, but for now I just want to get the release ready.

PR #1085, opened 2019-04-27
Use start/end channel parameters instead of duration

tsherif:
This makes it possible to offset channel timelines from the base timeline. Also added separate wrap modes for start and end.


Pessimistress:
As @ibgreen has commented on the previous PR, let's write a mini RFC for this.
Personally I find this less intuitive than the duration model. If I explicitly specified start/end time for my animation, I expect the animation to be confined to that time window.

PR #1088, opened 2019-04-30
Animation timeline RFC

tsherif:
RFC for a timeline management system

### DIFF
--- a/dev-docs/RFCs/v7.1/animation-timeline-rfc.md
+++ b/dev-docs/RFCs/v7.1/animation-timeline-rfc.md
@@ -22,7 +22,7 @@ The Elevate team has requested the ability to control transitions in deck.gl, sp
 
 ## Overview
 
-A timeline manager that can provide `time` values to be used in animations are independent of wall time. The timeline manager should support the following features:
+A timeline manager that can provide `time` values to be used in animations that are independent of wall time. The timeline manager should support the following features:
 
 - play: provide a `time` value that elapses at the same rate as wall time
 - pause: `time` remains constant at the current value
@@ -53,9 +53,9 @@ A timeline with two channels:
                    pause  play
                      |     |
 Wall time:      0----5----10----15----20
-`time`          0----5     5----10----15
-Channel 1:       0---2     2---2---2---2
-Channel 2:                 0----10
+time:           0----5     5----10----15
+channelTime 1:   0---2     2---2---2---2
+channelTime 2:             0----10
 ```
 
 ## Implentation
@@ -65,11 +65,11 @@ Channel 2:                 0----10
 - `play`: elapse `time` automatically with wall time
 - `pause`: stop elapsing `time` automatically with wall time
 - `reset`: set `time` to 0
-- `setTime(time): set `time` to a specific value
+- `setTime(time)`: set `time` to a specific value
 - `getTime`: get current `time`
 - `addChannel(props)`: create a new channel with given properties and return a handle to it
 - `getChannelTime(handle)`: get the current `channelTime` from the `channel` indicated by `handle`
-- `removeChannel(handle)`: remove a channel from the timeline`
+- `removeChannel(handle)`: remove a channel from the timeline
 
-The `time` property provided in `animationProps` will be the value returns by `animationLoop.timeline.getTime()`. This will ensure that all animations based on it will follow timeline controls rather than wall time. `animationLoop.timelin` will also be passed in `animatonProps` so that applications can easily manipulate it.
+The `time` property provided in `animationProps` will be the value returned by `animationLoop.timeline.getTime()`. This will ensure that all animations tracking `animationProps.time` will follow timeline controls rather than wall time. `animationLoop.timeline` will also be passed in `animatonProps` so that applications can easily manipulate it.
 
--- a/dev-docs/RFCs/v7.1/animation-timeline-rfc.md
+++ b/dev-docs/RFCs/v7.1/animation-timeline-rfc.md
@@ -7,7 +7,7 @@
 
 ## Summary
 
-This RFC specifies a timeline management system to facilate more complex animations.
+This RFC specifies a timeline management system to facilate more complex animations that can be easily controlled by the application.
 
 
 ## Background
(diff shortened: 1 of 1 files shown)


dev-docs/RFCs/v7.1/animation-timeline-rfc.md
```diff
@@ -0,0 +1,75 @@
...
+A timeline manager that can provide `time` values to be used in animations that are independent of wall time. The timeline manager should support the following features:
+
+- play: provide a `time` value that elapses at the same rate as wall time
+- pause: `time` remains constant at the current value
+- set: set `time` to a specific value
+- multiple `channels` that provide `channelTime` values, related to `time`, but with the following additional properties:
+  * `rate`: a scaling factor that indicates how quickly `channelTime` elapses relative to `time`
+  * `startTime`: an offset into `time` at which `channelTime` begins elapsing
+  * `duration`: how long `channelTime` runs for
+  * `wrapMode`: what to do when `duration` elapses. Options are `clamp` (stop elapsing) or `loop` (go back to 0)
```
Pessimistress: I think it's more common to define the loop count? E.g. [animation-iteration-count](https://developer.mozilla.org/en-US/docs/Web/CSS/animation-iteration-count) and [tween.js](https://github.com/tweenjs/tween.js/blob/master/docs/user_guide.md#repeat)

dev-docs/RFCs/v7.1/animation-timeline-rfc.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Overview
+
+A timeline manager that can provide `time` values to be used in animations that are independent of wall time. The timeline manager should support the following features:
+
+- play: provide a `time` value that elapses at the same rate as wall time
+- pause: `time` remains constant at the current value
+- set: set `time` to a specific value
+- multiple `channels` that provide `channelTime` values, related to `time`, but with the following additional properties:
+  * `rate`: a scaling factor that indicates how quickly `channelTime` elapses relative to `time`
```
Pessimistress: Call out that this is optional and default to `1`.

dev-docs/RFCs/v7.1/animation-timeline-rfc.md
```diff
@@ -0,0 +1,75 @@
...
+  * `startTime`: an offset into `time` at which `channelTime` begins elapsing
```
Pessimistress: Default to current time?

tsherif: I think default to 0 is most intuitive, e.g. a timeline reset will set default channels to 0 as well.

dev-docs/RFCs/v7.1/animation-timeline-rfc.md
tsherif: Good call. Went with the tween.js naming for `delay` as well.

dev-docs/RFCs/v7.1/animation-timeline-rfc.md
```diff
@@ -0,0 +1,75 @@
...
+
+`AnimatonLoop` will have a new `timeline` property which is an instance of the class `Timeline`. The `Timeline` class will provide the following methods:
+
+- `play`: elapse `time` automatically with wall time
+- `pause`: stop elapsing `time` automatically with wall time
+- `reset`: set `time` to 0
+- `setTime(time)`: set `time` to a specific value
+- `getTime`: get current `time`
+- `addChannel(props)`: create a new channel with given properties and return a handle to it
+- `getChannelTime(handle)`: get the current `channelTime` from the `channel` indicated by `handle`
```
Pessimistress: This can be merged with `getTime`?

tsherif:
@georgios-uber Sounds good. That'll be a nice use-case example.

tsherif:
@ibgreen I'll add a bit more background text to clarify the motivation behind this, but in general, I think the system I intend is much simpler than the one you have in mind. It's not a full animation system (though possibly one could be built on it later), just a system for producing `time` values that go into animations. We currently have a multitude of independent animation systems (transitions, uniforms, gltfAnimations) that would have to be coordinated if they were to be used together. The timeline reduces that problem to one of just producing appropriate `time` values to go into those animations.

> - channel: This seems to effectively be a single animation. Worth discussion renaming this to animation or some less generic name.
> - channel time: The output of the channels does not really seem to be a time. It is more an interpolated value. If we embrace and possibly rename, we could also add support for easings / general functions.

So the timeline (as I'm imagining it) shouldn't be thought of as producing "animation", nor an "interpolation", but remapped `time` values that allow disparate animations to be aligned.

dev-docs/RFCs/v7.1/animation-timeline-rfc.md
```diff
@@ -0,0 +1,82 @@
...
+- `getTime(handle)`: get the current `channelTime` from the `channel` indicated by `handle`. If no handle provided, get current `time`
+- `addChannel(props)`: create a new channel with given properties and return a handle to it
+- `removeChannel(handle)`: remove a channel from the timeline
+
+The `time` property provided in `animationProps` will be the value returned by `animationLoop.timeline.getTime()`. This will ensure that all animations tracking `animationProps.time` will follow timeline controls rather than wall time. `animationLoop.timeline` will also be passed in `animatonProps` so that applications can easily manipulate it.
+
+
+## Integration with GLTFAnimaton
+
+Integration with `GLTFAnimation`should amount to simply passing the timeline object, and optionally a channel handle to the constructor. Then the [animate method](https://github.com/uber/luma.gl/blob/7.0-release/modules/addons/src/gltf/gltf-animator.js) would simply use `timeline.getTime()` to update rather than receiving the `timeMs` argument.
```
georgios-uber: Should each `GLTFAnimation` get one `channel` of the timeline?

georgios-uber:
@Pessimistress What are the plans for integrating this into deck gl?

dev-docs/RFCs/v7.1/animation-timeline-rfc.md
tsherif: I think that would make sense. It could be optional though. If the animation's time doesn't need to be offset/scaled, it can just use the base `time` value.

tsherif:
@georgios-uber The initial integration will just be that the `AnimationLoop.animationProps.time` value that deck uses for transitions (and other animations?) will come from the timeline. The immediate upshot will be that those animations become controlable via `timeline.play`, `timeline.pause` and `timeline.setTime`.

PR #1090, opened 2019-04-30
Fix accessor size for constant multi-column attributes

tsherif:
This caused an problem in deck.gl with matrix attributes, if they were set to constant and the attribute 0 buffer was auto-created (e.g. open the following in Safari to see the issue: https://observablehq.com/@pessimistress/deck-gl-lightingeffect-demo).

### DIFF
--- a/modules/webgl/src/classes/vertex-array.js
+++ b/modules/webgl/src/classes/vertex-array.js
@@ -182,7 +182,9 @@ export default class VertexArray {
     const {location, accessor} = this._resolveLocationAndAccessor(
       locationOrName,
       arrayValue,
-      appAccessor
+      // Ensure that size isn't taken from program for multi-column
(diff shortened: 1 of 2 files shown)


modules/webgl/src/classes/vertex-array.js
```diff
@@ -182,7 +182,9 @@ export default class VertexArray {
     const {location, accessor} = this._resolveLocationAndAccessor(
       locationOrName,
       arrayValue,
-      appAccessor
+      // Ensure that size isn't taken from program for multi-column
+      // attributes
+      Object.assign({size: arrayValue.length}, appAccessor)
```
Pessimistress: Is this going to affect perf? Just send size separately?

Pessimistress:
Add a test?

PR #1093, opened 2019-05-02
Implementation of Timeline RFC

tsherif:
Updates to timeline base on RFC discussion (#1088)

Updated tests and added an example that demonstrates using multiple coordinated timeline channels.


modules/core/src/lib/timeline.js
```diff
@@ -101,11 +76,16 @@ export class Timeline {
   }
 
   _setChannelTime(channel, time) {
-    channel.time = time * channel.rate;
-    if (channel.wrapMode === WRAP_LOOP) {
-      channel.time %= channel.duration;
+    const offsetTime = time - channel.delay;
+    const totalDuration = channel.duration * channel.repeat;
+    // Note(Tarek): Don't loop on final repeat.
+    if (offsetTime >= totalDuration) {
```
Pessimistress: Nit: separate out the rate logic for readability

```js
    if (offsetTime < 0) {
      channel.time = 0;
    } else if (offsetTime >= totalDuration) {
      channel.time = channel.duration;
    } else {
      channel.time = offsetTime % channel.duration;
    }
    channel.time *= channel.rate;
```

georgios-uber:
Where is the integration with gltf animator?

tsherif:
@georgios-uber Done.

tsherif:
@Pessimistress docs updated and channel time calculation simplified.

modules/addons/src/gltf/gltf-instantiator.js
```diff
@@ -38,7 +38,7 @@ export default class GLTFInstantiator {
 
   createAnimator() {
     if (Array.isArray(this.gltf.animations)) {
-      return new GLTFAnimator(this.gltf);
+      return new GLTFAnimator(this.gltf, this.options.timeline, this.options.channel);
```
Pessimistress: - Channels need to be removed with the glTF scene is removed
- Separate out the glTF stuff to a different PR?

tsherif: The channel isn't added by the animator, so it shouldn't remove it.

examples/core/gltf/app.js
```diff
@@ -262,7 +262,11 @@ export default class AppAnimationLoop extends AnimationLoop {
       blend: false
     });
 
-    this.loadOptions = Object.assign({}, DEFAULT_OPTIONS);
+    const channel = this.timeline.addChannel({
+      rate: 0.5
```
georgios-uber: Why not 1.0

tsherif: Just to have it do something.

modules/addons/src/gltf/gltf-instantiator.js
georgios-uber: Separate out the glTF stuff to a different PR

examples/core/gltf/app.js
georgios-uber: ok

tsherif:
@georgios-uber I've already added the gltf integration based on your earlier request, and it's light enough that I don't think it's worth it now to revert and create a separate PR. 

The channel is just a handle that the timeline uses. It's optional and if omitted the base timeline time will be used.

modules/core/src/lib/timeline.js
```diff
@@ -53,30 +44,14 @@ export class Timeline {
   }
 
   setTime(time) {
-    this._setChannelTime(this, time);
+    this.time = Math.max(0, time);
```
rolyatmax: I wonder if this should throw if `time < 0`. If someone passes a negative time they either (1) don't understand `setTime`'s requirements and would benefit from the message, or (2) didn't expect their code to pass a negative number and would benefit from the message. If throwing is too aggressive, then maybe a console warning would do the job?

tsherif: I don't really consider negative input to be an error. It's just that it will be clamped to 0. If negative values are passed in unintentionally, I'm thinking it should be pretty easy to spot, since the time values won't move (e.g. an animation will stop progressing).

Issue #1099, opened 2019-05-13
Framebuffer setDrawBuffers issue


Tirondzo:
```js
this.rttFrameBuffer.update({drawBuffers:[
  GL.COLOR_ATTACHMENT0,
  GL.COLOR_ATTACHMENT1,
  GL.COLOR_ATTACHMENT2
]});
```
Or any other update drawBuffers raises an error.

That's why:
https://github.com/uber/luma.gl/blob/f0e8bc9de0365bf08281c79a4ad69511a30fde21/modules/webgl/src/classes/framebuffer.js#L147
Should be
```js
this._setDrawBuffers(gl, drawBuffers); 
```

1chandu:
Good find @Tirondzo , would you be able to make a PR.

Tirondzo:
https://github.com/uber/luma.gl/pull/1102

@1chandu, Done.
Also _setReadBuffers had the same issue, fixed that.

PR #1102, opened 2019-05-14
Framebuffer set buffers fix

Tirondzo:
Issue: https://github.com/uber/luma.gl/issues/1099

Added gl argument to _setDrawBuffers and _setReadBuffers calls that obviously was missed.


tsherif:
Actually, @Tirondzo can you update as suggested by @ibgreen? I.e, change the signature of `_setReadBuffer` and `_setDrawBuffer` not to take a `gl` argument, and have those methods use `this.gl`.

Tirondzo:
@tsherif, Sure, that's definitely the better.
Updated.

tsherif:
Awesome, thanks!

PR #1103, opened 2019-05-14
effects module RFC

jianhuang01:
#### Background
#1105
For better effect modules organization and management, here is the RFC for a new effects lib

#### Change List
- RFC

### DIFF
--- a/dev-docs/RFCs/v7.1/effects-module-rfc.md
+++ b/dev-docs/RFCs/v7.1/effects-module-rfc.md
@@ -62,7 +62,7 @@ effects/src
 </pre>
 
 * Objects under experimental folder will be exported as `_ObjectName`
-* `@luma.gl/glfx` will re-export necessary objects from @luma.gl/effects to backward compatibility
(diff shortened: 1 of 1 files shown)


dev-docs/RFCs/v7.1/effects-module-rfc.md
```diff
@@ -0,0 +1,69 @@
...
+	    |
+	    filter folders from `@luma.gl/gflx`
+	|
+	experimental
+	    |
+	    folders from old `@luma.gl/effects`
+</pre>
+
+* Objects under experimental folder will be exported as `_ObjectName`
+* `@luma.gl/glfx` will re-export necessary objects from @luma.gl/effects for backward compatibility
```
Pessimistress: I propose we drop the `@luma.gl/glfx` module entirely in 7.1. If no new release is published for this submodule, no backward compatibility is needed.

Pessimistress:
I would personally like to see render test coverage for the effects module in 7.1. This will be important for release quality once we start promoting this module with deck.gl.

Issue #1105, opened 2019-05-16
refactor effects and glfx libs


jianhuang01:
#### Background 
refactor @luma.gl/effects libs to manage all the effect shader modules and passes
#### To Do List 
  - [x] Add label and assign to milestone 
  - [x] Coding
  - [x] Doc update
  - [x] Test

PR #1106, opened 2019-05-16
Refactor effects and glfx libs (part 1)

jianhuang01:
For #1105 

#### Change List
- move current shader modules and passes to experimental folder


tsherif:
Is the idea that these effects aren't used? Were they exposed as part of luma's API? Do we still need them?

jianhuang01:
> Is the idea that these effects aren't used? Were they exposed as part of luma's API? Do we still need them?

We didn't publish @luma.gl/effects lib yet, so all the old code is for experimental, used in a wip example.

PR #1107, opened 2019-05-16
refactor effect and glfx libs (part2)

jianhuang01:
For #1105 

#### Change List
- move shader modules from glfx to effects
- re-export objects from effects in glfx
- clean up bad shader modules
- update glfx example


tsherif:
I'm fine with moving them to a wip folder for now, but I prefer it go into `wip/effects/etc` to keep all unfinished code in the same place.

jianhuang01:
> I have done a bunch of partial work on getting these to work. Can we move them to `modules/effects/wip` folder instead of deleting?

moved to wip/src/effects

Issue #1112, opened 2019-05-20
Transform based unit tests in deck.gl fail when upgrading Chromimum


1chandu:
Transform uses passthrough Fragment shader :
`'void main() {}';`
This was not an issue before, but with latest chromium builds :
#### Actual Result
[.WebGL-0x7fa0cf002a00]GL ERROR :GL_INVALID_OPERATION : glDrawArrays: buffer format and fragment output variable type incompatible
#### Expected Result
It should compile with out any errors.
#### Reproduce Steps
Upgrade chromium and run deck.gl browser tests.
#### To Do List 
  - [ ] Add label and assign to milestone
  - [ ] Coding
  - [ ] Test

#### To Do List 
  - [x] Add label and assign to milestone 
  - [ ] Coding

PR #1114, opened 2019-05-20
Fix for chromium GLSL compiler issue

1chandu:
For #1112

#### Background
Empty Fragment shader used to work : 
```
'void main() {}'
```
But when upgrading chromium fails all Transform related tests with this error :
```
[.WebGL-0x7fa0cf002a00]GL ERROR :GL_INVALID_OPERATION : glDrawArrays: buffer format and fragment output variable type incompatible
```
Change in this PR fixes the problem.

#### Change List
- Fix for chromium GLSL compiler issue

### DIFF
--- a/modules/shadertools/src/utils/shader-utils.js
+++ b/modules/shadertools/src/utils/shader-utils.js
@@ -1,6 +1,11 @@
 import {assert} from '../utils';
-const FS100 = 'void main() {}';
-const FS300 = `#version 300 es\n${FS100}`;
+const FS100 = 'void main() {gl_FragColor = vec4(0);}';
+const FS300 = `\
+#version 300 es
+out vec4 transform_output;
+void main() {
+  transform_output = vec4(0);
+}`;
 
 // Prase given glsl line and return qualifier details or null
 export function getQualifierDetails(line, qualifiers) {


1chandu:
@tsherif @Pessimistress 
Could this be Chromium specific issue, I didn't find any in google search.

tsherif:
It seems that behavior is correct for [WebGL 2](https://www.khronos.org/registry/webgl/specs/latest/2.0/#5.4):
> If any draw buffer with an attachment does not have a defined fragment shader output, draws generate INVALID_OPERATION, unless all 4 channels of colorMask are set to false.

Odd that there's inconsistency across browsers though...

tsherif:
It was a recent update to the spec: https://github.com/KhronosGroup/WebGL/pull/2780

Pessimistress:
I first observed this issue in Chromium 76. Is this going to be an issue for everyone when Chrome updates? (i.e. do we need to patch 6.x)?

tsherif:
I imagine it won't be an issue unless there's code somewhere that uses GLSL 3 shaders that don't output anything?

1chandu:
> It was a recent update to the spec: [KhronosGroup/WebGL#2780](https://github.com/KhronosGroup/WebGL/pull/2780)

Thanks for the link.
For all TF paths we also set `RASTERIZER_DISCARD` to true but wonder thats not enough. Will merge this PR given this should be safe change.

tsherif:
If RASTERIZER_DISCARD isn't working, I'd say that's a bug. We should file it with Chromium: https://bugs.chromium.org/p/v8/issues/list

PR #1125, opened 2019-05-28
Fix table contents height styling

xintongxia:
#### Background
- the calculation of css max-height for table contents are not considering nested children elements.

#### Change List
- fix max hight calculation
- remove babel and eslint config from package.json and add .eslint.js


website/.eslintrc.js
```diff
@@ -0,0 +1,23 @@
+// prettier-ignore
+module.exports = {
+  plugins: ['react'],
+  extends: ['uber-jsx', 'uber-es2015', 'prettier', 'prettier/react', 'plugin:import/errors'],
+  overrides: {
+    files: ['*.spec.js', 'webpack.config.js', '**/bundle/*.js'],
+    rules: {
+      'import/no-extraneous-dependencies': 0
```
Pessimistress: Why do we need this here

xintongxia: There are errors mentioning `import/no-extraneous-dependencies`
```
Warning: React version not specified in eslint-plugin-react settings. See https://github.com/yannickcr/eslint-plugin-react#configuration .

~/luma.gl/website/webpack.config.js
  2:17  error  'webpack' should be listed in the project's dependencies, not devDependencies                                 import/no-extraneous-dependencies
  7:17  error  'ocular-dev-tools' should be listed in the project's dependencies. Run 'npm i -S ocular-dev-tools' to add it  import/no-extraneous-dependencies

✖ 2 problems (2 errors, 0 warnings)

```

package.json
```diff
@@ -59,6 +59,9 @@
     "eslint-plugin-import": "^2.16.0",
     "eslint-plugin-luma-gl-custom-rules": "file:./dev-modules/eslint-plugin-luma-gl-custom-rules",
     "eslint-plugin-tree-shaking": "^1.7.3",
+    "eslint-config-prettier": "^4.1.0",
```
Pessimistress: This is already a dependency of ocular-dev-tools

website/.eslintrc.js
Pessimistress: These are irrelevant. The website uses root dev dependencies. This lint rule is for validating package.jsons of published modules.

PR #1128, opened 2019-05-29
glfx API audit and documentation

jianhuang01:
For #1105 

#### Change List
- improve glfx doc
- integrate new doc to website


docs/api-reference/effects/glfx-filters.md
```diff
@@ -0,0 +1,343 @@
...
+Saturation is implemented by scaling all color channel values either toward or away from the average color channel value.
+
+* `hue` - -1 to 1 (-1 is 180 degree rotation in the negative direction, 0 is no change, and 1 is 180 degree rotation in the positive direction)
+* `saturation` - -1 to 1 (-1 is solid gray, 0 is no change, and 1 is maximum contrast)
+
+<table style="border: 0;" align="center">
+  <tbody>
+    <tr>
+      <td align="center">
+        <img height=340 src="https://raw.githubusercontent.com/uber-common/deck.gl-data/master/images/samples/glfx/results/brightness.jpg" />
```
Pessimistress: Wrong image

docs/api-reference/effects/glfx-filters.md
```diff
@@ -0,0 +1,343 @@
...
+
+
+### brightnessContrast
+
+Provides additive brightness and multiplicative contrast control.
+
+* `brightness` -1 to 1 (-1 is solid black, 0 is no change, and 1 is solid white)
+* `contrast`   -1 to 1 (-1 is solid gray, 0 is no change, and 1 is maximum contrast)
+
+<table style="border: 0;" align="center">
```
Pessimistress: No need to use HTML if you don't need any custom layout.

`brightnessContrast`

docs/api-reference/effects/glfx-filters.md
```diff
@@ -0,0 +1,343 @@
+# glfx Filters
+
+
+Image effects packaged as reusable shader modules based on the [glfx library](http://evanw.github.io/glfx.js/).
```
Pessimistress: This page does not mention:

- What npm package this is
- How to use the filters

docs/api-reference/effects/glfx-filters.md
jianhuang01: with HTML style it is easier to control the image size and align the `Brightness / Contrast Effect` text with the image.

docs/api-reference/effects/glfx-filters.md
jianhuang01: did some changes, there is no official API in luma.gl to use these effects

ibgreen: > did some changes, there is no official API in luma.gl to use these effects

I guess the `_ShaderModulePass` is still experimental. Still, it is strange to list these with no information on how to use them. Maybe add a link to deck.gl effects page?

jianhuang01: yeah, that's better, will add a link here when the deck.gl effects doc done.

jianhuang01:
> Maybe move the deleted parts of the docs to the WIP folder so that we can restore them when we fix the remaining filters.

done.

docs/api-reference/effects/glfx-filters.md
```diff
@@ -0,0 +1,350 @@
...
+This is mostly a repackaging of shader code from [Evan Wallace](https://github.com/evanw/glfx.js)'s wonderful glfx library.
+
+The code and documentation is included here under MIT license.
+
+## Usage
+
+Import brightnessContrast shader module
+
+```js
+    import brightnessContrast from @luma.gl/effects;
```
Pessimistress: `{brightnessContrast}`

docs/api-reference/effects/glfx-filters.md
```diff
@@ -0,0 +1,350 @@
+# glfx Filters
+
+
+Image effects packaged as reusable shader modules in `@luma.gl/effects` based on the [glfx library](http://evanw.github.io/glfx.js/).
```
Pessimistress: Please structure the documents properly so that we don't break links in the follow up releases.

In documentation tree:

Effects
|- Overview
|- glfx shader modules
|- (future effects in this module)

In "Overview":

The `@luma.gl/effects` module offers modular, reusable image effects.

```
## Installation

    npm install @luma.gl/effects

## Attribution

...
```

In "glfx shader modules":

Everything in "Filters"

Pessimistress: There is no "filter" concept in luma.gl. Let's call these "shader modules" for consistency.

jianhuang01: good suggestion, done.

Pessimistress:
All the options are missing default values.

docs/api-reference/effects/glfx-shader-modules.md
```diff
@@ -0,0 +1,350 @@
...
+
+## Usage
+
+Import brightnessContrast shader module
+
+```js
+    import {brightnessContrast} from @luma.gl/effects;
+```
+
+## Filters
```
Pessimistress: Shader Modules

docs/api-reference/effects/glfx-shader-modules.md
```diff
@@ -0,0 +1,350 @@
...
+
+### hueSaturation
+
+Provides rotational hue and multiplicative saturation control. RGB color space can be imagined as a cube where the axes are the red, green, and blue color values.
+
+Hue changing works by rotating the color vector around the grayscale line, which is the straight line from black (0, 0, 0) to white (1, 1, 1).
+
+Saturation is implemented by scaling all color channel values either toward or away from the average color channel value.
+
+* `hue` - -1 to 1 (-1 is 180 degree rotation in the negative direction, 0 is no change, and 1 is 180 degree rotation in the positive direction)
```
Pessimistress: Remove `-`

docs/api-reference/effects/glfx-shader-modules.md
```diff
@@ -0,0 +1,350 @@
...
+    </tr>
+  </tbody>
+</table>
+
+
+### zoomBlur
+
+Blurs the image away from a certain point, which looks like radial motion blur.
+
+* `centerX`  The x coordinate of the blur origin.
```
Pessimistress: What are the units of this value? Where is the origin?

docs/api-reference/effects/glfx-shader-modules.md
```diff
@@ -0,0 +1,350 @@
...
+</table>
+
+
+### bulgePinch
+
+Bulges or pinches the image in a circle.
+
+* `centerX`  The x coordinate of the center of the circle of effect.
+* `centerY`  The y coordinate of the center of the circle of effect.
+* `radius`   The radius of the circle of effect.
```
Pessimistress: Units

docs/api-reference/effects/glfx-shader-modules.md
```diff
@@ -0,0 +1,342 @@
...
+
+The code and documentation is included here under MIT license.
+
+## Usage
+
+Import brightnessContrast shader module
+
+```js
+    import {brightnessContrast} from @luma.gl/effects;
+```
```
tsherif: This is unclear. Show how to actually use the shader module for rendering, set options etc.

docs/api-reference/effects/glfx-shader-modules.md
```diff
@@ -0,0 +1,342 @@
+# glfx Shader Modules
+
+
+Image effects packaged as reusable shader modules in `@luma.gl/effects` based on the [glfx library](http://evanw.github.io/glfx.js/).
```
tsherif: I'd call out that these are screen-space effect.

docs/api-reference/effects/overview.md
```diff
@@ -0,0 +1,7 @@
+# Overview
+
+The `@luma.gl/effects` module offers modular, reusable effects.
```
tsherif: Call out that these are screen space effects and give a few examples to make it clear what kinds of effects we're talking about.

docs/api-reference/effects/glfx-shader-modules.md
```diff
@@ -0,0 +1,342 @@
...
+    </tr>
+  </tbody>
+</table>
+
+
+### zoomBlur
+
+Blurs the image away from a certain point, which looks like radial motion blur.
+
+* `center`  [x, y] coordinate of the blur origin in pixels. Default value is [0, 0] which is the bottom left corner.
```
Pessimistress: Any chance we can make this coordinate relative to the screen size?

Consider this scenario in deck: how can I create a zoom blur effect that always center in the viewport? I have to listen to the `onResize` event of deck, and create a new `PostProcessEffect` every time, because there is no way for me to update the props of an existing one. This will lead to constant creation of new models.

docs/api-reference/effects/overview.md
Pessimistress: I think the intention is to add non-screenspace effects to this module later.
Maybe start a list of what's in this package here and link to sub-pages?

docs/api-reference/effects/glfx-shader-modules.md
jianhuang01: We don't have official API to use these shader modules in luma.gl, will add a deck.gl link here.

docs/api-reference/effects/glfx-shader-modules.md
jianhuang01: That's how the glfx example implemented, user doesn't resize frequently. We can add a screenSize uniform in all the shaders, and update examples. May not catch up the 7.1 release. Do we want to go this new solution?

jianhuang01: how about use texture size instead of screen size? in our case they are the same.

Pessimistress: > user doesn't resize frequently

This is not true. When the canvas resize is triggered by dragging the window or CSS animation, `onResize` would be called on every animation frame.

modules/effects/src/shader-modules/blur-filters/tiltshift.js
```diff
@@ -36,14 +36,14 @@ vec4 tiltShift_sampleColor(sampler2D texture, vec2 texSize, vec2 texCoord) {
   /* randomize the lookup values to hide the fixed number of samples */
   float offset = random(vec3(12.9898, 78.233, 151.7182), 0.0);
 
-  vec2 normal = normalize(vec2(start.y - end.y, end.x - start.x));
+  vec2 normal = normalize(vec2((start.y - end.y) * texSize.y, (end.x - start.x) * texSize.x));
```
Pessimistress: Nit: it's more efficient to transform `start` and `end` once than multiplying `texSize` on every usage.

PR #1130, opened 2019-05-30
Consolidate injection APIs

tsherif:
A bit of a chunky update to the injection APIs to consolidate the `inject` argument hooks and the module injections. Main thing is that module injections can now inject into prebuilt injection points (e.g. `vs:#main-start`) and the `inject` argument to `assembleShaders` can inject into hook functions defined by `createShaderHook`. Also made a few minor changes to the API to make the two types of injections more consistent, e.g. using a `vs` or `fs` prefix for hook function definitions as well:
```js
createShaderHook('vs:MY_SHADER_HOOK_pickColor(inout vec4 color)');

createModuleInjection('picking', {
  hook: 'vs:MY_SHADER_HOOK_pickColor',
  injection: 'color = picking_setPickingColor(color);',
  order: Number.POSITIVE_INFINITY
});
```

### DIFF
--- a/docs/api-reference/shadertools/assemble-shaders.md
+++ b/docs/api-reference/shadertools/assemble-shaders.md
@@ -98,7 +98,7 @@ And the injection for the picking module would be defined as follows:
 createModuleInjection('picking', {
   hook: 'fs:MYHOOK_fragmentColor',
   injection: 'color = picking_filterColor(color);',
-  priority: Number.POSITIVE_INFINITY
+  order: Number.POSITIVE_INFINITY
 });
 ```
 
@@ -111,7 +111,7 @@ Injecting to a predefined hook would be done as follows:
 createModuleInjection('picking', {
   hook: 'fs:#main-end',
   injection: 'color = picking_filterColor(color);',
-  priority: Number.POSITIVE_INFINITY
+  order: Number.POSITIVE_INFINITY
 });
 ```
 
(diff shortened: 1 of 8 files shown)


docs/api-reference/shadertools/assemble-shaders.md
```diff
@@ -32,31 +32,31 @@ Returns:
...
 * `moduleMap` - a map with all resolved modules, keyed by name
 
-### `setShaderHook(shaderType, opts)`
+### `createShaderHook(hook, [opts])`
 
 Creates a shader hook function that shader modules can injection code into. Shaders can call these functions, which will be no-ops by default. If a shader module injects code it will be executed upon the hook function call. This mechanism allows the application to create shaders that can be automatically extended by included shader modules.
 
-- `shaderType`: `vs` or `fs`, the type of shader
-- `opts.signature`: name and arguments of the function, e.g. `MYHOOK_func(inout vec4 value)`. Name of the function
+- `shaderStage`: `vs` or `fs`, the type of shader
```
Pessimistress: Is this still needed?

modules/shadertools/src/lib/inject-shader.js
```diff
@@ -18,44 +19,52 @@ export default function injectShader(source, type, inject, injectStandardStubs)
...
         break;
       case 'fs:#main-end':
         if (!isVertex) {
-          source = source.replace(REGEX_END_OF_MAIN, match => fragment + match);
+          source = source.replace(REGEX_END_OF_MAIN, match => fragmentString + match);
         }
         break;
 
       default:
+        // TODO(Tarek): I think this usage should be deprecated.
```
Pessimistress: Can we add a test case for this? Several deck.gl layers are still depending on this feature.

tsherif: Done.

PR #1135, opened 2019-05-31
unit tests for @luma.gl/effects

jianhuang01:
For #1105 

#### Change List
- add unit test


tsherif:
Since these are all so small, maybe put them in one spec?

jianhuang01:
> Since these are all so small, maybe put them in one spec?

could add more tests for these shades in the future.

Issue #1147, opened 2019-06-20
Texture: generateMipmap() did not check NPOT


xintongxia:
#### Background 
https://github.com/uber/luma.gl/blob/master/modules/webgl/src/classes/texture.js#L205

I tested this API in Safari, here is the error message.

```WebGL: INVALID_OPERATION: generateMipmap: level 0 not power of 2 or not all the same size```

Also it is confusing to me that these two functions use different fallback logic for NPOT texture.

https://github.com/uber/luma.gl/blob/master/modules/webgl/src/classes/texture.js#L711

https://github.com/uber/luma.gl/blob/master/modules/webgl/src/classes/texture.js#L726

#### To Do List 
  - [ ] Add label and assign to milestone 
  - [ ] Coding
  - [ ] Doc update
  - [ ] Whats new update 
  - [ ] Test

Issue #1146, opened 2019-06-14
Texture resize force mipmaps option to false


xintongxia:
#### Background 

Currently Texture class forces `mipmaps` option to be `false` when calling `resize`, I think it would make more sense to keep `mipmaps` setting as its original value before resizing. `mipmaps = this.mipmaps`

https://github.com/uber/luma.gl/blob/master/modules/webgl/src/classes/texture.js#L198

#### To Do List 
  - [ ] Add label and assign to milestone 
  - [ ] Coding
  - [ ] Doc update
  - [ ] Whats new update 
  - [ ] Test

tsherif:
Oh yeah, that doesn't make any sense.

PR #1148, opened 2019-06-21
Fix mipmap generation

xintongxia:
For #1147 and #1146 

#### Change List
- check mipmap enabled before calling generateMipmaps
- add test

### DIFF
--- a/modules/webgl/src/classes/texture.js
+++ b/modules/webgl/src/classes/texture.js
@@ -195,7 +195,7 @@ export default class Texture extends Resource {
         type: this.type,
         dataFormat: this.dataFormat,
         border: this.border,
-        mipmaps: false
+        mipmaps: this.mipmaps
       });
     }
     return this;
@@ -203,6 +203,15 @@ export default class Texture extends Resource {
 
   // Call to regenerate mipmaps after modifying texture(s)
   generateMipmap(params = {}) {
+    if (!this.mipmaps) {
+      if (this._isNPOT()) {
+        log.warn(`texture: ${this} is Non-Power-Of-Two, disabling mipmaping`)();
+      } else {
+        log.warn('mipmaps is not enabled.');
+      }
+      return this;
+    }
+
     this.gl.bindTexture(this.target, this.handle);
     withParameters(this.gl, params, () => {
       this.gl.generateMipmap(this.target);
--- a/modules/webgl/src/classes/texture.js
+++ b/modules/webgl/src/classes/texture.js
@@ -204,11 +204,7 @@ export default class Texture extends Resource {
   // Call to regenerate mipmaps after modifying texture(s)
   generateMipmap(params = {}) {
     if (!this.mipmaps) {
-      if (this._isNPOT()) {
-        log.warn(`texture: ${this} is Non-Power-Of-Two, disabling mipmaping`)();
-      } else {
-        log.warn('mipmaps is not enabled.');
-      }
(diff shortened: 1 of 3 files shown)


modules/webgl/test/classes/texture.spec.js
```diff
@@ -574,3 +574,44 @@ test('WebGL2#Texture2D setSubImageData', t => {
...
+  });
+
+  texture.resize({
+    width: 4,
+    height: 4
+  });
+
+  t.notOk(texture.mipmaps, 'mipmaps should be the same as before resizing for POT texture.');
+
+  texture.resize({
```
1chandu: I will move this to line 594, so we are testing mipmaps: true -> resize to NPOT -> verify mipmaps is false.

modules/webgl/src/classes/texture.js
```diff
@@ -195,14 +195,23 @@ export default class Texture extends Resource {
...
-        mipmaps: false
+        mipmaps: this.mipmaps
       });
     }
     return this;
   }
 
   // Call to regenerate mipmaps after modifying texture(s)
   generateMipmap(params = {}) {
+    if (!this.mipmaps) {
```
1chandu: Nit: We should have logged NPOT warning already during initialize/resize, and not needed here. 
```
if (!this.mipmaps) {
  log.warn('generateMipmap: mipmaps is not enabled.');
  return this;
}

File: modules/webgl/src/classes/texture.js [hunk 3]
```diff
@@ -195,14 +195,23 @@ export default class Texture extends Resource {
         type: this.type,
         dataFormat: this.dataFormat,
         border: this.border,
-        mipmaps: false
+        mipmaps: this.mipmaps
```
tsherif: This logic doesn't seem correct to me. What if you resize from a POT to NPOT size?

A better solution to me would be to just pass in mipmapping as an additional parameter (default to false).

modules/webgl/src/classes/texture.js
```diff
@@ -195,14 +195,23 @@ export default class Texture extends Resource {
...
   // Call to regenerate mipmaps after modifying texture(s)
   generateMipmap(params = {}) {
+    if (!this.mipmaps) {
+      if (this._isNPOT()) {
+        log.warn(`texture: ${this} is Non-Power-Of-Two, disabling mipmaping`)();
+      } else {
+        log.warn('mipmaps is not enabled.');
+      }
+      return this;
+    }
```
tsherif: It doesn't make sense to me to consider the mipmapping immutable like this. If I don't want to generate mipmaps immediate on initialization, but want to call this method later, there's no reason it should be blocked by the API. Something like this would make more sense to me:

```js
if (this._isNPOT()) {
  log.warn(`texture: ${this} is Non-Power-Of-Two, disabling mipmaping`)();
  return this;
}

this.mipmaps = true;
```

1chandu: 👍

modules/webgl/src/classes/texture.js
xintongxia: @tsherif 

inside `resize`, it calls `initialize`, which will check if the texture is `NPOT` or not and disable mipmap for `NPOT`. But it is a problem if the user resizes the texture from `POT` to `NPOT` and then back to `POT`, the `mipmap` will be false.

tsherif:
Looks great, @xintongxia

PR #1152, opened 2019-06-26
Fix injected declarations

### LINKED ISSUE #1151: Shader hooks should be injected after declarations
#### Actual Result

The `#decl` injection point is appended after the hook definitions, making it impossible to use custom uniforms in the hooks.

#### Expected Result

Hooks should be defined after the custom declarations.

#### Reproduce Steps

```js
assembleShaders(gl, {
  vs: '',
  fs: `
void main() {
  gl_FragColor = vec4(0.0, 0.0, 0.0, 1.0);
  FILTER_COLOR(gl_FragColor);
}
  `,
  inject: {
    'fs:#decl': `uniform bool shouldDraw;`,
    'fs:FILTER_COLOR': `if (!shouldDraw) discard;`
  }
});
```

Returns the following fragment shader:

```glsl
...

precision highp float;

void FILTER_COLOR(vec4 color) {
  if (!shouldDraw) discard;
}

uniform bool shouldDraw;

void main() {
  gl_FragColor = vec4(0.0, 0.0, 0.0, 1.0);
  FILTER_COLOR(gl_FragColor);
}
```

Which results in GLSL compilation error:

```
^^^ ERROR: 'shouldDraw' : undeclared identifier
```

#### To Do List 
  - [x] Add label and assign to milestone
  - [ ] Coding
  - [ ] Test

tsherif:
Ensure that declarations are injected before module functions. Fixes #1151

### DIFF
--- a/modules/shadertools/src/lib/assemble-shaders.js
+++ b/modules/shadertools/src/lib/assemble-shaders.js
@@ -1,9 +1,11 @@
 import {VERTEX_SHADER, FRAGMENT_SHADER} from './constants';
 import {resolveModules, getShaderModule} from './resolve-modules';
 import {getPlatformShaderDefines, getVersionDefines} from './platform-defines';
-import injectShader from './inject-shader';
+import injectShader, {DECLARATION_INJECT_MARKER} from './inject-shader';
 import {assert} from '../utils';
 /* eslint-disable max-depth, complexity */
+
+const INJECT_SHADER_DECLARATIONS = `\n\n${DECLARATION_INJECT_MARKER}\n\n`;
 
 const SHADER_TYPE = {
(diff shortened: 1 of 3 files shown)


modules/shadertools/src/lib/assemble-shaders.js
```diff
@@ -151,6 +151,9 @@ ${isVertex ? '' : FRAGMENT_SHADER_PROLOGUE}
     }
   }
 
+  // For injectShader
+  assembledSource += '\n\n__LUMA_INJECT_DECLARATIONS__\n\n';
```
Pessimistress: Import constant from `inject-shader.js`?

tsherif: Done.

PR #1158, opened 2019-07-03
Data Management RFC

tsherif:
A system to manage data sets and create buffers or textures from them. Supports caching and re-using resources. Would ideally take over some of the complexity that currently lives in deck.gl's `AttributeManager`.

### DIFF
--- a/dev-docs/RFCs/v7.3/data-management-rfc.md
+++ b/dev-docs/RFCs/v7.3/data-management-rfc.md
@@ -12,24 +12,26 @@ This RFC specifies subsystem for managing data in luma.gl and creating gl resour
 
 ## Background
 
-Buffer and texture management in luma.gl is currently done ad hoc manner, with the application required to build these resources from data and to decide when and how they can be re-used. In deck.gl this has lead to the complexity of the `AttributeManager` that creates buffers from input data and manages their lifecycles.
+Buffer and texture management in luma.gl is currently done in an ad hoc manner, with the application required to build these resources from data and to decide when and how they can be re-used. In deck.gl this has lead to the complexity of the `AttributeManager` that creates buffers from input data and manages their lifecycles.
 
 The lack of a global mechanism for managing data resources also has performance implications in that buffers and textures can be allocated to store redundant data. In deck.gl this might occur due to multiple layers being created from the same data (where each layer would allocate its own gl resources) or multiple instances of the same layer each creating their own layer-specific geometry buffers or texture data. The deck.gl `TileLayer` uses a `TileCache` to avoid redundantly creating textures, and a similar mechanism could be applied more generally to avoid allocating textures and buffers if like resources with the same data have already been created anywhere in the application.
 
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


dev-docs/RFCs/v7.3/data-management-rfc.md
```diff
@@ -0,0 +1,96 @@
...
+## Overview
+
+The proposed data manager would allow the application to register arbitrary named data sets, optionally providing accessors to parse particular views of the data. It would provide API methods to create buffers or textures from the
+registered data, returning cached versions if they are available.
+
+
+## Implementation
+
+A `DataManager` class that supports the following methods:
+- `addData(id, data, accessors)`: add a data set to the manager, optionally with accessors to create particular views.
```
Pessimistress: I feel that this is just moving the complexity from `AttributeManager` to `DataManager`. I'd prefer a minimal class that is just `ResourceManager` and not deal with data creation.

dev-docs/RFCs/v7.3/data-management-rfc.md
```diff
@@ -0,0 +1,96 @@
...
+  image2
+], {
+  image1: (data) => data[0],
+  image2: (data) => data[1]
+});
+
+dm.addData('anotherImage', image3); // Accessors not provided because the only view is the entire data set
+
+const positionBuffer = dm.getBuffer('myTable', 'positions');
+const positionBuffer2 = dm.getBuffer('myTable', 'positions'); // Gets cached version of same buffer
```
Pessimistress: In deck.gl `Attribute` class we need the following operations:

- Allocate an empty buffer of a specific size
- Create a new buffer from TypedArray
- Resize an existing buffer (keep the data but allocate bigger size)
- Partially update an existing buffer
- Copy a chunk from one buffer to another

dev-docs/RFCs/v7.3/data-management-rfc.md
tsherif: The issue is that the caching requires identifying when the same data is being used. If the accessors are too much, I could just make it take a typed array or an image (i.e. all the parsing would be done in deck, and then the `value` arrays would be registered with the `DataManager`).

dev-docs/RFCs/v7.3/data-management-rfc.md
tsherif: These should all be doable.

tsherif:
@Pessimistress Got rid of accessors and added the methods you asked for.

dev-docs/RFCs/v7.3/data-management-rfc.md
Pessimistress: In deck.gl we do update a typed array in-place before updating the buffer. It's a poor indicator for using cached result since the content of the same typed array and the same buffer may be out of sync.

tsherif: Hmmm... maybe it would be better to make the API closer to the `ProgramManager` then. Something like `dm.addBuffer('myBuffer', buffer)`, where the application would be responsible for providing a key that uniquely identifies a buffer. And all the other operations would be based on buffer ids. Might be tricky to do caching for data, but things like `Geometry` attributes would be easy to cache.

PR #1162, opened 2019-07-11
support uniform texture and matrix array

jianhuang01:
#### Background
Need support binding texture and matrix array

#### Change List
- add uniform names for array

### DIFF
--- a/modules/webgl/src/classes/program.js
+++ b/modules/webgl/src/classes/program.js
@@ -370,8 +370,14 @@ export default class Program extends Resource {
     for (let i = 0; i < this._uniformCount; i++) {
       const info = this.gl.getActiveUniform(this.handle, i);
       const {name, isArray} = parseUniformName(info.name);
-      const location = gl.getUniformLocation(this.handle, name);
+      let location = gl.getUniformLocation(this.handle, name);
       this._uniformSetters[name] = getUniformSetter(gl, location, info, isArray);
+      if (info.size > 1) {
+        for (let l = 0; l < info.size; l++) {
+          location = gl.getUniformLocation(this.handle, `${name}[${l}]`);
+          this._uniformSetters[`${name}[${l}]`] = getUniformSetter(gl, location, info, isArray);
+        }
+      }
     }
     this._textureIndexCounter = 0;
   }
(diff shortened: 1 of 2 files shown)


Pessimistress:
Add a unit test?

jianhuang01:
> Add a unit test?

of course.

modules/webgl/src/classes/program.js
```diff
@@ -370,8 +370,14 @@ export default class Program extends Resource {
     for (let i = 0; i < this._uniformCount; i++) {
       const info = this.gl.getActiveUniform(this.handle, i);
       const {name, isArray} = parseUniformName(info.name);
-      const location = gl.getUniformLocation(this.handle, name);
+      let location = gl.getUniformLocation(this.handle, name);
       this._uniformSetters[name] = getUniformSetter(gl, location, info, isArray);
+      if (info.size > 1) {
+        for (let l = 0; l < info.size; l++) {
+          location = gl.getUniformLocation(this.handle, `${name}[${l}]`);
```
tsherif: I think it would be better to load data for the entire array at once, rather than treating them as separate uniforms. I.e. these two are equivalent for a uniform `uniform vec2 myVec2s[2]`
```js
const location = gl.getUniformLocation(program, "myVec2s");
gl.uniform2fv(location, new Float32Array([1, 2, 3, 4]);
```

```js
const location1 = gl.getUniformLocation(program, "myVec2s[0]");
gl.uniform2fv(location1, new Float32Array([1, 2]);
const location2 = gl.getUniformLocation(program, "myVec2s[1]");
gl.uniform2fv(location2, new Float32Array([3, 4]);
```

But I prefer the former because it involves fewer API calls.

jianhuang01: this is not only for texture array but also for matrix array, for matrix array we need bind them one by one anyway, so this change is need anyway. For texture array there are ways to improve just as you proposed, we can add a new change on the texture binding part later on because there are more changes involved.

tsherif: Matrices can be handled in the same way, you pass the data for all matrices at the same time.

The texture case is more complicated though, since we'd need a change in luma to take an array of textures and turn it into an array of texture unit values. So I'm ok with this update.

PR #1164, opened 2019-07-17
Fxaa

tsherif:
FXAA implementation based the one in Cesium (which was tweaked to work in GLSL). I modified it by upping the quality and changing the luma calculation based on RGB rather than just green.

Before:
Screen Shot 2019-07-18 at 4 35 49 PM

After:
Screen Shot 2019-07-18 at 4 35 57 PM

### DIFF
--- a/modules/effects/src/shader-modules/fxaa/fxaa.js
+++ b/modules/effects/src/shader-modules/fxaa/fxaa.js
@@ -85,7 +85,7 @@ const fs = `
 //  _ = the lowest digit is directly related to performance
 // _  = the highest digit is directly related to style
 //
-#define FXAA_QUALITY_PRESET 12
+#define FXAA_QUALITY_PRESET 23
 
 
 #if (FXAA_QUALITY_PRESET == 10)
@@ -276,7 +276,7 @@ const fs = `
 #define FxaaTexTop(t, p) texture2D(t, p)
 #define FxaaTexOff(t, p, o, r) texture2D(t, p + (o * r))
 
-FxaaFloat FxaaLuma_(FxaaFloat4 rgba) { return rgba.y; }
+FxaaFloat FxaaLuma_(FxaaFloat4 rgba) { return dot(rgba.rgb, vec3(0.2126, 0.7152, 0.0722)); }
 
 FxaaFloat4 FxaaPixelShader_(
     //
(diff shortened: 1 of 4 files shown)


tsherif:
Ah, I think I know the issue. FXAA requires linear filtering on the texture it's sampling from. I'll test that tomorrow and make the necessary changes.

tsherif:
@Pessimistress 	@jianhuang01 Fixed the remaining issues. This is ready now.

modules/effects/src/shader-modules/fxaa/fxaa.js
```diff
@@ -0,0 +1,681 @@
...
+// -----------------------------------------------------------------------
+// 12 = slightly faster then FXAA 3.9 and higher edge quality (default)
+// 13 = about same speed as FXAA 3.9 and better than 12
+// 23 = closest to FXAA 3.9 visually and performance wise
+//  _ = the lowest digit is directly related to performance
+// _  = the highest digit is directly related to style
+//
+
+const fs = `
+#define FXAA_QUALITY_PRESET 29
```
Pessimistress: Should this be supplied by the user of this module?

modules/webgl/src/classes/framebuffer.js
```diff
@@ -433,10 +433,11 @@ export default class Framebuffer extends Resource {
         // behavior we always disable mipmaps.
         mipmaps: false,
         // Set MIN and MAG filtering parameters so mipmaps are not used in sampling.
+        // Use LINEAR so subpixel algos like fxaa work.
         // Set WRAP modes that support NPOT textures too.
         parameters: {
-          [GL.TEXTURE_MIN_FILTER]: GL.NEAREST,
-          [GL.TEXTURE_MAG_FILTER]: GL.NEAREST,
+          [GL.TEXTURE_MIN_FILTER]: GL.LINEAR,
+          [GL.TEXTURE_MAG_FILTER]: GL.LINEAR,
```
Pessimistress: Can we avoid changing the defaults here? Is this going to break e.g. the picking buffer? Add to documentation?

modules/effects/src/shader-modules/fxaa/fxaa.js
tsherif: Could be, but I've found the difference to be fairly subtle. Don't feel it would really be a meaningful option.

modules/webgl/src/classes/framebuffer.js
tsherif: My preference would be to update the defaults to avoid complicating the usage in deck, where you'd have to define the framebuffer attachments explicitly ([like this](https://github.com/uber/luma.gl/blob/master/examples/core/dof/app.js#L300-L316)).
It won't change anything other than post-processing, since it only affects sampling from framebuffer attachments. It won't break picking, because texture filtering doesn't affect `readPixels`.

I'll update the documentation, though, to reflect the new defaults.

tsherif:
@Pessimistress 

Before:
Screen Shot 2019-07-18 at 4 35 49 PM

After:
Screen Shot 2019-07-18 at 4 35 57 PM

modules/webgl/src/classes/framebuffer.js
1chandu: Also add an entry in upgrade guide.

PR #1166, opened 2019-07-18
Add unit tests for all Scenegraph classes

1chandu:
For #
Improve coverage for Core Module:
Add unit tests for:
- 'GroupNode`
- `SceneGraphNode`
- `CameraNode`
- `ModelNode`
- `format-value` utility methods.

.. in the process found and fixed following issues:

- `Model`: Fix a crash, use default `props` object
- `GroupNode` : Fix child node check, to make sure all children are `ScenegraphNode` instances
- `SceneGraphNode` : Fix a crash, use default `props` object
- `SceneGraphNode` : Comment out un-used code
- `SceneGraphNode` : Fix `setRotation` method.

#### Change List
-

### DIFF
--- a/modules/core/src/scenegraph/nodes/group-node.js
+++ b/modules/core/src/scenegraph/nodes/group-node.js
@@ -1,12 +1,13 @@
 import {Matrix4} from 'math.gl';
-import {log, assert} from '../../utils';
+import {log} from '../../utils';
 import ScenegraphNode from './scenegraph-node';
 
 export default class GroupNode extends ScenegraphNode {
   constructor(props = {}) {
     props = Array.isArray(props) ? {children: props} : props;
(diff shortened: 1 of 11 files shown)


modules/core/src/scenegraph/nodes/group-node.js
```diff
@@ -6,7 +6,7 @@ export default class GroupNode extends ScenegraphNode {
   constructor(props = {}) {
     props = Array.isArray(props) ? {children: props} : props;
     const {children = []} = props;
-    children.every(child => assert(child instanceof ScenegraphNode));
+    assert(children.every(child => child instanceof ScenegraphNode));
```
Pessimistress: Use `log.assert` and drop the util

PR #1171, opened 2019-07-19
Geometry bug fixes & tests

Pessimistress:
#### Change List
- Fix bugs in core/geometry
- Unit tests


modules/core/src/geometry/geometry-utils.js
```diff
@@ -1,33 +1,29 @@
...
-    unpackedPositions[x * 3 + 2] = POSITION[index * 3 + 2];
-    unpackedNormals[x * 3 + 0] = NORMAL[index * 3 + 0];
-    unpackedNormals[x * 3 + 1] = NORMAL[index * 3 + 1];
-    unpackedNormals[x * 3 + 2] = NORMAL[index * 3 + 2];
-    unpackedTexCoords[x * 2 + 0] = TEXCOORD_0[index * 2 + 0];
-    unpackedTexCoords[x * 2 + 1] = TEXCOORD_0[index * 2 + 1];
+  for (const attributeName in attributes) {
+    const attribute = attributes[attributeName];
+    const {constant, value, size} = attribute;
+    if (constant || !size) {
```
tsherif: You can remove the `size` requirement by getting the number of vertices via `attribute.value.length / attribute.size` for any attribute, or `max(indices) / verticesPerPrimitive`, then taking `attribute.value.length / vertexCount` as the size.

PR #1173, opened 2019-07-20
Clean up normalizeShaderModule

Pessimistress:
#### Change List
- Clean up uniform type system
- Unit tests


modules/shadertools/src/lib/shader-module.js
```diff
@@ -10,7 +11,8 @@ export default class ShaderModule {
     vs,
     fs,
     dependencies = [],
-    getUniforms = () => ({}),
+    uniforms,
```
1chandu: Add a note to `writing-shader-module.md`

PR #1185, opened 2019-07-26
Clean up model classes

Pessimistress:
#### Change List
- Remove duplicate code between `BaseModel` and `Model` classes
- Clean up unreachable code
- Unit tests


modules/core/src/lib/model-utils.js
```diff
@@ -52,9 +60,6 @@ function mapAttributeName(name, options) {
 export function inferAttributeAccessor(attributeName, attribute) {
   let category;
   switch (attributeName) {
-    case 'indices':
```
1chandu: should we log a deprecated message/warning?

Pessimistress: The use case is still handled, I just moved the code.

PR #1188, opened 2019-07-27
Transform: fix resource leak

1chandu:
For #

#### Background
- Fix a `Texture2D` resource leak, found while working on [issue](https://github.com/uber/deck.gl/issues/2)
- Add unit tests

#### Change List
- Transform: fix resource leak


1chandu:
Coverage drop report seems to be invalid , merging.

PR #1199, opened 2019-08-14
Implement program manager

tsherif:
- Implement `ProgramManager` class proposed in [Program Management RFC](https://github.com/uber/luma.gl/blob/master/dev-docs/RFCs/v7.3/program-manager-rfc.md)
- `BaseModel` updates
  - Support acquiring program from a `ProgramManager`.
  - Support program sharing
  - Support updating program via `setProgram` method
- `ProgramManager` example

### DIFF
--- a/docs/api-reference/core/resource-management/program-manager.md
+++ b/docs/api-reference/core/resource-management/program-manager.md
@@ -1,0 +1,104 @@
+# ProgramManager
+
+The `ProgramManger` manages the creation and caching of programs. It allows the application to request a program based on a vertex shader, fragment shader and set of defines, modules and code injections. The `ProgramManager` will return the requested program, creating it the first time, and re-using a cached version if it is requested more than once. It also allows for the definition of hook functions and module code injections to be inserted into shaders.
+
+
+## Usage
+
+```js
+const pm = new ProgramManager(gl);
+
+const vs = `
+attribute vec4 position;
+
+void main() {
+#ifdef MY_DEFINE
+  gl_Position = position;
+#else
+  gl_Position = position.wzyx;
+#endif
+}
+`;
+
+const fs = `
+void main() {
+  gl_FragColor = vec4(1.0, 0.0, 0.0, 1.0);
+  MY_SHADER_HOOK(gl_FragColor);
+}
+`;
+
+pm.addShaderHook('fs:MY_SHADER_HOOK(inout vec4 color)');
+
+pm.addModuleInjection(picking, {
+  hook: 'fs:#MY_SHADER_HOOK',
+  injection: 'color = picking_filterColor(color);',
+  order: Number.POSITIVE_INFINITY
+});
+
+const program1 = pm.get(vs, fs);   // Basic, no modules or defines
+const program2 = pm.get(vs, fs);   // Cached, same as program 1, use count 2
+const program3 = pm.get(vs, fs, {  // New program, with different source based on define
+  defines: {
+    MY_DEFINE: true
+  }
+});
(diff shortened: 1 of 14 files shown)


docs/api-reference/core/resource-management/program-manager.md
```diff
@@ -0,0 +1,104 @@
...
+
+pm.addModuleInjection(picking, {
+  hook: 'fs:#MY_SHADER_HOOK',
+  injection: 'color = picking_filterColor(color);',
+  order: Number.POSITIVE_INFINITY
+});
+
+const program1 = pm.get(vs, fs);   // Basic, no modules or defines
+const program2 = pm.get(vs, fs);   // Cached, same as program 1, use count 2
+const program3 = pm.get(vs, fs, {  // New program, with different source based on define
```
Pessimistress: Can we align this API with that of Model, Program and Shader (fs and vs as part of the named parameters).

modules/core/src/resource-management/program-manager.js
```diff
@@ -0,0 +1,120 @@
...
+      fs: {}
+    };
+
+    this._hashes = {};
+    this._hashCounter = 0;
+    this._useCounts = {};
+  }
+
+  addModuleInjection(module, opts) {
+    const moduleName = module.name;
```
Pessimistress: - What about default shader modules?
- to dedup this and the existing global registration logic, can we create a default program manager and expose its methods for backward compatibility?

docs/api-reference/core/resource-management/program-manager.md
tsherif: Sure, that makes sense.

modules/core/src/resource-management/program-manager.js
tsherif: > What about default shader modules?
What I was imagining is deck could just track what it considers the default shader modules and request programs that include them on each draw, e.g.
```js
// In some layer
draw({uniforms}) {
	this.state.model.setProgram({
       vs,
       fs,
       defaultModules.concat(layerModules)
    });
    this.state.model.draw();
}
```
> to dedup this and the existing global registration logic, can we create a default program manager and expose its methods for backward compatibility?

Is there a way you can think of that wouldn't create a dependency between shadertools and the program manager? I'll add deprecation warning to the global functions in the meantime, since I think they should be removed in 8.

Pessimistress: We can do whatever under the hood (LayerManager or Layer class), but I feel strongly about not breaking the existing layers. The current standard implementation:

```js
draw({uniforms}) {
	this.state.model.draw({uniforms});
}
```

Should continue to work. Many users write custom layers and we should not add complexity to this process.

tsherif: Sounds good. I think we can use `Layer.drawLayer` and `Layer.getShaders` to get the kind of functionality I'm aiming for. Can we land this implementation as-is, and I'll dig into how things will hook up in deck.

modules/core/src/resource-management/program-manager.js
```diff
@@ -0,0 +1,128 @@
...
+      varyings = [],
+      bufferMode = 0x8c8d
+    } = props; // varyings/bufferMode for xform feedback, 0x8c8d = SEPARATE_ATTRIBS
+
+    const vsHash = this._getHash(vs);
+    const fsHash = this._getHash(fs);
+    const moduleHashes = modules.map(m => this._getHash(m.name)).sort();
+    const varyingHashes = varyings.map(v => this._getHash(v));
+
+    const defineKeys = Object.keys(defines).sort();
```
Pessimistress: Nit: make a utility function to hash a dictionary?

PR #1207, opened 2019-08-21
Add DOUBLE to gl constants

Pessimistress:
For https://github.com/uber/deck.gl/pull/3468#discussion_r315899498

Source: https://www.khronos.org/registry/OpenGL/api/GL/glcorearb.h

#### Change List
- Add `DOUBLE` to `@luma.gl/constants`


tsherif:
`DOUBLE` is not a WebGL constant. Why is this being added?

PR #1216, opened 2019-08-28
Program manager hooks

tsherif:
Ensure that changing the hook functions on a program manager changes the program hashes.


tsherif:
> I am a bit out of the loop, but when reading wondered what does it mean that hooks are modified? Does that mean that programs saved in the cache under the old hashes are no longer ever going to be used again? I.e. should perhaps the cache just be flushed, rather than keep the old keys around in the cache?

Not a usage I'm really expecting, but wanted to safeguard against confusing behavior if, for example: You get some programs, add hooks/injections, then get some new programs. Without changing the hash before and after the hooks are added, you'd get the old cached programs without the hooks.

The previously created programs do have to be maintained so they can be released properly.

PR #1219, opened 2019-08-28
RFC: Transform Refactor

1chandu:
For #1193

#### Background
Adding RFC for `Transform` class refactor.

#### Change List
- RFC: Transform Refactor

Issue #1193, opened 2019-08-07
Transform: split into multiple classes


1chandu:
Current version of `Transform` class supports both `Buffers` and `Textures` as source and targets. Though this is convenient for users, it makes `Texture` only usage limited by `WebGL2`, by splitting `Texture` API into separated class this can be avoided. Also `Transform`'s `swap` features can be separated from the base class to distribute code complexity into multiple modules.

This task covers following sub tasks:

- [ ] Class supporting only `Textures`, this class can be moved to `gpgpu` module
- [ ] Class just performing single TF loop (`Buffers` only), this class can remain in `core` module
- [ ] Class supporting  `Buffers` and `Textures` , with `swap` feature, this class can be moved to `gpgpu` module.
- [ ] Updating all Transform use cases to above changes (Unit Tests, Wind Demo, Attribute Transitions, Histopyramid utility methods and Heatmap Layer)

Other issues to be addressed

- [ ] Remove auto creation of resources, instead provide this code as `Buffer` and `Texture` cloning utility methods.
- [ ] Make `_targetTextureVarying` optional, since users might want to not finalize the texture data in `vs`
- [ ] Texture parameters are updated (to NEAREST) so correct pixel is sampled, restore the parameter setting once processed.

1chandu:
Phase-1 as per RFC is complete , Phase-2 will be a breaking change and targeting for 8.0.

1chandu:
Phase-2 is pending, moving to backlog

PR #1221, opened 2019-08-28
Transform refactor

1chandu:
For #1193

#### Background
Phase#1 changes as described in [RFC](https://github.com/uber/luma.gl/pull/1219)
This is still WIP, I will be collecting team's feedback on RFC and update this PR.
Verified using existing `Transform` unit tests and `core/transform` example.
Remaining items:
- Addressing any feedback from RFC
- Docs
- Unit tests specific to new modules/classes.

#### Change List
- [POC] Transform refactor


1chandu:
> perhaps writing/updating markdown docs as you go could help reviewers a lot.

I did write how these classes work and how they interact in RFC, I will write docs after collecting first round of feedback to make sure overall approach.

> One question with this design (I need to reread the RFC) is whether the user can combine buffers and textures in one call, or whether it always uses one type.

The code as is, yes you can combine buffer and textures in one call. I discussed few options in RFC for Phase#2, some options support and some options don't, but this PR doesn't covers Phase#2, just lays foundations for it.

> I count ~800 lines, so this is class is getting pretty big, and I suspect it will keep growing.
> It raises the question whether this should be kept in core (it probably should stay in core, but I think we want to keep a close eye on module growth).

Yep, these are the goals as describe in RFC is to divide this class, more details in RFC.

modules/core/src/lib/transform/buffer-transform-binding.js
```diff
@@ -0,0 +1,47 @@
+import {isObjectEmpty} from '../../utils';
+
+import {TransformFeedback} from '@luma.gl/webgl';
+
+// swapping of two instances, instead of maintaing two arra
```
Pessimistress: Incomplete documentation?

modules/core/src/lib/transform/buffer-transform.js
```diff
@@ -0,0 +1,176 @@
...
+  // Delete owned resources.
+  delete() {
+    for (const name in this.resources) {
+      this.resources[name].delete();
+    }
+  }
+
+  // Private
+
+  initialize(props = {}) {
```
Pessimistress: Use underscore prefix for private method names

modules/core/src/lib/transform/texture-transform.js
```diff
@@ -0,0 +1,304 @@
...
+      discard = false;
+      parameters.viewport = [0, 0, framebuffer.width, framebuffer.height];
+    }
+
+    return Object.assign(opts, {attributes, framebuffer, uniforms, discard, parameters});
+  }
+
+  swap() {
+    if (this._swapTexture) {
+      this.currentIndex = (this.currentIndex + 1) % 2;
```
Pessimistress: `this.getNextIndex()`?

modules/core/src/lib/transform/transform.js
```diff
@@ -0,0 +1,166 @@
...
+    this.elementIDBuffer = null;
+    this.initialize(props);
+    Object.seal(this);
+  }
+
+  // Delete owned resources.
+  delete() {
+    this.model.delete();
+    /* eslint-disable no-unused-expressions */
+    this.bufferTransform && this.bufferTransform.delete();
```
Pessimistress: Nit: use `if`

modules/core/src/lib/transform/transform.js
```diff
@@ -0,0 +1,166 @@
...
+  getFramebuffer() {
+    return this.textureTransform && this.textureTransform.getFramebuffer();
+  }
+
+  // Update some or all buffer/texture bindings.
+  update(opts = {}) {
+    if (opts.elementCount) {
+      this.model.setVertexCount(opts.elementCount);
+    }
+    const resourceTransforms = [this.bufferTransform, this.textureTransform].filter(Boolean);
```
Pessimistress: Make a helper function?

modules/core/src/lib/transform/transform.js
```diff
@@ -0,0 +1,166 @@
...
+    if (!isObjectEmpty(props.sourceBuffers)) {
+      return true;
+    }
+    if (!isObjectEmpty(props.feedbackBuffers)) {
+      return true;
+    }
+    return false;
+  }
+
+  canCreateTextureTransform(props) {
```
Pessimistress: These two can be plain functions instead of class methods - method names cannot be minified and inflate bundle size.

1chandu:
Addressed RFC Feedback :

- Merge `BufferTransformBinding` into `BufferTransform` and `TextureTransformBinding` into `TextureTransform`
- Add docs for `BufferTransform` and `TextureTransform`, didn't link them to website yet, since these are still private internal classes, added for easier review.
- Address other review comments.

docs/api-reference/core/transform/texture-transform.md
```diff
@@ -0,0 +1,67 @@
...
+  * `sourceTextures` (`Object`, Optional) - key and value pairs, where key is the name of vertex shader attribute and value is the corresponding `Texture` object.
+  * `targetTexture` (`Texture`|`String`, Optional) - `Texture` object to which data to be written. When it is a `String`, it must be one of the source texture attributes name, a new texture object is cloned from it.
+  * `targetTextureVarying` (`String`) : varying name used in vertex shader who's data should go into target texture.
+  * `swapTexture` (`String`) : source texture attribute name, that is swapped with target texture every time `swap()` is called.
+  * `fs` (`String`, Optional) - fragment shader string, when rendering to a texture, fragments can be processed using this custom shader, when not specified, pass through fragment shader will be used.
+
+
+## Methods (Model props)
+
+### getDrawOptions(opts: Object) : Object
```
1chandu: I feel, `updateDrawOptions` is more suitable, here and for `BufferTransform`.

docs/api-reference/core/transform/buffer-transform.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Transform(gl : WebGL2RenderingContext, props: Object)
+
+* `gl` (`WebGLRenderingContext`) gl - context
+* `props` (`Object`, Optional) - contains following data.
+  * `sourceBuffers` (`Object`, Optional) - key and value pairs, where key is the name of vertex shader attribute and value is the corresponding `Attribute`, `Buffer` or attribute descriptor object.
+  * `feedbackBuffers` (`Object`, Optional) - key and value pairs, where key is the name of vertex shader varying and value is the corresponding `Buffer` object or buffer params object. If a buffer params object is specified, it will contain following fields, these can be used to capture data into the buffer at particular offset and size.
+    * `buffer`=(Buffer) - Buffer object to be bound.
+    * `byteOffset`=(Number, default: 0) - Byte offset that is used to start recording the data in the buffer.
+    * `byteSize`=(Number, default: remaining buffer size) - Size in bytes that is used for recording the data.
```
Pessimistress: Maybe not for this PR: Can we follow the same accessor convention that `model.setAttributes` uses?

docs/api-reference/core/transform/texture-transform.md
Pessimistress: It's quite awkward how the props/options are passed from the parent class to the child then back again. Shouldn't `BufferTransform` and `TextureTransform` own their models instead of `Transform`?

1chandu: as discussed offline, these are internal methods to Transform, I will address these concerns in Phase-2, where I am planning to make this API public.

PR #1226, opened 2019-09-03
Program manager updates

tsherif:
- BaseModel: avoid program manager hash checks when not necessary
- Default program manager


tsherif:
> should all classes have managed flags for all resources, or is it OK long term that this class does things differently? Generally we want as few surprises/differences between similar resource managements situations as possible.

I'm curious about this too. The question for me in this particular case is whether we still need the option to create programs externally to the model and pass them in. If not, the need for this kind of flag goes away. We'll be able to figure out what the more general patterns are as we start using these systems more.

PR #1268, opened 2019-10-01
Add attribute picking to the picking module

Pessimistress:
For https://github.com/uber/deck.gl/pull/3721

#### Background

Move generic-purpose shader code into the picking module.

#### Change List
- Add `pickingAttribute` module option
- Add `picking_setPickingAttribute` vertex shader function

### DIFF
--- a/modules/shadertools/src/modules/picking/picking.js
+++ b/modules/shadertools/src/modules/picking/picking.js
@@ -30,60 +30,63 @@ function getUniforms(opts = DEFAULT_MODULE_OPTIONS) {
 }
 
 const vs = `\
+uniform bool picking_uActive;
 uniform bool picking_uAttribute;
 uniform vec3 picking_uSelectedColor;
 uniform bool picking_uSelectedColorValid;
 
-out vec4 picking_vRGBcolor_Aselected;
+out vec4 picking_vRGBcolor_Avalid;
 
 const float COLOR_SCALE = 1. / 255.;
+
+bool picking_isColorValid(vec3 color) {
+  return dot(color, vec3(1.0)) > 0.001;
+}
 
 bool isVertexPicked(vec3 vertexColor) {
   return
     picking_uSelectedColorValid &&
-    abs(vertexColor.r - picking_uSelectedColor.r) < 0.001 &&
-    abs(vertexColor.g - picking_uSelectedColor.g) < 0.001 &&
-    abs(vertexColor.b - picking_uSelectedColor.b) < 0.001;
+    !picking_isColorValid(abs(vertexColor - picking_uSelectedColor));
 }
 
 void picking_setPickingColor(vec3 pickingColor) {
-  if (picking_uAttribute) {
+  if (picking_uActive) {
     // Use alpha as the validity flag. If pickingColor is [0, 0, 0] fragment is non-pickable
-    picking_vRGBcolor_Aselected.a = step(0.001, dot(pickingColor, vec3(1.0)));
+    picking_vRGBcolor_Avalid.a = float(picking_isColorValid(pickingColor));
+
+    if (!picking_uAttribute) {
+      // Stores the picking color so that the fragment shader can render it during picking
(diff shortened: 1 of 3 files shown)


modules/shadertools/src/modules/picking/picking.js
```diff
@@ -48,44 +41,65 @@ const float COLOR_SCALE = 1. / 255.;
...
+    abs(vertexColor.b - picking_uSelectedColor.b) < 0.001;
 }
 
 void picking_setPickingColor(vec3 pickingColor) {
-  // Do the comparison with selected item color in vertex shader as it should mean fewer compares
-  picking_vRGBcolor_Aselected.a =
-    float(isVertexPicked(pickingColor));
+  if (picking_uAttribute) {
+    // Use alpha as the validity flag. If pickingColor is [0, 0, 0] fragment is non-pickable
+    picking_vRGBcolor_Aselected.a = step(0.001, length(pickingColor));
```
tsherif: You can do `dot(pickingColor, vec3(1.0))` instead of `length(pickingColor)` here. Avoid the sqrt.

modules/shadertools/src/modules/picking/picking.js
```diff
@@ -96,11 +110,20 @@ vec4 picking_filterHighlightColor(vec4 color) {
...
-    discard;
+  if (picking_uAttribute) {
+    if (picking_vRGBcolor_Aselected.a == 0.0) {
+      discard;
+    }
+    return picking_vRGBcolor_Aselected;
+  }
+  if (picking_uActive) {
+    vec3 pickingColor = picking_vRGBcolor_Aselected.rgb;
+    if (length(pickingColor) < 0.001) {
```
tsherif: Like above. `dot(pickingColor, vec3(1.0))`

modules/shadertools/src/modules/picking/picking.js
```diff
@@ -48,44 +41,65 @@ const float COLOR_SCALE = 1. / 255.;
 bool isVertexPicked(vec3 vertexColor) {
   return
     picking_uSelectedColorValid &&
-    abs(vertexColor.r - picking_uSelectedColor.r) < picking_uThreshold &&
-    abs(vertexColor.g - picking_uSelectedColor.g) < picking_uThreshold &&
-    abs(vertexColor.b - picking_uSelectedColor.b) < picking_uThreshold;
```
tsherif: Probably not used much, but this is a breaking change?

Pessimistress: This is not used anywhere in luma or deck.

modules/shadertools/src/modules/picking/picking.js
```diff
@@ -48,44 +41,65 @@ const float COLOR_SCALE = 1. / 255.;
...
+    abs(vertexColor.r - picking_uSelectedColor.r) < 0.001 &&
```
1chandu: Define a const , for `0.001`.

modules/shadertools/src/modules/picking/picking.js
1chandu: Are we changing the behavior, i.e. isn't it `picking_uThreshold` 1.0 by default, and now it is `0.001`, should look out for any regressions if so and make a note in the code.

Pessimistress: What use case does this break? `0.001` works for both normalized and non-normalized colors.

Pessimistress:
> when writing the attribute, is pickingColor same for all instances, if so we are actually writting same value for all pixels?

I don't understand this question. `pickingColor` is per vertex.

> Can we use existing highlighting logic, does it work if we set attribute value as highlight color and re-run the picking pass (may have to adjust some uniforms) so that only picked object color is written to Framebuffer.

No. The highlight color must be provided as a uniform. We don't have the "attribute value" on the CPU, that's the thing we are trying to pick/query.

Issue #1258, opened 2019-09-24
Cache FEATURE query results in a global scope


1chandu:
- `hasFeature` and `hasFeatures` query for a feature as per the FEATURES table.
- `getFeatures` queries all the features and saves the results in global scope, `gl.luma.caps`

During `hasFeature` and `hasFeatures` we should first check the cache , if not available query, save the result to cache and return. Also update `getFeatures` to only query for non cached FEATURES and save the results.

PR #1270, opened 2019-10-04
WebGL Features: add caching

1chandu:
For #1258 

#### Background
More details in above issue.

#### Change List
-

### DIFF
--- a/modules/webgl/src/features/features.js
+++ b/modules/webgl/src/features/features.js
@@ -39,27 +39,20 @@ function isFeatureSupported(gl, cap) {
   gl.luma = gl.luma || {};
   gl.luma.caps = gl.luma.caps || {};
 
-  let isSupported;
-
-  if (gl.luma.caps[cap] !== undefined) {
-    isSupported = gl.luma.caps[cap];
-  } else {
-    isSupported = queryFeature(gl, cap);
-    gl.luma.caps[cap] = isSupported;
+  if (gl.luma.caps[cap] === undefined) {
+    gl.luma.caps[cap] = queryFeature(gl, cap);
   }
 
-  if (!isSupported) {
+  if (!gl.luma.caps[cap]) {
     log.log(LOG_UNSUPPORTED_FEATURE, `Feature: ${cap} not supported`)();
   }
 
-  return isSupported;
+  return gl.luma.caps[cap];
 }
 
 function queryFeature(gl, cap) {
   const feature = WEBGL_FEATURES[cap];
   assert(feature, cap);
-  // Shouldn't be calling if it is already queried
-  assert(gl.luma.caps[cap] === undefined);
 
   let isSupported;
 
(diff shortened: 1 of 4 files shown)


modules/webgl/src/features/features.js
```diff
@@ -37,14 +36,36 @@ export function getFeatures(gl) {
 
 // TODO - cache the value
 function isFeatureSupported(gl, cap) {
+  gl.luma = gl.luma || {};
+  gl.luma.caps = gl.luma.caps || {};
+
+  let isSupported;
+
+  if (gl.luma.caps[cap] !== undefined) {
```
1chandu: @tsherif , I was thinking, for feature overrides, we can add `gl.luma.overrideCaps` and check that here first and if set use it.  We can have `setFeatureOverrides(caps)` and `clearFeatureOverrides()` methods to update `gl.luma.overrideCaps`

modules/webgl/src/features/features.js
```diff
@@ -37,14 +36,36 @@ export function getFeatures(gl) {
...
+  }
+
+  return isSupported;
+}
+
+function queryFeature(gl, cap) {
   const feature = WEBGL_FEATURES[cap];
   assert(feature, cap);
+  // Shouldn't be calling if it is already queried
+  assert(gl.luma.caps[cap] === undefined);
```
tsherif: This is unnecessary since this function is only called in one place (also an exception is overkill for something like this).

modules/webgl/src/features/features.js
```diff
@@ -37,14 +36,36 @@ export function getFeatures(gl) {
...
+  gl.luma.caps = gl.luma.caps || {};
+
+  let isSupported;
+
+  if (gl.luma.caps[cap] !== undefined) {
+    isSupported = gl.luma.caps[cap];
+  } else {
+    isSupported = queryFeature(gl, cap);
+    gl.luma.caps[cap] = isSupported;
+  }
```
tsherif: You can simplify the conditional a bit.

```js
  if (gl.luma.caps[cap] === undefined) {
    gl.luma.caps[cap] = queryFeature(gl, cap);
  }

  isSupported = gl.luma.caps[cap];
```

modules/webgl/src/features/features.js
tsherif: Yeah this is a nice idea. I could imagine an API like this:

```js

withFeatureOverrides(caps, () => {
    // Do your thing
});
```
This could use the cache to temporarily override feature checks. (I'd worry about allowing feature overrides to exist long term as it could lead to super hard-to-track-down bugs).

1chandu: provided we are querying and drawing within that block. We might still need individual methods to set/re-set overrides for uses cases such as deck.gl Layers, where querying happens during layer init time and drawing happens later.

modules/webgl/src/features/features.js
1chandu: removed.

modules/webgl/src/features/features.js
1chandu: Done.

modules/webgl/src/features/features.js
tsherif: I don't think this is something that should be used except for unit testing.

PR #1283, opened 2019-10-16
Optimizations

tsherif:
A round of optimizations:
- Uniform setting/diffing: separate setter functions based on whether the uniform will be set as an array or not, so you don't have to do those checks on each update.
- Move logging/debug info in `Program` and `Model` behind debug flags.
- Track whether textures are loading in a property so the full check doesn't have to be done on every draw.
- Avoid allocating in `Model.draw` and `Program.draw`

Some API changes to support this:
- `VertexArray.getDrawParams` no longer takes overrides as an argument. Overriding can be done by the caller.
- After a uniform is set, it's size cannot be changed. Only case where this would occur is if application was setting both an entire uniform array and the first element using the same name.

Results on the stress test:

Metric  |Before|After|Improvement
--------|------|-----|-----------
CPU Time|32ms  |22ms |10ms (31%)
GPU Time|28ms  |24ms |4ms  (14%)


modules/core/src/lib/model.js
```diff
@@ -236,25 +240,34 @@ export default class Model {
     this.updateModuleSettings(moduleSettings);
     this.setUniforms(uniforms);
 
-    const logPriority = this._logDrawCallStart(2);
+    let logPriority;
+
+    if (this.debug) {
```
1chandu: why not use existing `log.priority` variable, and change this if check to `log.priority >= 2` , instead of adding new `this.debug` variable.

modules/webgl/src/classes/vertex-array.js
```diff
@@ -114,7 +114,7 @@ export default class VertexArray {
     this.drawParams = this.drawParams || this._updateDrawParams();
 
     // Override with any application supplied draw parameters
-    return Object.assign({}, this.drawParams, appParameters);
+    return this.drawParams;
```
1chandu: Remove comment, and also remove argument `appParameters` given its not used any more.

modules/webgl/src/classes/program.js
```diff
@@ -185,36 +188,64 @@ export default class Program extends Resource {
     return true;
   }
 
-  setUniforms(uniforms = {}, _onChangeCallback = () => {}) {
-    // Simple change detection - if all uniforms are unchanged, do nothing
-    let somethingChanged = false;
-    const changedUniforms = {};
-    for (const key in uniforms) {
-      if (!areUniformsEqual(this.uniforms[key], uniforms[key])) {
```
1chandu: So this check is turning out to be more expensive than actually setting the uniforms?

tsherif: I've moved the check to inside the setter: https://github.com/uber/luma.gl/pull/1283/files#diff-a257102d168076156bae989f2d59cdb1R296

It is much faster this way.

modules/core/src/lib/model.js
tsherif: It's currently just logging, but I want to move towards all debug logic being separated out from the main render logic. This mirrors what I've done in `Program`.

PR #1288, opened 2019-10-28
Module refactor

tsherif:
Re-arranging modules around "semantic" groupings, and initial changes to documentation (a more complete rewrite of the docs will take place in an upcoming PR).

The following breaks down the new module structure along with each module's purpose, bundle size and components incorporated from the previous structure.

### constants
Purpose: WebGL constants.
- constants

### shadertools
Purpose: Tools for composing/transpiling shaders. Text-only, no WebGL dependencies.
Bundle size: 97kB
- shadertools

### gltools
Purpose: Tools for supporting, optimizing direct coding with the WebGL API.
Bundle size: 36kB
- webgl2-polyfill
- webgl-state-tracker 

### webgl
Purpose: Object-oriented API wrapping WebGL constructs
Bundle size: 146kB
- webgl

### engine
Purpose: High-level drawing APIs
Bundle size: 252kB
- core/animation-loop
- core/model
- core/program-manager
- addons/animation
- core/geometry
- core/transform
- effects

### debug
Purpose: Debug tooling for other modules
Bundle size: 202kB
- debug

### test-utils
Purpose: Utilities to support testing.
Bundle size: 12kB
test-utils

### core
Purpose: Curated re-exports from other modules, maintaining the same API as the current core. Essentially, a one-stop-shop to start drawing without worrying about separate modules, optimizing bundle size.
Bundle size: 299kB

### addons
Purpose: A place for APIs that are unfinished or currently unused. APIs should not be considered officially supported, users use at their own risk. APIs should only be in experimental temporarily. If they spend more than one major release cycle in experimental, they should be strongly considered for removal.
- core/scenegraph
- gpgpu
- addons


modules/engine/package.json
```diff
@@ -30,9 +30,14 @@
...
-    "@luma.gl/constants": "7.4.0-alpha.2"
-  },
-  "peerDependencies": {
-    "@luma.gl/core": "^7.1.0"
+    "@babel/runtime": "^7.0.0",
+    "@luma.gl/constants": "7.4.0-alpha.2",
+    "@luma.gl/shadertools": "7.4.0-alpha.2",
+    "@luma.gl/webgl": "7.4.0-alpha.2",
+    "@luma.gl/webgl-state-tracker": "7.4.0-alpha.2",
+    "@luma.gl/webgl2-polyfill": "7.4.0-alpha.2",
```
Pessimistress: These two no longer exist?

PR #1289, opened 2019-11-05
Fix tables Markdown format

khwj:
There is a missing pipe in Geometry Module doc.

#### Change List
- Fix typo in Geometry Module doc


tsherif:
Thanks for the contribution!

These wip docs have been removed, so we'll close this issue.

PR #1290, opened 2019-11-07
DevicePixelRatio: Handle invalid client sizes and remove custom width/height support.

1chandu:
Related to deck.gl issue 3842

#### Background
- When canvas's `clientWidth` or `clientHeight` is 0 or `undefined` we are falling back to `gl.canvas.width/height`. But `gl.canvas.width/height` may have been scaled with device pixel ratio in previous frame. To avoid this continuous scaling, set `devicePixelRatio` to 1 in such cases.
- Add unit tests.

#### Change List
- DevicePixelRatio: Handle the case when clientWidth/Height not available

### DIFF
--- a/modules/webgl/src/context/context.js
+++ b/modules/webgl/src/context/context.js
@@ -204,19 +204,16 @@ function getVersion(gl) {
 // use devicePixelRatio to set canvas width and height
 function setDevicePixelRatio(gl, devicePixelRatio, options) {
   // NOTE: if options.width and options.height not used remove in v8
-  let clientWidth =
-    'width' in options ? options.width : gl.canvas.clientWidth;
-  let clientHeight =
-    'height' in options ? options.height : gl.canvas.clientHeight;
+  let clientWidth = 'width' in options ? options.width : gl.canvas.clientWidth;
+  let clientHeight = 'height' in options ? options.height : gl.canvas.clientHeight;
 
   if (!clientWidth || !clientHeight) {
-    log.warn('Canvas clientWidht orclientHeight is 0, using devicePixelRatio of 1.0')();
+    log.warn('Canvas clientWidth/clientHeight is 0')();
     // by forcing devicePixel ratio to 1, we do not scale gl.canvas.width and height in each frame.
     devicePixelRatio = 1;
     clientWidth = gl.canvas.width || 1;
     clientHeight = gl.canvas.height || 1;
   }
-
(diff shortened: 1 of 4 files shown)


modules/webgl/src/context/context.js
```diff
@@ -171,15 +171,9 @@ export function resizeGLContext(gl, options = {}) {
   // Resize browser context
   if (gl.canvas) {
     const devicePixelRatio = getDevicePixelRatio(options.useDevicePixels);
-    setDevicePixelRatio(gl, devicePixelRatio, options);
+    setDevicePixelRatio(gl, devicePixelRatio);
     return;
   }
-
-  // Resize headless gl context
```
1chandu: @Pessimistress @ibgreen , any known use cases of code block?

Pessimistress: headless-gl support.

modules/webgl/src/context/context.js
```diff
@@ -202,12 +196,17 @@ function getVersion(gl) {
...
+function setDevicePixelRatio(gl, devicePixelRatio) {
   // NOTE: if options.width and options.height not used remove in v8
-  const clientWidth =
-    'width' in options ? options.width : gl.canvas.clientWidth || gl.canvas.width || 1;
-  const clientHeight =
-    'height' in options ? options.height : gl.canvas.clientHeight || gl.canvas.height || 1;
+  let {clientWidth, clientHeight} = gl.canvas;
+
+  if (!clientWidth || !clientHeight) {
+    log.warn('Canvas clientWidht orclientHeight is 0, using devicePixelRatio of 1.0')();
```
Pessimistress: Remove the warning? It's most likely intentional if the canvas is detached from the document, or at least very obvious to the user.

modules/webgl/src/context/context.js
1chandu: ok, I will keep this for now.

modules/webgl/src/context/context.js
1chandu: This helps isolate the problem, when it occurs,  I can reduce the string to fewer words, if the bundle size is a concern.

Pessimistress: Maybe use `log.log` with a lower priority?

1chandu: Done.

PR #1292, opened 2019-11-12
Fix deps, simplify materials

tsherif:
- Fix missing core dependency on engine
- Export `Timeline` from core
- Allow for default material parameters in phong lighting by passing `material: true`
- Fix case of `gouraudLighting` and `phongLighting` module names

### DIFF
--- a/modules/shadertools/src/modules/phong-lighting/phong-lighting.js
+++ b/modules/shadertools/src/modules/phong-lighting/phong-lighting.js
@@ -21,22 +21,17 @@ const phongLighting = {
   getUniforms
 };
 
-const DEFAULT_MATERIAL_PROPERTIES = {
-  ambient: 0.35,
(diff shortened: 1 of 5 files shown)


modules/shadertools/src/modules/phong-lighting/phong-lighting.js
```diff
@@ -43,7 +50,7 @@ function getUniforms(opts = INITIAL_MODULE_OPTIONS) {
     return {lighting_uEnabled: false};
   }
 
-  return getMaterialUniforms(material);
+  return getMaterialUniforms(Object.assign({}, DEFAULT_MATERIAL_PROPERTIES, material));
```
Pessimistress: It's more efficient to use

```js
const {ambient = 0.35, specularColor = [30, 30, 30], ...} = material;
```

PR #1295, opened 2019-11-22
Remove unpackFlipY option from Texture

Pessimistress:
#### Background

`unpackFlipY` only affects Image and Canvas data sources, and not typed arrays or `ImageBitmap`. This default behavior makes the resulting texture unpredictable when used with `@loaders.gl/images`, which automatically selects the image format at runtime.

#### Change List
- Remove the `unpackFlipY` option from `Texture`
- Upgrade guide

### DIFF
--- a/docs/upgrade-guide.md
+++ b/docs/upgrade-guide.md
@@ -18,6 +18,18 @@ The module structure has been significantly changed for v8.0 with the intention 
 | debug      | Debug tooling for the other modules | Same as before |
 | test-utils | Test tooling for the other modules | Same as before |
 
+### Breaking changes
+
+- `Texture2D`'s `unpackFlipY` option is removed. This change ensures that all data sources (Image, ImageBitmap, typed array) are treated consistently. As a result, textures created from Image objects and URL strings are now y-flipped from the v7.3 default. To get the old behavior, specify the `pixelStore` option:
+
+```js
+new Texture2D({
+  data,
+  pixelStore: {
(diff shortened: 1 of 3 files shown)


xintongxia:
Does this mean the `IconLayer` and `BitmapLayer` need to change to adopt this API?

Pessimistress:
> Does this mean the `IconLayer` and `BitmapLayer` need to change to adopt this API?

Yes. In deck.gl layers, image textures will need to be treated as if they use the top-left coordinate system. It doesn't affect users of these layers because the generation of the textures are internal to the layers.

This flipped y direction is consistent with UV maps you'd generate from a 3D modeling software. The `SimpleMeshLayer` is currently handling uv maps incorrectly, and will be fixed by this change.

tsherif:
Keep the `unpackFlipY` option but make it default to `false`. 

A problem with removing it completely is that pixel storage params are GL state, so if one texture changes them they're changed for all subsequent pixel ops. I notice this is actually a bug in the current implementation as `UNPACK_FLIP_Y_WEBGL` isn't set in `Texture.setData`. We should probably also be setting `UNPACK_PREMULTIPLY_ALPHA_WEBGL` per texture update.

Pessimistress:
> if one texture changes them they're changed for all subsequent pixel ops

`pixelStore` is applied with `withParameters`: https://github.com/uber/luma.gl/blob/master/modules/webgl/src/classes/texture.js#L284

PR #1297, opened 2019-11-26
ClipSpace, getParameters, resolveModules fixes

tsherif:
Fix some breakages caused in deck.gl:
- Restore ClipSpace
- Improve getParameters API so it can be used with a single parameter
- Update resolveModules to report when a string is passed in (deprecated usage)

### DIFF
--- a/modules/gltools/src/state-tracker/unified-parameter-api.js
+++ b/modules/gltools/src/state-tracker/unified-parameter-api.js
@@ -71,6 +71,14 @@ export function getParameters(gl, parameters) {
   // default to querying all parameters
   parameters = parameters || GL_PARAMETER_DEFAULTS;
   // support both arrays of parameters and objects (keys represent parameters)
+
+  if (typeof parameters === 'number') {
+    // single GL enum
+    const key = parameters;
+    const getter = GL_PARAMETER_GETTERS[key];
+    return getter ? getter(gl, Number(key)) : gl.getParameter(Number(key));
+  }
+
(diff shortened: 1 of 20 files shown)


modules/gltools/src/state-tracker/unified-parameter-api.js
```diff
@@ -71,6 +71,14 @@ export function getParameters(gl, parameters) {
   // default to querying all parameters
   parameters = parameters || GL_PARAMETER_DEFAULTS;
   // support both arrays of parameters and objects (keys represent parameters)
+
+  if (typeof parameters === 'number') {
+    // single GL enum
+    const key = parameters;
+    const getter = GL_PARAMETER_GETTERS[key];
+    return getter ? getter(gl, Number(key)) : gl.getParameter(Number(key));
```
Pessimistress: key is already a number

modules/shadertools/src/lib/resolve-modules.js
```diff
@@ -59,14 +59,16 @@ function getDependencyGraph({modules, level, moduleMap, moduleDepth}) {
...
 
-// registers any supplied modules, resolves any names into modules
-// returns a list of modules
 function instantiateModules(modules, seen) {
   return modules.map(module => {
     if (module instanceof ShaderModule) {
       return module;
     }
 
+    assert(
```
Pessimistress: Use `log.removed`

tsherif: Shadertools currently doesn't depend on probe.gl. I'd rather not add a dependency for this.

PR #1299, opened 2019-11-27
Move context creation/query functions to gltools

tsherif:
- Moves the following to gltools:
  - Context creation and query functions: `createGLContext`, `resizeGLContext`, `instrumentGLContext`, `getContextDebugInfo`, `cssToDeviceRatio`, `cssToDevicePixels`, `isWebGL`, `isWebGL2`
  - Environment test variables: `isBrowser`, `isBrowserMainThread`, `isNode`, `isElectron`, `global`, `window`, `self`
- Consolidate gltools utilities 
- Remove global polyfills of `WebGL2RenderingContext` and `Image` (are these used anywhere? luma.gl and deck.gl tests pass without them)
- Remove unreferenced context creation options `throwOnError`


modules/gltools/src/context/context.js
```diff
@@ -34,12 +31,12 @@ export function assertWebGL2Context(gl) {
...
 }
 
-const CONTEST_DEFAULTS = {
+const CONTEXT_DEFAULTS = {
   // COMMON CONTEXT PARAMETERS
   // Attempt to allocate WebGL2 context
   webgl2: true, // Attempt to create a WebGL2 context (false to force webgl1)
   webgl1: true, // Attempt to create a WebGL1 context (false to fail if webgl2 not available)
-  throwOnFailure: true,
+  throwOnError: false,
```
Pessimistress: Is there documentation that needs to be updated?

modules/webgl/src/index.js
```diff
@@ -8,19 +8,22 @@
...
-
-export {createGLContext, resizeGLContext, instrumentGLContext} from './context/context';
+export {
+  resetParameters,
+  getParameters,
+  setParameters,
+  withParameters,
+  isWebGL,
+  isWebGL2
+} from '@luma.gl/gltools';
```
Pessimistress: I think the atom modules should avoid re-exporting APIs from their dependencies. Whoever importing these from the webgl module should switch to importing from gltools, so that we are perfectly clear where things live.

modules/gltools/src/context/context.js
tsherif: Never mentioned anywhere, never used anywhere. Just a ghost prop leftover from a prior refactor I'm guessing.

modules/webgl/src/index.js
tsherif: Sounds good to me.

PR #1303, opened 2019-11-28
Reduce core to modules used by deck.gl

tsherif:
Begin using deck.gl usage as the metric for what should be included in core.


tsherif:
Core bundle size reduction isn't super exciting: went fro 279kB to 272kB

PR #1308, opened 2019-12-06
Shader module tutorial and example

tsherif:


docs/get-started/shader-modules.md
```diff
@@ -0,0 +1,184 @@
...
+```
+
+We have two vertex and fragment shader pairs: one will move vertices to the left, the other moves vertices to the right. Both fragment shaders call a `color_getColor` function to set the final fragment color, but `color_getColor` isn't defined anywhere, so these shaders will not compile as-is.
+
+We define `color_getColor` in a shader module:
+```js
+const colorModule = {
+  name: "color",
+  fs: `
+    vec4 color_getColor() {
```
Pessimistress: Nit: should we show something more useful here, e.g. [HSL to RGB](https://github.com/uber/streetscape.gl/blob/master/modules/core/src/layers/point-cloud-layer/point-cloud-layer-vertex.glsl.js#L46)?

PR #1312, opened 2019-12-10
Shader hook tutorial and example

tsherif:


tsherif:
These changes were merged in as part of #1313

PR #1315, opened 2019-12-11
Transform feedback tutorial and example

tsherif:
Also:
- Removed the old `TransformFeedback` example because it wasn't that interesting and wasn't using the `Transform` class.
- Removed images from the website directory that are no longer used.


docs/getting-started/transform-feedback.md
```diff
@@ -0,0 +1,224 @@
+# Transform Feedback
+
+In this tutorial, we'll learn how to use [transform feedback](https://www.khronos.org/opengl/wiki/Transform_Feedback) using luma.gl's high-level API. Transform feedback allows us to store the results of vertex transforms we do in one draw call and use them in subsequent calls. It is a powerful tool that can be used to set up massively parrallelized animations or data transformations. Note that transform feedback can only be used with WebGL 2.
```
1chandu: Nit: update this text, seems to represent transfrom and draw performed in single draw.
Nit: `WebGL 2` -> `WebGL2`

test/render/index.js
```diff
@@ -20,8 +20,8 @@ test('RenderTest', t => {
 
       imageDiffOptions: {
         // uncomment to save screenshot to disk
-        // saveOnFail: true,
-        // saveAs: '[name].png'
+        saveOnFail: true,
```
1chandu: Nit: remove

docs/getting-started/transform-feedback.md
tsherif: It's WebGL 2 or 2.0 in most official docs:
https://www.khronos.org/registry/webgl/specs/latest/2.0/
https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API#WebGL_2

tsherif: I decided to go the other way and consistently refer to it as "WebGL 2" throughout the docs.

Issue #1339, opened 2020-01-23
Warning when rendering with vertexCount 0


1chandu:
When rendering (non instanced) with vertex count `0`, webgl generates following warning :

`RENDER WARNING: Render count or primcount is 0.`

This is not an issue when using instanced rendering, no warning is generated when `instanceCount` is `0` (probably WebGL is only checking vertex count).

NOTE: This warning is seen in several deck.gl layer tests, that use empty `data` prop.

tsherif:
Fixed by #1340

PR #1340, opened 2020-01-23
[Chore]: Skip draw when when rendering 0 primitives

1chandu:
For #1339 

#### Background
Skip issuing `draw` call when vertex count is 0.

#### Change List
- [Chore]: Skip draw when vertex count is 0

### DIFF
--- a/modules/webgl/src/classes/program.js
+++ b/modules/webgl/src/classes/program.js
@@ -147,10 +147,15 @@ export default class Program extends Resource {
 
     this.gl.useProgram(this.handle);
 
-    // Note: async textures set as uniforms might still be loading.
-    // Now that all uniforms have been updated, check if any texture
-    // in the uniforms is not yet initialized, then we don't draw
-    if (!this._areTexturesRenderable()) {
+    if (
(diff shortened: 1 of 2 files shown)


modules/webgl/src/classes/program.js
```diff
@@ -147,10 +147,14 @@ export default class Program extends Resource {
...
-    // Now that all uniforms have been updated, check if any texture
-    // in the uniforms is not yet initialized, then we don't draw
-    if (!this._areTexturesRenderable()) {
+    if (
+      // Note: async textures set as uniforms might still be loading.
+      // Now that all uniforms have been updated, check if any texture
+      // in the uniforms is not yet initialized, then we don't draw
+      !this._areTexturesRenderable() ||
+      // Avoid WebGL warning when vertexCount is 0
+      vertexCount <= 0
```
tsherif: Make this `=== 0`. A negative value should be considered invalid, and we'd want that error message to appear. Also, might as well add `instanceCount === 0` just to avoid the useless API call.

PR #1346, opened 2020-01-28
Fix broken link

klyap:
For #

#### Background
The current link for Frames navigates to https://github.com/uber/luma.gl/blob/master/docs/api-reference/engine/animation/key-frames, which gives a 404. 

Updating it to https://github.com/uber/luma.gl/blob/master/docs/api-reference/engine/animation/key-frames.md

#### Change List
- Update Frames link in Timeline Readme


klyap:
@tsherif Just signed!

tsherif:
Perfect, thanks!

PR #1355, opened 2020-02-07
Fix model and program logging

### LINKED ISSUE #1341: Fix luma.gl logging
Was causing performance issues, so it was put behind a flag in 8.0. Re-enable in a way that doesn't affect performance.

Affecting deck.gl users: https://github.com/uber/deck.gl/issues/4195

tsherif:
Fixes #1341

### DIFF
--- a/modules/webgl/src/classes/program.js
+++ b/modules/webgl/src/classes/program.js
@@ -54,10 +54,9 @@ export default class Program extends Resource {
   }
 
   initialize(props = {}) {
-    const {hash, vs, fs, varyings, bufferMode = GL_SEPARATE_ATTRIBS, debug = false} = props;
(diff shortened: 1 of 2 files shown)


modules/webgl/src/classes/program.js
```diff
@@ -132,7 +131,7 @@ export default class Program extends Resource {
       this.setUniforms(uniforms || {});
     }
 
-    if (this.debug && logPriority !== undefined) {
+    if (log.priority >= 2 && logPriority !== undefined) {
```
Pessimistress: This should be `log.priority >= logPriority`?

tsherif: Done.

tsherif:
Can do.

Issue #1356, opened 2020-02-10
GPU Accelerated polygon clipping.


1chandu:
#### To Do List 
  - [x] Add label and assign to milestone
  - [x] RFC
  - [x] Proof of concept and Performance
  - [x] Coding
  - [x] Doc update
  - [x] Whats new update 
  - [x] Test

PR #1360, opened 2020-02-12
GPU Accelerated Point In Polygon Test

1chandu:
For #1356 

#### Background
- Add new public class `GPUPointInPolygon` to perform point-in-polygon test.
- Transform: Add support for `TextureTransform` to be created with source as `Buffers` and `Texture` as target. (Shouldn't impact any existing use cases)
- Add polygon triangulation code (`polygon.js`) from deck.gl, we should move this to `math.gl` eventually to avoid duplication.
- Add Docs, Unit Tests and Bench test.

#### Change List
-


modules/engine/src/transform/texture-transform.js
```diff
@@ -47,9 +47,11 @@ export default class TextureTransform {
   }
 
   getDrawOptions(opts = {}) {
-    const {sourceTextures, framebuffer, targetTexture} = this.bindings[this.currentIndex];
+    const {sourceTextures, framebuffer, targetTexture, sourceBuffers} = this.bindings[
```
tsherif: NIT: Put source buffers at the beginning with sourceTextures.

modules/engine/src/transform/texture-transform.js
```diff
@@ -211,14 +213,16 @@ export default class TextureTransform {
...
+    const {sourceTextures, targetTexture, sourceBuffers} = opts;
     if (!binding) {
       binding = {
+        sourceBuffers: {},
         sourceTextures: {},
         targetTexture: null
       };
     }
     Object.assign(binding.sourceTextures, sourceTextures);
+    Object.assign(binding.sourceBuffers, sourceBuffers);
```
tsherif: I'm concerned about all these `Object.assign` allocations. How frequently is `_updateBinding` called?

modules/experimental/src/gpgpu/point-in-polygon/gpu-point-in-polygon.js
```diff
@@ -0,0 +1,183 @@
...
+import {default as textureFilterModule} from './texture-filter';
+import {POLY_TEX_VS, FILTER_VS} from './shaders';
+const TEXTURE_SIZE = 512;
+import * as Polygon from './polygon';
+
+export default class GPUPointInPolygon {
+  constructor(gl, opts = {}) {
+    this.gl = gl;
+    assert(isWebGL2(gl)); // supports WebGL2 only
+    this.textureSize = TEXTURE_SIZE;
```
tsherif: What is this used for?

modules/experimental/src/gpgpu/point-in-polygon/gpu-point-in-polygon.js
```diff
@@ -0,0 +1,183 @@
...
+      texHeight = texWidth / whRatio;
+    } else {
+      texWidth = texHeight * whRatio;
+    }
+
+    this.boundingBox = boundingBox;
+    this.polygonTexture.resize({width: texWidth, height: texHeight, mipmaps: false});
+    this.positionBuffer.setData(new Float32Array(vertices));
+    this.idBuffer.setData(new Float32Array(ids));
+    this.indexBuffer.setData(new Uint16Array(indices));
```
tsherif: Any chance this polygon could get bigger than 65k vertices? Maybe worth putting an assert here?

modules/experimental/src/gpgpu/point-in-polygon/gpu-point-in-polygon.js
```diff
@@ -0,0 +1,183 @@
...
+  const vertices = [];
+  const indices = [];
+  const ids = [];
+  let count = 0;
+  let polygonId = 0;
+  for (let i = 0; i < polygons.length; i++) {
+    const normalized = Polygon.normalize(polygons[i], SIZE);
+    const curVertices = normalized.positions || normalized;
+    const curCount = curVertices.length / SIZE;
+    const curIds = new Float32Array(curCount).fill(polygonId);
```
tsherif: Why is this Float32Array being created? Looks like you just need an array.

modules/experimental/src/gpgpu/point-in-polygon/gpu-point-in-polygon.js
```diff
@@ -0,0 +1,183 @@
...
+    vertices.push(...curVertices);
+    ids.push(...curIds);
```
tsherif: How big can these arrays get? Would recommend pre-allocating the arrays to avoid re-allocating and they dynamically extend.

modules/experimental/src/gpgpu/point-in-polygon/gpu-point-in-polygon.js
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  // PRIVATE
+
+  _setupResources() {
+    const {gl} = this;
+
+    // texture to render polygons to
+    this.polygonTexture = new Texture2D(gl, {
+      format: GL.RGB,
```
tsherif: Why create an RGB texture if you're only writing one value: https://github.com/uber/luma.gl/pull/1360/files#diff-a588d4a9222caddc58312b3ea2d63b3aR13

This can just be GL.RED

modules/experimental/src/gpgpu/point-in-polygon/gpu-point-in-polygon.js
```diff
@@ -0,0 +1,183 @@
...
+      type: GL.UNSIGNED_BYTE,
```
tsherif: So a limit of 255 polygons? Probably fine, but maybe an assert somewhere would be worthwhile.

modules/engine/src/transform/texture-transform.js
1chandu: These object sizes are small, (not more than 5 fields for typical use case), and `assign` is executed during initialization and when a binding (Buffer or Texture) is updated. Most of the times this is done once for multiple draw calls.

modules/experimental/src/gpgpu/point-in-polygon/gpu-point-in-polygon.js
1chandu: Create a texture where polygons are rendered to. Later it is sampled to perform clipping.

modules/experimental/src/gpgpu/point-in-polygon/gpu-point-in-polygon.js
1chandu: `GL.RED` is WebGL2 only, intention of this feature is to support this feature in WebGL environments (which will use Texture Generation and Sampling parts, except TF) by exporting it from luma.gl, we can do that in later iterations.

1chandu: Also we need two channels,  R channel for polygonID [0, 255] and G channel for inside (1) and outside (0). I will update docs to specify support for maximum of 256 polygons. If user not interested in distinguishing polygons by ID, there can be more than 256, while id getting clamped to max value of 255.

modules/experimental/src/gpgpu/point-in-polygon/gpu-point-in-polygon.js
1chandu: Added an assert where these indices are generated.

modules/experimental/src/gpgpu/point-in-polygon/gpu-point-in-polygon.js
1chandu: Updated the docs to clarify this.

modules/experimental/src/gpgpu/point-in-polygon/gpu-point-in-polygon.md
```diff
@@ -0,0 +1,81 @@
...
+  0.25, -0.25,
+  0.25, 8.25,
+  -0.25, 0.25,
+  -3.45, 0.45
+];
+
+const positionBuffer = new Buffer(gl2, new Float32Array(points));
+const count = 6;
+// Allocate result buffer with enough space (2 floats for each point)
+const filterValueIndexBuffer = new Buffer(gl2, count * 2 * 4);
```
heshan0131: is there any reason why I have to create `filterValueIndexBuffer`, why can't it be associated with the `positionBuffer`? and I just call

`positionBuffer.getFilteredData();`

Also the current implementation seems completely independent of Layer. But most of the time people want to apply the polygon filters for an existing point layer. in which case the position Buffer has already existed as `instancePositions`. Would it be possible to have some kind of API to obtain the existing positionBuffer from a layer? Or better create a PolygonFilter layer extension that does this under the hood.

1chandu: The idea to seprerate them, is so users can use existing positionBuffer, if already have it.

This is luma.gl module, `Layer` support will be part of deck.gl feature either through `Effect` or `LayerFilter`.

modules/experimental/src/gpgpu/point-in-polygon/gpu-point-in-polygon.js
1chandu: We can add array manager support in the future PRs, that can be independent of this feature and used across all luma modules.

PR #1379, opened 2020-04-28
Picking module: normalize highlight color

Pessimistress:
#### Background

`setUniforms` crashes if `pickingHighlightColor` is a vec3.

#### Change List
- Normalize `pickingHighlightColor`
- Test

### DIFF
--- a/modules/shadertools/src/modules/picking/picking.js
+++ b/modules/shadertools/src/modules/picking/picking.js
@@ -20,7 +20,11 @@ function getUniforms(opts = DEFAULT_MODULE_OPTIONS) {
     }
   }
   if (opts.pickingHighlightColor) {
-    uniforms.picking_uHighlightColor = opts.pickingHighlightColor.map(x => x / 255);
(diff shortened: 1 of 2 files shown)


modules/shadertools/src/modules/picking/picking.js
```diff
@@ -20,7 +20,11 @@ function getUniforms(opts = DEFAULT_MODULE_OPTIONS) {
     }
   }
   if (opts.pickingHighlightColor) {
-    uniforms.picking_uHighlightColor = opts.pickingHighlightColor.map(x => x / 255);
+    const color = opts.pickingHighlightColor.map(x => x / 255);
+    if (!Number.isFinite(color[3])) {
+      color[3] = 1;
```
tsherif, line 25: This won't work if `color` is a typed array (do we care about that)? An alternative approach:
```js
if (color.length < 4) {
    color = new Float32Array(4).set(color);
    color[3] = 1;
}
```

Pessimistress: Good catch!

PR #1418, opened 2020-11-27
Texture supports continuous update from HTMLVideoElement

Pessimistress:
The Texture class docs claim that it continuously updates if `HTMLVideoElement` is supplied, while in fact it doesn't.

#### Change List
- Prevent `Texture` from initialization until `element.videoWidth` and `element.videoHeight` are loaded
- Add `texture.update` method. It re-uploads the `HTMLVideoElement` if:
  a. the video timestamp has changed
  b. the current frame is loaded
- `Program` updates samplers before every draw call
- Clarify update behavior in documentation

### DIFF
--- a/docs/api-reference/webgl/texture.md
+++ b/docs/api-reference/webgl/texture.md
@@ -176,6 +176,11 @@ Note: does not allocate storage
 See also [gl.compressedTexSubImage2D](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/compressedTexSubImage2D), [gl.texSubImage2D](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/texSubImage2D), [gl.bindTexture](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/bindTexture), [gl.bindBuffer](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/bindBuffer)
 
 
+### update()
+
+Update this texture if `HTMLVideoElement` is used as the data source. This method is automatically called before every draw call if this texture is bound to a uniform.
+
+
 ### getActiveUnit()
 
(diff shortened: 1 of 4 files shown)


ibgreen:
> The Texture class docs claim that it continuously updates if HTMLVideoElement is supplied, while in fact it doesn't.

Unless this is an urgent fix, let's make sure we are on the same page here before landing. 

Video support is built in to WebGL textures - it is not something the framework should need to provide logic for. A WebGL texture that is initialized with an HTMLVideo element will use the latest frame from the video every time it is rendered. There should not be a need for `subImage` calls.

All that should be needed is to make sure that we keep `needsUpdate = true`.

If we do need a bunch of video-specific code, maybe consider organizing it like THREE.js does: https://threejs.org/docs/#api/en/textures/VideoTexture

Pessimistress:
> A WebGL texture that is initialized with an HTMLVideo element will use the latest frame from the video every time it is rendered. 

I don't think so:
https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/Tutorial/Animating_textures_in_WebGL

I'm pretty sure three.js does something similar under the hood.

ibgreen:
This is not speculation. I have used video textures in the past without any manual updates. Unless the browser implementations have changed.

Indeed the THREE.js documentation supports my expectation:

> [VideoTexture] is almost the same as the base Texture class, except that it continuosly sets needsUpdate to true so that the texture is updated as the video plays. Automatic creation of mipmaps is also disabled.

As does the code:

https://github.com/mrdoob/three.js/blob/master/src/textures/VideoTexture.js

Pessimistress:
The code you linked does not show how three.js handles `needsUpdate`.

Here is a break point set on the page https://threejs.org/examples/#webgl_materials_video:

image

ibgreen (approved):
Thanks for investigating.

PR #1449, opened 2021-02-05
Fix shader module injection order

Pessimistress:
For https://github.com/visgl/deck.gl/issues/5436

When normalizing a shader module, `injection.order` is not populated, breaking the injection sorting here:

https://github.com/visgl/luma.gl/blob/784b4def3ba0ad375991481d092cf324e22e0aa3/modules/shadertools/src/lib/assemble-shaders.js#L233

#### Change List
- Assign default injection order
- Tests

### DIFF
--- a/modules/shadertools/src/lib/shader-module.js
+++ b/modules/shadertools/src/lib/shader-module.js
@@ -144,6 +144,7 @@ function normalizeInjections(injections) {
 
     if (typeof injection === 'string') {
       injection = {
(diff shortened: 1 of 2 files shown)


ibgreen (approved):
Good fix. 

I suppose some judicious use of JSDoc types could haven tightened things up here. The typescript conversion effort was too big, had to focus on the surface types, didn't reach this far into the code...

PR #1454, opened 2021-02-20
Fix isWebGL check on non-instrumented context

Pessimistress:
Currently, the following code throws error `Invalid WebGLRenderingContext`:

```js
const gl = canvas.getContext('webgl2');
const texture = new Texture2D(gl, {});
```

Because `isWebGL` does not recognize `WebGL2RenderingContext` that is not instrumented by luma.

#### Change List
- `isWebGL` should recognize `WebGL2RenderingContext`

### DIFF
--- a/modules/gltools/src/utils/webgl-checks.js
+++ b/modules/gltools/src/utils/webgl-checks.js
@@ -8,6 +8,9 @@ export function isWebGL(gl) {
   if (typeof WebGLRenderingContext !== 'undefined' && gl instanceof WebGLRenderingContext) {
     return true;
   }
+  if (typeof WebGL2RenderingContext !== 'undefined' && gl instanceof WebGL2RenderingContext) {
+    return true;
+  }
   // Look for debug contexts, headless gl etc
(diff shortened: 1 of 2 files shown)


ibgreen (approved):
Interesting. `WebGL2RenderingContext` should be an instance of `WebGLRenderingContext`. https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext

Have relied on that it has worked, and never run into this issue before, but maybe it is needed. 

Regardless, it doesn't hurt.

Pessimistress:
Running in Chrome 88.0.4324.182 with luma.gl@8.4.1:

```js
const {isWebGL, isWebGL2} = luma;
const canvas = document.getElementById('canvas');
const gl = canvas.getContext('webgl2');

console.log(gl);
// WebGL2RenderingContext
console.log(isWebGL(gl));
// false
console.log(isWebGL2(gl));
// true
```

https://codepen.io/Pessimistress/pen/MWbmxYE

PR #1469, opened 2021-05-11
Rename constants bundle namespace

Pessimistress:
Constants are currently exported under `luma.default`.

This PR moves them to `luma.GL`.


zakjan:
Unfortunately, apparently this is a breaking change for those who already accessed `luma.default` implicitly. https://github.com/kylebarron/deck.gl-raster/pull/72

PR #1496, opened 2021-09-04
Remove overly broad platform defines

Pessimistress:
iPhone/Apple M1 have been using the custom tan_fp32 function meant for integrated intel GPUs. This PR removes the flag and we can add it back on a more targeted basis if necessary.


ibgreen (approved):

felixpalmer:
When bumping to the latest luma in deck (https://github.com/visgl/deck.gl/pull/6267) I hit some failing render tests (https://github.com/visgl/deck.gl/runs/3812697747) which I've pinned down to the removal of the `#defines` in this PR. I would suggest we remove this PR from the luma.gl package for now and do some more testing. The failures are quite subtle, but seem to be related to the order of the objects being incorrect, see below

old
icon-lnglat-facing-up

new (failing)
icon-lnglat-facing-up-failed

PR #1640, opened 2022-04-22
chore(texture): change compressed textures memory calculation

dryabinin-actionengine:


dryabinin-actionengine:
@ibgreen Could you take a look at browser tests? They starts failing on 8.5-release.

ibgreen (approved):

PR #1746, opened 2023-05-01
Fix cssToDeviceRatio when auto resize is disabled

Pessimistress:
If `autoResizeFramebuffer` is disabled then `luma.canvasSizeInfo` cache is never populated.

#### Change List
- Fallback to `canvas.clientWidth` if cache is not available


ibgreen (approved):
This has changed dramatically in 9.0, this won't be easy to cherry pick.

Issue #1741, opened 2023-04-21
Cannot set a value of false, or 0 for uniform on version 9.0.0-alpha.15


Seungup:
#### Actual Result

```ts
import { Model } from '@luma.gl/engine';
import { picking } from '@luma.gl/shadertools';

const model = new Model( device, { ... , modules: [ picking ]  } );

model.updateModuleSettings( {  pickingActive: 1  } ); // avaliable.
model.updateModuleSettings( {  pickingActive: true } ); // avaliable.
model.updateModuleSettings( {  pickingActive: [1] } ); // avaliable.
model.updateModuleSettings( {  pickingActive: [true]  } ); // avaliable.
model.draw();
model.updateModuleSettings( {  pickingActive: 0  } ); // not avaliable.
model.updateModuleSettings( {  pickingActive: false  } ); // not avaliable.
model.updateModuleSettings( {  pickingActive: [0] } ); // not avaliable.
model.updateModuleSettings( {  pickingActive: [false]  } ); // not avaliable.
model.draw(); // `picking_uActive` still have `true` value.
```

```ts
// picking.ts

...

function getUniforms(opts = DEFAULT_MODULE_OPTIONS): Record<string, any> {
  const uniforms: Record<string, any> = {};
  if (opts.pickingSelectedColor !== undefined) {
    if (!opts.pickingSelectedColor) {
      uniforms.picking_uSelectedColorValid = 0; // -> set 0
    } else {
      const selectedColor = opts.pickingSelectedColor.slice(0, 3);
      uniforms.picking_uSelectedColorValid = 1;
      uniforms.picking_uSelectedColor = selectedColor;
    }
  }
  if (opts.pickingHighlightColor) {
    const color = Array.from(opts.pickingHighlightColor, (x) => x / 255);
    if (!Number.isFinite(color[3])) {
      color[3] = 1;
    }
    uniforms.picking_uHighlightColor = color;
  }
  if (opts.pickingActive !== undefined) {
    uniforms.picking_uActive = Boolean(opts.pickingActive); // -> set false
    uniforms.picking_uAttribute = Boolean(opts.pickingAttribute); // -> set false
  }
  return uniforms;
}

...
```

```ts
// @luma.gl/webgl
// src/adapter/resources/webgl-render-pipline.ts

...

  _applyUniforms() {
    for (const uniformLayout of this.layout.uniforms || []) {
      const {name, location, type, textureUnit} = uniformLayout;
[...]

PR #1766, opened 2023-07-15
fix: 🐛 Fix issue with setting 0 or false values

Seungup:
The change involves replacing the logical OR operator (`||`) with the nullish coalescing operator (`??`) in the `_applyUniforms()` method. This modification addresses a specific issue where 0 or false values were not being correctly assigned. By using the nullish coalescing operator, the code ensures that if the value of `this.uniforms[name]` is either 0 or false, it will be properly assigned as the value for `value`. This change enhances the accuracy and reliability of the code by handling these specific cases effectively.

✅ Closes: #1741

For #1741

#### Background

The issue with the original code was that it did not handle 0 and false values correctly when assigning them to uniforms. This resulted in unexpected behavior when attempting to set these values. The change to use the nullish coalescing operator ensures that 0 and false values are properly assigned.

#### Change List
- Replaced logical OR operator (`||`) with nullish coalescing operator (`??`) in the `_applyUniforms()` method to handle 0 and false values correctly when assigning them to uniforms.

### DIFF
--- a/modules/webgl/src/adapter/resources/webgl-render-pipeline.ts
+++ b/modules/webgl/src/adapter/resources/webgl-render-pipeline.ts
@@ -359,7 +359,7 @@ export class WEBGLRenderPipeline extends RenderPipeline {
   _applyUniforms() {
     for (const uniformLayout of this.layout.uniforms || []) {
       const {name, location, type, textureUnit} = uniformLayout;
-      const value = this.uniforms[name] || textureUnit;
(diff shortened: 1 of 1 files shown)


Seungup:
PS: Uniform object type is `Record<string, any>`

Personally, I think there is a need for the type to be defined as follows.

```ts
export type UniformValue = number | boolean | number[] | boolean[];
```

ibgreen (approved):

PR #1805, opened 2023-08-30
Unify setAttributes and setConstantAttributes

Pessimistress:
#### Change List
- Remove `model.setConstantAttributes`, merge with `model.setAttributes`


ibgreen (commented):
While this should certainly work on WebGL, it pushes a lot of WebGL polyfill logic into the core layer Device API (the RenderPass class), which goes in the opposite direction of months of work aimed at making the Device API as minimal and WebGPU API focused as possible.

My suggestion is that we first explore what exposing VertexArray from Device to Model (engine layer) would look like. I will put up a draft PR.

I will also explore removing of name to location mapping from the core API. If we can make it work on location indices only, and keep the name to location mapping in the engine layer.

PR #1855, opened 2023-11-30
Move WebGL resource tests under modules/core-tests/

donmccurdy:
As best as I can tell, tests under `modules/webgl/test/adapter/resources/*` are duplicates of tests under `modules/core-tests/test/adapter/resources/*`, and we want to centralize them under `core-tests` so we can run the same tests for WebGL and WebGPU later. If I've misunderstood the 'tests' section of https://github.com/visgl/luma.gl/issues/1501 please let me know!

Relocated:

- modules/core-tests/test/adapter/resources/render-pipeline.spec.ts
- modules/core-tests/test/adapter/resources/vertex-array.spec.ts

Removed (duplicate):

- modules/webgl/test/adapter/resources/webgl-buffer.spec.ts
- modules/webgl/test/adapter/resources/webgl-command-buffer.spec.ts
- modules/webgl/test/adapter/resources/webgl-framebuffer.spec.ts
- modules/webgl/test/adapter/resources/webgl-sampler.spec.ts
- modules/webgl/test/adapter/resources/webgl-texture.spec.ts


ibgreen (approved):
I understand that this is just a prep diff. Some observations:
-  The idea was that tests in `core-tests` module should not import webgl or webgpu but just core. If we needed backend specific tests those could remain in respective module. 
- If so a next step could be to uncomment these tests and change the imports

Also. the `Buffer` class has a special issue. 
- In the luma v8 API the Buffer has grown to contain a long list of loosely related convenience method and layout data.
- This doesn't fit with the device model, where resource classes should focus on portable GPU APIs.
- To avoid breaking deck.gl, the webgl module adds all those methods in a `BufferWithAccessor` sublass which it then exports as `Buffer`.
- The buffer tests you removed probably contained some tests for those additional methods.
- The goal is to remove that subclass (as it is not supported in WebGPU) so losing some tests now is not a very big issue, but something to be aware of.

donmccurdy:
I'd missed the non-duplicate tests in modules/webgl/test/adapter/resources/webgl-buffer.spec.ts, thanks! I've restored those tests, and got them passing again. I've also tried to split some of the WebGL-specific tests back into that directory.

I think we're on the same page in terms of structure, and it makes sense about keeping API-agnostic tests in the core-tests folder and API-specific tests elsewhere. I think my main goal for this PR is to make sure we only have a single copy of each test. Agreed that next step would be to uncomment and fix the tests, and if it turns out some tests must be moved to API-specific folders to be fixed, we can do that.

PR #1863, opened 2023-12-05
Remove 'modules', 'test', and 'examples' from .prettierignore

donmccurdy:
These directories were added to `.prettierignore` in #1745, I'm guessing that was not intended to be a long-term change – ok to re-enable them? I don't personally care much about formatting `examples/` at the moment, but being able to format `modules/` and `test/` would save me a bit of time.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -4,7 +4,4 @@ workers/
 wip/
 encoding.js
 encoding-indexes.js
-examples
-modules
-test
 .ocularrc.js


donmccurdy:
Test failure:

```
Forgot to run prettier? There are files without correct code style:
[250+ files]
```

I guess that is why. I'll keep running prettier manually for now then.  😅

donmccurdy:
Appears not to be ESLint, but is instead enforced here:

https://github.com/uber-web/ocular/blob/0939f48187f4cd062d20fc51695c8ed0aba9dc19/modules/dev-tools/scripts/lint.sh#L76-L79

I can't see a way around that now, so instead I think I'll just start incrementally opting-in to prettier for directories I touch in the process of doing other PRs, if that's OK.

PR #1869, opened 2023-12-08
website: device-store to create hidden canvases for device

zbigg:
For #1501

#### Change List
- DeviceTabs creates hidden container for canvases used to create temporary devices
- DeviceTabs is browser only (enabled static build)
- WebGPUDevice to pass height, width and container to WebGPUCanvasContext context


ibgreen (approved):
This looks good. Let's land and align on further refinements.

donmccurdy:
CI appears to be flaky right now because of the babel plugin I re-enabled in #1867, I'll disable that for now with #1870. In the meantime, manually re-running CI usually resolves the failures.

PR #1873, opened 2023-12-11
Add .nycrc.json, exclude .spec.ts files from coverage

donmccurdy:
Currently some `.spec.ts` files are being included when reporting test coverage; this PR removes them from the report.

Should this change be made to upstream ocular-dev-tools instead? Testing locally, I don't see the same problem on the deck.gl repository, but I'm not sure why results would be different.

Total reported coverage may be reduced by this change.


donmccurdy:
Missed that, thanks!

Hm interesting result though, if I just move this 'exclude' line over into the existing config, it doesn't work and the test files are stilled counted in coverage. I'll try to figure out what's going on there...

donmccurdy:
Oh this might be it:

https://github.com/bcoe/c8#cli-options--configuration

> A configuration file can be specified by passing its path on the command line with --config or -c. If no config option is provided, c8 searches for files named .c8rc, .c8rc.json, .nycrc, or .nycrc.json, starting from cwd and walking up the filesystem tree.

We're using c8, which implements nyc's API, but doesn't appear to support a `nyc.config.js` (CommonJS) config. Perhaps we need to rename the config file then.

ibgreen (approved):

ibgreen (approved):

PR #1880, opened 2023-12-14
feat(gltf) Support both size & components in GLTF attributes

felixpalmer:
For https://github.com/visgl/luma.gl/issues/1847

#### Background

The `gltfPrimitive` sometimes has the number of components specified by `size`, and sometimes by `components`, due to postprocessing done in loaders.gl https://github.com/visgl/loaders.gl/blob/master/modules/gltf/src/lib/api/post-process-gltf.ts#L397

It would be nice to tidy this up, for now this PR makes the code support both

#### Change List
- Support `size/components` when creating GLTF geometry


donmccurdy (approved):

ibgreen (approved):
This must be a long standing bug...  I was considering updating the post processing to return a more WebGPU friendly format.

PR #1881, opened 2023-12-14
Restore FP64 arithmetic tests

donmccurdy:
Restores FP64 arithmetic tests, which previously bailed out with `TransformFeedback.isSupported()` hard-coded to `false` during v9 API development. With TransformFeedback working, and a minimum Transform implementation in #1879, we can run the FP64 tests again and they'll pass in CI (Linux, Vulkan, unknown GPU).

Many tests still fail on Apple GPUs, and the `invariant` workaround unfortunately hasn't solved that issue, on my machine at least. For now I've enabled FP64 tests for Apple devices on the cases that do pass.

### DIFF
--- a/modules/engine/src/transform/transform.ts
+++ b/modules/engine/src/transform/transform.ts
@@ -14,12 +14,12 @@ export type TransformProps = {
   vs?: string;
   fs?: string;
   vertexCount?: number;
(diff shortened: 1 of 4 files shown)


donmccurdy:
Merging into the `transform` branch.

PR #1883, opened 2023-12-15
Move 'transform-feedback' example out of WIP, fix errors

donmccurdy:
Fixes the `transform-feedback` example. Some cleanup still needed. Like #1879, I'm relying on an extra buffer instead of reading/writing the same buffer, and I'm not sure how to avoid that yet. The 'swap' implementation is more verbose than in v8, I'll wait for other ideas on the API to fix that.

https://github.com/visgl/luma.gl/assets/1848368/c0447fcb-13ea-4c9e-9681-0ec276a863a1

Related:

- https://github.com/visgl/luma.gl/pull/1879


donmccurdy:
Marking as 'ready for review' just in a general aim to move more code out of WIP into a stable folder, but clearly there are some non-ideal things still in this PR:

- requires separate read/write buffers which could be consolidated
- swap() implementation is verbose
- final Transform API for v9 is still TBD

ibgreen (approved):
Thanks. This used to be part of the tutorials I think, but since TransformFeedback is webgl specific better to keep it in another section, maybe API, maybe `webgl`

This also affects the transform feedback tutorial page in the docs folder.

If you are interested in adding to the to the website, build the website (cd website; yarn; yarn start) and make sure it shows up in the right place and the tutorial pages work. That could be a separate PR of course.

donmccurdy:
I've assumed we intend the `Transform` class to be platform-independent, and (aside from one bit of API leakage) the example doesn't touch `TransformFeedback` directly. Maybe it'd make sense to keep this in `example/api`, but rename it something that isn't WebGL-related, like `transform/` or `buffer-transitions/`?

ibgreen:
> I've assumed we intend the Transform class to be platform-independent

That would be great, but I felt that first we have to create a computeShader variant for WebGPU and see how that looks. Until then it is hard to say how a portable API would look, or if making Transform portable truly makes sense. 

Transform class accept code snippets I think, which might be very different.

Also computeShaders can e.g. do random access on memory through storage buffers which could lead to very different programming patterns from per-row attribute -> attribute processing.

donmccurdy:
Hmmm good points. So for v9.0 I'm looking at splitting BufferTransform / TextureTransform as two (immediately deprecated) classes we'd replace in v9.1. Given that they're temporary, and only nominally WebGL-agnostic... maybe we'll put this in a `webgl` folder but keep it out of the website for now?

ibgreen:
> as two (immediately deprecated) classes we'd replace in v9.1.

Well, we can always "un-deprecate" them in 9.1, should they grow on us. 

It is just one of our mechanisms for signaling that we are not sure. We can also export with leading underscores etc.

ibgreen:
> maybe we'll put this in a webgl folder but keep it out of the website for now?

That is reasonable.

donmccurdy:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/visgl/luma.gl/1883?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* `master`
  * **PR #1904** <a href="https://app.graphite.dev/github/pr/visgl/luma.gl/1904?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #1896** <a href="https://app.graphite.dev/github/pr/visgl/luma.gl/1896?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #1883** <a href="https://app.graphite.dev/github/pr/visgl/luma.gl/1883?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/visgl/luma.gl/1883?utm_source=stack-comment).

PR #1885, opened 2023-12-18
Enable tests for copyTextureToBuffer with Framebuffer input

donmccurdy:
Enables tests for `CommandBuffer#copyTextureToBuffer` where the source is a Framebuffer. I've removed references to the "bufferCreation" option, which doesn't appear to be implemented or perhaps necessary. Is it safe to say the destination buffer can be required when using this function?

Related:

    - "CommandEncoder: encapsulate webgl readPixel()"


ibgreen (approved):
I don't think this method needs to create the buffer, but it could be tricky for the app to know how to allocate a buffer of the correct size. We may eventually want some support for size calculations somewhere...

donmccurdy:
> ...it could be tricky for the app to know how to allocate a buffer of the correct size. We may eventually want some support for size calculations somewhere...

Agreed — though I think if the app doesn't know how to allocate the correct size, it also won't know how to read the result, so one or more utilities might be ideal.

PR #1896, opened 2023-12-20
feat(engine): Transform → BufferTransform + TextureTransform

donmccurdy:
Work in progress. Adding more functionality required for deckgl aggregation layers.

Changes:

- Renames `Transform` → `BufferTransform`
- Adds `TextureTransform` class


donmccurdy:
My intentions here are:

- **BufferTransform:** Is a theoretically API-agnostic wrapper around `Model` which takes the comparable input props, and some additional output-related props, runs the Model, and writes the result of a vertex transformation (via Transform Feedback for now) to a Buffer.
- **TextureTransform:** Is a theoretically API-agnostic wrapper around `Model` which takes comparable input tprops, and some additional output-related props, runs the model, and writes output of the _fragment shader_ to a texture.

In comparison to the current implementation I'm hoping to avoid all of the shader transformation happening in code like [`transform-shader-utils.ts`](https://github.com/visgl/luma.gl/blob/master/wip/modules-wip/webgl-legacy/src/transform/transform-shader-utils.ts), with the expectation that callers of the Transform classes do a bit more work to set up a valid Model and shaders. Simple passthrough fragment shaders can still be autogenerated.

I'm hoping this can considerably simplify the code in preparation for API changes in v9.1. I realize there may be some wrinkles in this plan. I did for example see code in deckgl running transform feedback for transitions and _also_ writing output to a 1-pixel fragment shader to detect when a transition has completed. That can probably be handled here but if there are other complicating factors it would be good to know!


I'd also like to try to ensure that the important functionality of the Transform classes is verified directly with unit tests, which will make future changes much easier. This PR contains a simple 'mean' aggregation, which is not working yet. Probably either I'm not drawing pixels to the right coordinates of the framebuffer, or I haven't bound the framebuffer correctly. Currently after `transform.run()` only the clear color comes back.

donmccurdy:
To do:

- [ ] fix minor test failures
- [ ] add a test with texture as input
- [ ] add a test with 'swap' behavior
- [ ] clean up TODOs

donmccurdy:
This should be ready for review. I think the remaining TODOs in code represent open questions, related to my comment above. Thoughts or preferences would be welcome. In the meantime I'll do some tests locally updating deckgl against this PR, to see if I have further opinions on the TODOs.

This is a 'breaking' change to the previous transform API and is, hopefully:

1. sufficient to cover deckgl's current needs
2. easier to understand and update

donmccurdy:
* `master`
  * **PR #1904** <a href="https://app.graphite.dev/github/pr/visgl/luma.gl/1904?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #1896** <a href="https://app.graphite.dev/github/pr/visgl/luma.gl/1896?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/visgl/luma.gl/1896?utm_source=stack-comment).

modules/engine/src/transform/texture-transform.ts
```diff
@@ -0,0 +1,183 @@
...
+  /** @deprecated */
+  update(...args: any[]): void {
+    // TODO(v9): Method should likely be removed for v9. Keeping a method stub
+    // to assist with migrating DeckGL usage.
+    // eslint-disable-next-line no-console
+    console.warn('TextureTransform#update() not implemented');
+  }
+
+  getData({packed = false} = {}) {
+    throw new Error('getData() not implemented');
```
felixpalmer, line 113: Add a note on why?

modules/engine/src/transform/texture-transform.ts
```diff
@@ -0,0 +1,183 @@
...
+    const {sourceTextures} = this.bindings[index];
+    for (const name in sourceTextures) {
+      sourceTextures[name].sampler = this.sampler;
+    }
+  }
+}
+
+/** Creates an element ID buffer, using float32 for GLSL 1.00 compatibility. */
+function createElementIDBuffer(device: Device, count: number): Buffer {
+  const data = new Float32Array(count);
```
felixpalmer, line 180: Can this not be done with `Uint16Array`?

felixpalmer (approved):

modules/engine/src/transform/texture-transform.ts
donmccurdy: Appears this was for GLSL 1.0. Given other changes to luma v9 and other feedback in this PR, let's use `gl_InstanceID` yes.

PR #1898, opened 2023-12-21
Add ShaderInputs to GLTF example

felixpalmer:
Follow up to https://github.com/visgl/luma.gl/pull/1872

#### Change List
- Fix `hello-cube` example
- Type `color` module in `hello-instancing`
- Use module dependencies when instantiating `ShaderInputs`
- Attempt to use `ShaderInputs` in `hello-gltf`


examples/tutorials/hello-gltf/app.ts
```diff
@@ -58,7 +58,18 @@ export default class AppAnimationLoopTemplate extends AnimationLoopTemplate {
         u_NormalMatrix: new Matrix4(worldMatrix).invert().transpose()
       })
 
-      model.updateModuleSettings({lightSources});
+      model.shaderInputs.setProps({
+        lights: {lightSources}
```
felixpalmer, line 62: The source of the issue is likely the use of the `lights` module. Instead we should use the `lighting` module, by switching out the dependency in the `pbr` shader module

PR #1906, opened 2024-01-03
fix(shadertools): Correct order of uniforms in picking module

felixpalmer:
#### Background

The order of uniforms in the `uniformTypes` must match that used in the shader. This was not the case in the picking module, which led to the incorrect layout being generated

Screenshot 2024-01-03 at 14 27 51

with the positions of `useFloatColors` and `isHighlightActive` being swapped

Screenshot 2024-01-03 at 15 11 31

#### Change List
- Correct order of uniforms in picking module

### DIFF
--- a/modules/shadertools/src/modules/engine/picking/picking.ts
+++ b/modules/shadertools/src/modules/engine/picking/picking.ts
@@ -210,16 +210,16 @@ export const picking: ShaderModule<PickingProps, PickingUniforms> = {
   uniformTypes: {
     isActive: 'f32',
     isAttribute: 'f32',
-    useFloatColors: 'f32',
     isHighlightActive: 'f32',
+    useFloatColors: 'f32',
     highlightedObjectColor: 'vec3<f32>',
     highlightColor: 'vec4<f32>'
   },
   defaultUniforms: {
     isActive: false,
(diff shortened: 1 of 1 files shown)


donmccurdy (approved):
> The order of uniforms in the uniformTypes must match that used in the shader.

That sounds painful to discover! Do you know whether this is an intended constraint of ShaderInputs, or an unexpected one? 

I wonder if we should include a (temporary?) lint before shader compilation to detect and warn about mismatched order. If it's an unintended or new constraint, there may be more cases like this.

felixpalmer:
> Do you know whether this is an intended constraint of ShaderInputs, or an unexpected one?

Not sure, @ibgreen?

> I wonder if we should include a (temporary?) lint before shader compilation to detect and warn about mismatched order. If it's an unintended or new constraint, there may be more cases like this.

That would be nice, but not really sure how to go about it, and my current feeling is that we have a lot on our plate to land the picking UBO plate as is. If you want to take a stab though, that would be helpful :)

donmccurdy:
Ah yeah! Just floating the idea, no expectation or ask on that. 😁 I may take a shot at it if I run into similar issues.

PR #1907, opened 2024-01-04
feat(engine) Partial prop updates in ShaderInputs.setProps

felixpalmer:
#### Background

deck.gl doesn't update all of the props for picking in one place, for example the `highlightColor` is set once when the layer is created. In contrast the `isActive` is changed very frequently, depending on whether a screen or picking pass is being rendered. In order to save the application (deck) from having to keep track of the state this change:

- Modifies `picking.getUniforms` to only return uniforms which are relevant to the passed in props
- Removes all knowledge of `picking.defaultUniforms` from `picking.getUniforms`
- Moves the responsibility for reading `defaultUniforms` into the `ShaderInputs` constructor
- Merges old uniforms with new uniforms in `ShaderInputs`

This design allows the `ShaderModule` to remain stateless, and only specify how props are translated into uniforms, while `ShaderInputs` holds the state.

### DIFF
--- a/modules/engine/src/shader-inputs.ts
+++ b/modules/engine/src/shader-inputs.ts
@@ -89,7 +89,7 @@ export class ShaderInputs<
       const moduleProps = props[moduleName];
       const module = this.modules[moduleName];
 
-      const oldUniforms = this.moduleUniforms[moduleName] || {};  
+      const oldUniforms = this.moduleUniforms[moduleName];  
       const uniforms = module.getUniforms?.(moduleProps, this.moduleUniforms[moduleName]) || moduleProps as any;
       // console.error(uniforms)
       this.moduleUniforms[moduleName] = {...oldUniforms, ...uniforms};
--- a/modules/engine/src/shader-inputs.ts
+++ b/modules/engine/src/shader-inputs.ts
@@ -72,7 +72,7 @@ export class ShaderInputs<
       const moduleName = name as keyof ShaderPropsT;
 
       // Get default uniforms from module
-      this.moduleUniforms[moduleName] = module.getUniforms?.({}) || module.defaultUniforms || {};
+      this.moduleUniforms[moduleName] = module.defaultUniforms || {};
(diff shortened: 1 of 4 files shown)


modules/engine/src/shader-inputs.ts
```diff
@@ -89,9 +89,10 @@ export class ShaderInputs<
       const moduleProps = props[moduleName];
       const module = this.modules[moduleName];
 
-      const uniforms = module.getUniforms?.(moduleProps, this.moduleUniforms[moduleName]);
+      const oldUniforms = this.moduleUniforms[moduleName] || {};  
+      const uniforms = module.getUniforms?.(moduleProps, this.moduleUniforms[moduleName]) || moduleProps as any;
       // console.error(uniforms)
-      this.moduleUniforms[moduleName] = uniforms || moduleProps as any;
+      this.moduleUniforms[moduleName] = {...oldUniforms, ...uniforms};
```
donmccurdy, line 95: There's no nesting within a given module's uniforms, correct?

modules/engine/src/shader-inputs.ts
```diff
@@ -89,9 +89,10 @@ export class ShaderInputs<
       const moduleProps = props[moduleName];
       const module = this.modules[moduleName];
 
-      const uniforms = module.getUniforms?.(moduleProps, this.moduleUniforms[moduleName]);
+      const oldUniforms = this.moduleUniforms[moduleName] || {};
```
donmccurdy, line 92: nit, optional: `this.moduleUniforms[moduleName]` should be non-nullable based on the constructor and types. So either the fallback to `{}` should not be needed, or the types are incorrectly strict?

donmccurdy (approved):
I think the updates clarify the code as well, looks good to me!

modules/engine/src/shader-inputs.ts
felixpalmer: Correct, the type is `PropsT extends Record<string, unknown> `

PR #1915, opened 2024-01-12
feat(shadertools): Export PickingProps

felixpalmer:
#### Change List
- Export PickingSettings


ibgreen (approved):
Nits: 

I was moving to `...Props` naming for module types, but I suspect this is still in flux across all the shader modules so we can still decide otherwise.

But I do not see a reason for us to define both Settings and Props. 

Also I am not a fan of the `Omit<...>` construction that is currently used to define the `PickingSettings` type, especially for the primary public type because it makes it harder for users to inspect that type (whether hovering in vscode or navigating to the definition).

If we do want to keep `Omit<>` to save some typing, it it would make more sense to me that the now largely internal type `PickingUniforms` type was defined this way. 

I suppose I think of them as separate types that are mapped explicitly here, not as minor variations of each other.

PR #1921, opened 2024-01-17
bug(engine) Handle null/undefined in attribute debug

felixpalmer:
#### Background

I found `attribute.toString()` throwing an error in deck

#### Change List
- Handle null/undefined in attribute debug

### DIFF
--- a/modules/engine/src/model/model.ts
+++ b/modules/engine/src/model/model.ts
@@ -612,8 +612,10 @@ export class Model {
   // TODO - fix typing of luma data types
   _getBufferOrConstantValues(attribute: Buffer | TypedArray, dataType: any): string {
     if (attribute === null) {
+      log.warn('Missing attribute')();
       return 'null';
     } else if (attribute === undefined) {
+      log.warn('Missing attribute')();
       return 'undefined';
     }
     const TypedArrayConstructor = getTypedArrayFromDataType(dataType);
--- a/modules/engine/src/model/model.ts
+++ b/modules/engine/src/model/model.ts
@@ -611,6 +611,11 @@ export class Model {
 
   // TODO - fix typing of luma data types
   _getBufferOrConstantValues(attribute: Buffer | TypedArray, dataType: any): string {
+    if (attribute === null) {
+      return 'null';
+    } else if (attribute === undefined) {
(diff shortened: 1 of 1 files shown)


ibgreen (approved):
While the guard is harmless, this is inconsistent with the typing and should never happen. Somewhere we have a "type hole", either in luma.gl or deck.gl. Some code is untyped or overrides the type system with `!` or ``// @ts-ignore/@ts-expect-error/@ts-nocheck`. We should chase it down!

Fine to merge but perhaps a comment explaining why we check for something that is not supposed to happen.

As an alternate approach we could also throw an error or at least log since this is supposed to be invalid?

ibgreen:
> on the other hand should the debug logging ever cause the app to crash?

Well for a any single case that argument is not hard to make, but where do we draw the line? How much overhead are we willing to add to our libraries to catch cases where the application doesn't respect the type system. In the pre-typescript version we had significantly more null checks but these have intentionally been removed.

The debug logging should not be active unless the log level has been set higher than 0. Maybe adding an early exit making sure that debug logging code doesn't execute in that case is the right approach? There really should not be any overhead incurred by the debug code for non-production applications.

felixpalmer:
Fair point. I'm coming around to the view that perhaps the app crashing here is correct. If you agree @ibgreen I will revert this PR

ibgreen:
>  I'm coming around to the view that perhaps the app crashing here is correct.

Yes, though one thing that confuses me is that I would expect it to crash somewhere else as well, or at least that warnings would also be generated somewhere else. If this silently works and only crashes here we might want to investigate a bit more to understand what is going on.

felixpalmer:
It only "silently works" because the calling code is untyped, when the type is added the error shows up as expected

Screenshot 2024-01-18 at 10 48 04

ibgreen:
> It only "silently works" because the calling code is untyped

Yes, that explains the lack of typing errors, but not why the presence of null attributes doesn't cause a crash. Does it actually render correctly? Or does it fail to render and just doesn't crash?

I.e. are null attributes supported in the rest of the code, in a way that we want to preserve? If so adding corresponding checks to the logging was perhaps the right decision?

PR #1950, opened 2024-02-21
feat(engine): Add ShaderFactory utility and caching

donmccurdy:
Adds a `ShaderFactory` utility, similar to PipelineFactory, for caching and reusing Shader instances. Upstream in DeckGL, each tile currently creates a model, and each model creates a RenderPipeline and two Shader instances. With these changes, the models reuse cached RenderPipeline and Shader instances where possible.

The before/after images below show performance profiles running while animating the map from point A to point B.

### Before

Screenshot 2024-02-20 at 3 55 52 PM

### After

after

### Related

- Fixes https://github.com/visgl/deck.gl/issues/8458#issuecomment-1955087738

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -16,6 +16,7 @@ test
 !modules/core/src/adapter/types/shader-layout.ts
 !modules/core/src/index.ts
 !modules/engine/src/model/model.ts
+!modules/engine/src/lib/shader-factory.ts
 !modules/webgl/src/adapter/resources/webgl-buffer.ts
 !modules/webgl/src/adapter/resources/webgl-render-pipeline.ts
 !modules/webgl/src/adapter/resources/webgl-transform-feedback.ts
--- a/modules/engine/src/lib/shader-factory.ts
+++ b/modules/engine/src/lib/shader-factory.ts
@@ -1,0 +1,52 @@
+import {Device, Shader, ShaderProps} from '@luma.gl/core';
+
+/** Manages a cached pool of Shaders for reuse. */
+export class ShaderFactory {
+  static readonly defaultProps: Required<ShaderProps> = {...Shader.defaultProps};
+
+  public readonly device: Device;
+
+  private readonly _shaderCache: Record<string, Shader> = {};
+  private readonly _useCounts: Record<string, number> = {};
+
+  /** Returns the default ShaderFactory for the given {@link Device}, creating one if necessary. */
+  static getDefaultShaderFactory(device: Device): ShaderFactory {
+    device._lumaData.defaultShaderFactory =
+      device._lumaData.defaultShaderFactory || new ShaderFactory(device);
+    return device._lumaData.defaultShaderFactory as ShaderFactory;
+  }
+
+  /** @internal */
+  constructor(device: Device) {
+    this.device = device;
+  }
+
+  /** Requests a {@link Shader} from the cache, creating a new Shader only if necessary. */
+  createShader(props: ShaderProps): Shader {
+    const key = this._hashShader(props);
+    if (!this._shaderCache[key]) {
+      this._shaderCache[key] = this.device.createShader(props);
+      this._useCounts[key] = 0;
+    }
+    this._useCounts[key]++;
+    return this._shaderCache[key];
+  }
+
+  /** Releases a previously-requested {@link Shader}, destroying it if no users remain. */
+  release(shader: Shader): void {
+    const key = this._hashShader(shader);
+    this._useCounts[key]--;
+    if (this._useCounts[key] === 0) {
+      const shader = this._shaderCache[key];
+      delete this._shaderCache[key];
+      delete this._useCounts[key];
+      shader.destroy();
+    }
+  }
+
+  // PRIVATE
+
+  private _hashShader(value: Shader | ShaderProps): string {
+    return `${value.stage}:${value.source}`;
+  }
+}
--- a/modules/engine/src/lib/shader-factory.ts
+++ b/modules/engine/src/lib/shader-factory.ts
@@ -11,8 +11,7 @@ export class ShaderFactory {
 
   /** Returns the default ShaderFactory for the given {@link Device}, creating one if necessary. */
   static getDefaultShaderFactory(device: Device): ShaderFactory {
-    device._lumaData.defaultShaderFactory =
-      device._lumaData.defaultShaderFactory || new ShaderFactory(device);
+    device._lumaData.defaultShaderFactory ||= new ShaderFactory(device);
     return device._lumaData.defaultShaderFactory as ShaderFactory;
   }
 
--- a/modules/engine/src/lib/shader-factory.ts
+++ b/modules/engine/src/lib/shader-factory.ts
@@ -25,7 +25,11 @@ export class ShaderFactory {
 
     let cacheEntry = this._cache[key];
     if (!cacheEntry) {
-      this._cache[key] = cacheEntry = {shader: this.device.createShader(props), useCount: 0};
+      const shader = this.device.createShader({
+        ...props,
+        id: props.id ? `${props.id}-cached` : undefined
+      });
+      this._cache[key] = cacheEntry = {shader, useCount: 0};
     }
 
     cacheEntry.useCount++;
--- a/modules/engine/src/lib/shader-factory.ts
+++ b/modules/engine/src/lib/shader-factory.ts
@@ -6,8 +6,7 @@ export class ShaderFactory {
 
   public readonly device: Device;
 
-  private readonly _shaderCache: Record<string, Shader> = {};
-  private readonly _useCounts: Record<string, number> = {};
+  private readonly _cache: Record<string, {shader: Shader; useCount: number}> = {};
 
   /** Returns the default ShaderFactory for the given {@link Device}, creating one if necessary. */
   static getDefaultShaderFactory(device: Device): ShaderFactory {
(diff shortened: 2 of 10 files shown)


donmccurdy:
* **#1950** <a href="https://app.graphite.dev/github/pr/visgl/luma.gl/1950?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/visgl/luma.gl/1950?utm_source=stack-comment).

ibgreen (commented):
I suppose a shader cache should be a reasonable thing to add - with a small potential downside be that it might end up holding on to potentially a lot of unused shaders,

But I am a little surprised that it is needed, i.e that the `RenderPipelines` are not being reused between tiles via the RenderPipelineFactory in the first place. Do we understand why that is the case? If Pipelines were reused, then shaders should automatically be reused?

Are we including something unnecessary in the renderpipelineFactory hash? 

---

Another observation: the traces seem to show that it is the querying of the compilation/linking status that is the bottleneck (somewhat unsurprisingly, as any response from the driver syncs the GPU or otherwise slows things down). 

While caching helps for repeated creation of the same shader, the status querying could still penalize apps with lots of different shaders, we could also add a way to avoid checking shader compilation status in production apps.

Note that in WebGPU the querying of compile status is async so should not suffer from this problem. AFAIK there is a WebGL extension to query compile status async but I believe it is not implemented on MacOS so we never tried to integrate it.

donmccurdy:
> Are we including something unnecessary in the renderpipelineFactory hash?

That sounds possible. I quickly tried replacing `device.createRenderPipeline` with `pipelineFactory.createRenderPipeline` in `model.ts`, and that alone wasn't enough to resolve the issue, but I haven't dug deeper into that part yet. I think I can take a closer look at this tomorrow.

donmccurdy:
@ibgreen this seems to be the root of the issue —

https://github.com/visgl/luma.gl/blob/34bafc3024cb159d47df5926ed1e091573334b58/modules/engine/src/model/model.ts#L577-L598

Shaders are created and compiled unconditionally, then passed into `device.createRenderPipeline`. We can replace that with `pipelineFactory.createRenderPipeline` and it will cache the pipeline (correctly as far as I can see), but pipeline caching does not prevent the unused shaders from compiling.

donmccurdy:
@ibgreen sounds good! I've split this PR in two, opening:

- https://github.com/visgl/luma.gl/pull/1953

donmccurdy:
Continued performance profiling —

With two recent PRs for PipelineFactory and async shader compilation:

- https://github.com/visgl/luma.gl/pull/1953
- https://github.com/visgl/luma.gl/pull/1952

| flamechart | detail |
|---|---|
| total_async | detail_async |

With ShaderFactory:

| flamechart | detail |
|---|----|
| total_shaderfactory | detail_shaderfactory |

Environment:

```
System:
    OS: macOS 14.3.1
    CPU: (10) arm64 Apple M2 Pro
    Memory: 74.92 MB / 16.00 GB
```

Pay attention the X axis — my screenshots are not to scale sorry. :)

***

Async compilation improves the situation dramatically. That said, we're still compiling shaders for new tiles and throwing them away, and this costs up to 25-30ms on my (fairly modern) laptop. I believe Model should not do that, but whether introducing ShaderFactory is the right solution, I am flexible on. Some TODOs (removed in https://github.com/visgl/luma.gl/pull/1953) suggested that we wanted PipelineFactory to take Shaders as input, like the Pipeline classes do. But if we want to revert that part of the change and take raw shader strings as input, that could be another way of working around this.

.prettierignore
```diff
@@ -16,6 +16,7 @@ test
 !modules/core/src/adapter/types/shader-layout.ts
 !modules/core/src/index.ts
 !modules/engine/src/model/model.ts
+!modules/engine/src/lib/shader-factory.ts
```
ibgreen, line 19: I never took the time to figure out what is happening here, why is prettier not stable for some files?

modules/engine/src/lib/shader-factory.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  public readonly device: Device;
+
+  private readonly _shaderCache: Record<string, Shader> = {};
+  private readonly _useCounts: Record<string, number> = {};
+
+  /** Returns the default ShaderFactory for the given {@link Device}, creating one if necessary. */
+  static getDefaultShaderFactory(device: Device): ShaderFactory {
+    device._lumaData.defaultShaderFactory =
+      device._lumaData.defaultShaderFactory || new ShaderFactory(device);
```
ibgreen, line 15: I think this works?
```suggestion
    device._lumaData.defaultShaderFactory ||= new ShaderFactory(device);
```

modules/engine/src/model/model.ts
```diff
@@ -221,6 +231,7 @@ export class Model {
 
     this.pipelineFactory =
       props.pipelineFactory || PipelineFactory.getDefaultPipelineFactory(this.device);
+    this.shaderFactory = props.shaderFactory || ShaderFactory.getDefaultShaderFactory(this.device);
```
ibgreen, line 234:
```suggestion
    this.shaderFactory ||= ShaderFactory.getDefaultShaderFactory(this.device);
```

modules/engine/src/lib/shader-factory.ts
```diff
@@ -0,0 +1,52 @@
+import {Device, Shader, ShaderProps} from '@luma.gl/core';
+
+/** Manages a cached pool of Shaders for reuse. */
+export class ShaderFactory {
+  static readonly defaultProps: Required<ShaderProps> = {...Shader.defaultProps};
+
+  public readonly device: Device;
+
+  private readonly _shaderCache: Record<string, Shader> = {};
```
ibgreen, line 9: This part always gave me pause in the PipelineFactory... I wonder if a single record per key isn't more readable? (I hope you are not just following my bad examples :)

```suggestion
  private readonly _cache: Record<string, {shader: Shader, useCount: number}> = {};
```

modules/engine/src/model/model.ts
```diff
@@ -277,6 +288,8 @@ export class Model {
   }
 
   destroy(): void {
+    this.shaderFactory.release(this.pipeline.vs);
+    this.shaderFactory.release(this.pipeline.fs);
     this.pipelineFactory.release(this.pipeline);
```
ibgreen, line 293: Somehow releasing the pipeline first seems more safe, since it might still be referencing those destroyed shaders?

modules/engine/src/model/model.ts
```diff
@@ -565,23 +578,25 @@ export class Model {
 
   _updatePipeline(): RenderPipeline {
     if (this._pipelineNeedsUpdate) {
+      const prevShaders = [] as Shader[];
```
ibgreen, line 581:
```suggestion
      const prevShaders: Shader[] = [];
```

modules/engine/src/model/model.ts
```diff
@@ -565,23 +578,25 @@ export class Model {
...
       if (this.pipeline) {
         log.log(
           1,
           `Model ${this.id}: Recreating pipeline because "${this._pipelineNeedsUpdate}".`
         )();
+        prevShaders.push(this.pipeline.vs, this.pipeline.fs);
```
ibgreen, line 587: fs can be null in transform feedback, do we need to do `prevShaders.filter(Boolean)`?

modules/engine/src/model/model.ts
```diff
@@ -565,23 +578,25 @@ export class Model {
...
       }
 
       this._pipelineNeedsUpdate = false;
 
-      const vs = this.device.createShader({
+      const vs = this.shaderFactory.createShader({
         id: `${this.id}-vertex`,
```
ibgreen, line 593: How do we handle the id in the shader cache hash? 
- If we ignore it, the first id would be frozen, so we might as well overwrite it or decorate it somehow (`-cached`?) to make clear that it might not be correct.
- If we include it, we could create multiple identical shaders with just different ids, but ids would be correct.
I think this behavior should be documented.

modules/engine/src/model/model.ts
```diff
@@ -601,6 +616,10 @@ export class Model {
         this.pipeline.shaderLayout,
         this.bufferLayout
       );
+
+      for (const shader of prevShaders) {
+        this.shaderFactory.release(shader);
+      }
```
ibgreen, line 622: Wouldn't just having two temp variables would be more readable than using an array? Sometimes simple is best...
```suggestion
      for (const shader of prevShaders) {
        this.shaderFactory.release(previousVs);
        this.shaderFactory.release(previousFs);
      }
```

ibgreen (approved):
- tests
- docs
- whats-new.md
- Makes me think that a performance page, perhaps next to the profiling page would be nice https://github.com/visgl/luma.gl/blob/master/docs/developer-guide/profiling.md

.prettierignore
donmccurdy: My intention here is to tell Prettier that it *can* format this file, as an exception to all of `modules` being excluded from Prettier formatting above. Since it's a new file we might as well keep it formatted. 

However, my VSCode-Prettier plugin doesn't seem to understand these `!` negations, despite documentation saying it should work, so it's possible this line is doing nothing at all...

modules/engine/src/model/model.ts
donmccurdy: Ah, makes sense yes!

modules/engine/src/model/model.ts
donmccurdy: Done!

modules/engine/src/model/model.ts
donmccurdy: The `-cached` suffix feels reasonable to me – I've gone with that for now, and will include it in the docs.

PR #1978, opened 2024-03-01
chore(build): size optimization

Pessimistress:
#### Change List
- Bump ocular-dev-tools
- Use `inline-webgl-constants` transformer
- Use `remove-glsl-comments` transformer

### DIFF
--- a/modules/constants/package.json
+++ b/modules/constants/package.json
@@ -35,6 +35,7 @@
   ],
   "sideEffects": false,
   "scripts": {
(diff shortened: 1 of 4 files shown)


modules/constants/package.json
```diff
@@ -35,6 +35,7 @@
   ],
   "sideEffects": false,
   "scripts": {
+    "pre-build": "npx esbuild src/index.ts --bundle --format=cjs --outfile=dist/index.cjs",
```
Pessimistress, line 38: The constants module is a dependency of the TS plugin

PR #2010, opened 2024-03-07
Move core-tests into core/tests

Pessimistress:
#### Change List
- Remove core-tests "module"


ibgreen (approved):
I am a bit suspicious as I had big problems making this setup work with tsconfigs etc and I don't really see any key changes that would address that class of problems here. But a quick test on your branch looked good, so let's give a try.

PR #2032, opened 2024-03-13
fix(webgl): Pair calls to bind/unbind VAO

donmccurdy:
Fixes:

- https://felixpalmer.github.io/deck.gl/examples/tile-layer
- https://felixpalmer.github.io/deck.gl/examples/tile-layer-non-geospatial

It's not obvious to me that we should be able to bind the index buffer only once... if not, we need to bind on every frame.

Related:

- https://github.com/visgl/deck.gl/issues/8582

### DIFF
--- a/modules/webgl/src/adapter/resources/webgl-vertex-array.ts
+++ b/modules/webgl/src/adapter/resources/webgl-vertex-array.ts
@@ -57,8 +57,8 @@ export class WEBGLVertexArray extends VertexArray {
   /**
   // Set (bind/unbind) an elements buffer, for indexed rendering.
   // Must be a Buffer bound to GL.ELEMENT_ARRAY_BUFFER or null. Constants not supported
-   * 
-   * @param elementBuffer 
+   *
+   * @param elementBuffer
    */
   setIndexBuffer(indexBuffer: Buffer | null): void {
     const buffer = indexBuffer as WEBGLBuffer;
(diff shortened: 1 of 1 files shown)


donmccurdy:
* **#2032** <a href="https://app.graphite.dev/github/pr/visgl/luma.gl/2032?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

ibgreen (approved):
If the index buffer is not stored in the underlying VertexArray, then one could question whether it should be managed by the VertexArray class.

donmccurdy:
On closer look at the spec, I think I see the issue – will try to bump this to avoid rebinding on each frame.

modules/webgl/src/adapter/resources/webgl-vertex-array.ts
```diff
@@ -97,16 +97,16 @@ export class WEBGLVertexArray extends VertexArray {
       // Attaches ARRAY_BUFFER with specified buffer format to location
       this.device.gl.vertexAttribPointer(location, size, type, normalized, stride, offset);
     }
-    // Clear binding - keeping it may cause [.WebGL-0x12804417100]
-    // GL_INVALID_OPERATION: A transform feedback buffer that would be written to is also bound to a non-transform-feedback target
-    this.device.gl.bindBuffer(GL.ARRAY_BUFFER, null);
```
donmccurdy, line 102: @ibgreen probably this had the same root cause, and we can remove it now?  Unit tests are passing and I double-checked the attribute-transitions demo in deckgl against the change, let me know if there's anything else you know to test for this issue.

donmccurdy:
Done! I think the trick here will be to pair every call to `bindVertexArray(vao)` with a later call to `bindVertexArray(null)`, to make sure nothing modifies our VAO unintentionally. With that, we can do the index buffer binding once up front, as you'd intended.

modules/webgl/src/adapter/resources/webgl-vertex-array.ts
ibgreen: Wow - was it just a stupid bug all along?

ibgreen (approved):
Ok Let's land it and keep an eye on it. It will be great if this binding/unbinding can be remove.

modules/webgl/src/adapter/resources/webgl-vertex-array.ts
Pessimistress: The invalid operation error is back with the removal of this :(

PR #2037, opened 2024-03-14
feat(webgl): Add device.resetWebGL() for debugging

donmccurdy:
When debugging issues with interleaved layers ([example](https://github.com/visgl/deck.gl/issues/8602)) it can be helpful to have some way of resetting WebGL state, to isolate whether state is leaking between applications. I realize this method is one we'd probably want to avoid in production code, but perhaps with providing on the device API during v9.0 and v9.1 development? Thanks!

### DIFF
--- a/modules/core/src/adapter/device.ts
+++ b/modules/core/src/adapter/device.ts
@@ -479,7 +479,7 @@ export abstract class Device {
     throw new Error('not implemented');
   }
 
-  /** @deprecated */
+  /** @deprecated - will be removed - should use for debugging only */
   resetWebGL(): void {
     throw new Error('not implemented');
   }
--- a/modules/core/src/adapter/device.ts
+++ b/modules/core/src/adapter/device.ts
@@ -479,6 +479,11 @@ export abstract class Device {
     throw new Error('not implemented');
   }
 
+  /** @deprecated */
+  resetWebGL(): void {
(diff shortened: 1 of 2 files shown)


donmccurdy:
* **#2037** <a href="https://app.graphite.dev/github/pr/visgl/luma.gl/2037?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

modules/core/src/adapter/device.ts
```diff
@@ -479,6 +479,11 @@ export abstract class Device {
     throw new Error('not implemented');
   }
 
+  /** @deprecated */
```
felixpalmer, line 482: Add a note saying it is just for debugging?

modules/webgl/src/adapter/webgl-device.ts
```diff
@@ -412,6 +416,10 @@ ${device.info.vendor}, ${device.info.renderer} for canvas: ${device.canvasContex
     clear(this, options);
   }
 
+  override resetWebGL(): void {
+    resetGLParameters(this.gl);
```
felixpalmer, line 420: `log.warn` to briefly say it is for debug only, to avoid it creeping into production apps?

felixpalmer (approved):

PR #2040, opened 2024-03-15
feat(core): Add missing blend factors, restrict types in device-parameters

### LINKED ISSUE #2039: `convertBlendFactorToFunction()` doesn't support all constants from `BlendFactor`
#### Reproduce Steps

`convertBlendFactorToFunction('blendColorSrcFactor', 'blend-color')` is not going to work as the [mapping](https://github.com/visgl/luma.gl/blob/master/modules/webgl/src/adapter/converters/device-parameters.ts#L379-L390) doesn't include the following keys [present](https://github.com/visgl/luma.gl/blob/master/modules/core/src/adapter/types/parameters.ts#L132-L134) in `BlendFactor`:

- `'src-alpha-saturated'`
- `'blend-color'`
- `'one-minus-blend-color'`

#### To Do List 
  - [ ] Coding
  - [ ] Test

#### To Do List 
  - [ ] Coding
  - [ ] Doc update
  - [ ] Whats new update 
  - [ ] Test

donmccurdy:
Changes:

- Add missing blend factors
- Tighten types in device-parameters to catch similar issues
- Fixes #2039

### DIFF
--- a/modules/constants/src/webgl-types.ts
+++ b/modules/constants/src/webgl-types.ts
@@ -368,7 +368,7 @@ export type GLFunctionParameters = {
   colorMask?: [boolean, boolean, boolean, boolean] | boolean[];
 
   cull?: boolean;
-  cullFace?: GLCullFaceMode;
+  cullFace?: GL.FRONT | GL.BACK | GL.FRONT_AND_BACK;
 
   depthTest?: boolean;
   depthFunc?: GLFunction;
@@ -610,13 +610,15 @@ type EXT_depth_clamp = {
 type WEBGL_provoking_vertex = {
   // Constants in GL enum
   /** Set the provoking vertex */
-  provokingVertexWEBGL(provokeMode: GLProvokingVertex): void;
+  provokingVertexWEBGL(
+    provokeMode: GL.FIRST_VERTEX_CONVENTION_WEBGL | GL.LAST_VERTEX_CONVENTION_WEBGL
+  ): void;
 };
 
 /** WEBGL_polygon_mode https://registry.khronos.org/webgl/extensions/WEBGL_polygon_mode/ */
 type WEBGL_polygon_mode = {
   /** Set polygon mode of face to fill or line */
-  polygonModeWEBGL(face: GL.FRONT | GL.BACK, mode: GLPolygonMode): void;
+  polygonModeWEBGL(face: GL.FRONT | GL.BACK, mode: GL.LINE_WEBGL | GL.FILL_WEBGL): void;
 };
 
 /** WEBGL_clip_cull_distance https://registry.khronos.org/webgl/extensions/WEBGL_clip_cull_distance/ */
--- a/modules/constants/src/webgl-types.ts
+++ b/modules/constants/src/webgl-types.ts
@@ -206,6 +206,10 @@ export type GLStencilOp =
   | GL.DECR_WRAP
   | GL.INVERT;
 
+export type GLPolygonMode = GL.FILL_WEBGL | GL.LINE_WEBGL;
+export type GLCullFaceMode = GL.FRONT | GL.BACK | GL.FRONT_AND_BACK;
+export type GLProvokingVertex = GL.FIRST_VERTEX_CONVENTION_WEBGL | GL.LAST_VERTEX_CONVENTION_WEBGL;
+
 /** Parameters for textures and samplers */
 export type GLSamplerParameters = {
   /** Sets the wrap parameter for texture coordinate  to either GL_CLAMP_TO_EDGE, GL_MIRRORED_REPEAT, or GL_REPEAT. */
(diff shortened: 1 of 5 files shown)


donmccurdy:
* **#2040** <a href="https://app.graphite.dev/github/pr/visgl/luma.gl/2040?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

modules/webgl/src/adapter/converters/device-parameters.ts
```diff
@@ -156,10 +164,14 @@ export function setDeviceParameters(device: Device, parameters: Parameters) {
     const ext = extensions.WEBGL_provoking_vertex;
 
     if (parameters.provokingVertex) {
-      const vertex = map('provokingVertex', parameters.provokingVertex, {
-        first: GL.FIRST_VERTEX_CONVENTION_WEBGL,
-        last: GL.LAST_VERTEX_CONVENTION_WEBGL
-      });
+      const vertex = map<ProvokingVertex, GLProvokingVertex>(
```
ibgreen, line 167: Can't we deduce the type?

modules/core/src/adapter/types/parameters.ts
```diff
@@ -130,8 +132,10 @@ export type BlendFactor =
   | 'dst-alpha'
   | 'one-minus-dst-alpha'
   | 'src-alpha-saturated'
-  | 'blend-color'
-  | 'one-minus-blend-color';
+  | 'constant-color'
+  | 'one-minus-constant-color'
+  | 'constant-alpha'
+  | 'one-minus-constant-alpha';
```
donmccurdy, line 138: ⚠️ NOTICE

I don't see `blend-color` in the WebGPU spec, but probably this is what was intended for luma's constants? The exact [GPUBlendFactor values of the WebGPU spec](https://www.w3.org/TR/webgpu/#enumdef-gpublendfactor) are a bit different, so if we meant for this to be a 1:1 mapping to WebGPU, more changes would be needed.

modules/webgl/src/adapter/converters/device-parameters.ts
```diff
@@ -169,9 +181,9 @@ export function setDeviceParameters(device: Device, parameters: Parameters) {
     const ext = extensions.WEBGL_polygon_mode;
 
     if (parameters.polygonMode) {
-      const mode = map('polygonMode', parameters.provokingVertex, {
+      const mode = map<PolygonMode, GLPolygonMode>('polygonMode', parameters.polygonMode, {
```
ibgreen, line 184: I guess we could alternatively just declare the result variable type 'constant mode: polygonmode = map...' to to force a check

modules/webgl/src/adapter/converters/device-parameters.ts
donmccurdy: By declaring the type, we force TypeScript errors if any enum values have been missed — note the "lint" -> "line" correction earlier in the file. If we infer types from the inputs, nothing is validated.

modules/core/src/adapter/types/parameters.ts
ibgreen: I can't check now but generally the goal is to follow webgpu and if support for some webgl-only constant is needed, we suffix it with -webgl.

So... This could just be a simple mistake on my part, but also I started writing some of this code before  the spec was frozen and there have already been a few places where the luma code had to be updated.

modules/webgl/src/adapter/converters/device-parameters.ts
donmccurdy: Yes – I think that approach would validate `V` for our `Record<K, V>` mapping. But alone, it wouldn't tell us whether we've covered every possible `K`.

modules/constants/src/webgl-types.ts
```diff
@@ -364,7 +368,7 @@ export type GLFunctionParameters = {
   colorMask?: [boolean, boolean, boolean, boolean] | boolean[];
 
   cull?: boolean;
-  cullFace?: GL.FRONT | GL.BACK | GL.FRONT_AND_BACK;
+  cullFace?: GLCullFaceMode;
```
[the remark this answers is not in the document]
donmccurdy: Oh, that makes sense! I've removed the intermediate types here, so the exposed APIs in this file are explicit. I have kept some aliases above, so that `device-parameters.ts` has a source of truth to typecheck against for misspellings and such.

modules/core/src/adapter/types/parameters.ts
donmccurdy: Ok! It's probably easier to have two Luma parameters map to the same WebGPU constant, than to have the same Luma parameter map to two different WebGL constants depending on context. So I think I slightly prefer this as you have it, with a slight divergence from WebGPU for an easier mapping to WebGL.

PR #2045, opened 2024-03-19
Do not reset scissorTest in WEBGLRenderPass

Pessimistress:
This is causing a regression in deck.gl's picking and collision filter. Calling `device.beginRenderPass()` or `renderPass.setParameters()` will always set `scissorTest` to `false`.

It could be argued that `beginRenderPass` is expected to reset the state (questionable because it doesn't do this to any other parameter), however `setParameters()` as a public method should never affect a value not present in the argument.

Alternatively we can allow `scissorRect` to accept `null` on which we disable scissor test.

#### Change List
- Only sets `scissorTest` if `scissorRect` is present

### DIFF
--- a/modules/webgl/src/adapter/resources/webgl-render-pass.ts
+++ b/modules/webgl/src/adapter/resources/webgl-render-pass.ts
@@ -81,8 +81,8 @@ export class WEBGLRenderPass extends RenderPass {
         glParameters.viewport = parameters.viewport;
       }
     }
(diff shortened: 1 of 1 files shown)


ibgreen (approved):
Yes I think it makes sense. We want to emulate WebGPU parameter semantics in WebGL but WebGPU doesn't support inheriting a dirty context so WebGL module has some freedom to define how this works.

Some comments may be appropriate and maybe even a short section in the parameters page or separate about attaching to existing WebGL contexts

PR #2059, opened 2024-03-31
fix(shadertools) attribute picking regression

Pessimistress:
For https://github.com/visgl/deck.gl/issues/8704


ibgreen (approved):
There was a lot of hacking to get picking to work with uniform buffers, I suspect that is where this regression happened.

Note that our main example `examples/showcase/instancing` example uses picking, just make a quick check that it still works after this change.

PR #2067, opened 2024-04-05
add(core) enforceWebGL2

chrisgervang:
#### Background
Use enforceWebGL2 when a library is creating its own WebGL1 context privately, in order to create a WebGL2 context that is compatible with luma.gl.

#### Change List
- Adds enforceWebGL2

### DIFF
--- a/.ocularrc.js
+++ b/.ocularrc.js
@@ -13,7 +13,7 @@ const config = {
 
   lint: {
     paths: ['modules', 'docs', 'test', 'examples'],
-    extensions: ['js', 'ts', 'jsx', 'tsx'],
+    extensions: ['js', 'ts', 'jsx', 'tsx']
   },
 
   typescript: {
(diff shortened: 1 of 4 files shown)


Pessimistress:
What's the use case here? I think the latest versions of all supported base map libraries create WebGL2 by default.

chrisgervang:
This can be used when an application isn't ready to upgrade from Mapbox v1 but still needs to do interleaved rendering. We're trying to upgrade Hubble and Kepler to v9 but are blocked since MapLibre has an unresolved rendering issue.

ibgreen (approved):

ibgreen (approved):
This looks good, and since it unblocks the hubble.gl v9 upgrade I am strongly in favor of landing it. 
- I made a push on the docs. 
- Possible we add one test case?
- I called out that there may be an issue with some WebGL1 extensions not being available under WebGL2, potentially causing mapbox to not be able to use certain blend modes etc. It is not very hard to emulate these extensions, if we see any issues we could scan their code and see which extensions they are using.

chrisgervang:
Thanks! Yes, I'll add a test

PR #2109, opened 2024-06-20
feat(shadertools) Export Uniform types

felixpalmer:
#### Change List
- Export Uniform types


ibgreen (approved):
There are a few duplicates of types between core and Shadertools to help untangle the dependency so that shadertools can be completely independent of core (though unfortunately a small dependency has crept back in). 

These types might be some of those duplicated types. That might create a partial argument against exporting them.

felixpalmer:
@ibgreen [you're right](https://github.com/visgl/luma.gl/blob/a35c4dd0a28549d151cbdb3f0c329f5fe70338de/modules/core/src/gpu-type-utils/shader-types.ts#L41-L62), I can directly import these from core then

PR #2159, opened 2024-08-06
shadertools: lighting module fixes

felixpalmer:
For https://github.com/visgl/deck.gl/pull/9074

#### Change List
- Rename modules to fit convention & UBO blocks
- Consistent use of early returns in shaders
- Correct order of props in `uniformTypes`
- Swap `GOURAUD_VS` and `GOURAUD_FS`


ibgreen (approved):
LGTM, just make sure master is aligned with your naming changes etc.

PR #2181, opened 2024-08-11
Tex3d patch

xinaesthete:
For #2115 
#### Change List
In `webgl-texture.ts`, `setImageData` was being called with `format: glFormat` which in the `'3d'` case would cause a problem.

There were also a couple of other changes that seemed to be necessary to get the example to work.

### DIFF
--- a/examples/api/texture-3d/app.ts
+++ b/examples/api/texture-3d/app.ts
@@ -1,7 +1,7 @@
 // Ported from PicoGL.js example: https://tsherif.github.io/picogl.js/examples/3Dtexture.html
 
 import {makeRandomNumberGenerator, glsl} from '@luma.gl/core';
-import {AnimationLoopTemplate, AnimationProps, Model} from '@luma.gl/engine';
+import {AnimationLoopTemplate, AnimationProps, Geometry, Model} from '@luma.gl/engine';
 import {Matrix4, radians} from '@math.gl/core';
 import {perlin, lerp, shuffle, range} from './perlin';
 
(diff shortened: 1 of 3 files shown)


ibgreen:
@xinaesthete Thanks for the PR! 

Does the example actually work for you? I just get a white cube. It is supposed to show an animated stream of colors, you should be able to run it on 8.5-release to see the original version.

image

ibgreen:
The version on master partially works:
https://luma.gl/examples/api/texture-3d/


image

Original version looked like this:
image

xinaesthete:
It was working for me a couple of weeks ago, just realised I am seeing the same thing - apologies... it definitely was working with some variant of tweak to a shader, should be quick to fix...

xinaesthete:
Screenshot 2024-08-12 at 10 19 36

That's what it looks like with a different alpha
```
  float alpha = texture(uTexture, vUV + vec3(0.0, 0.0, uTime)).r;// * 0.1;
```

xinaesthete:
So it seems to me that the actual 3D texture is ok, there may be other issues with the example not setting all of the blend parameters in an equivalent way.

ibgreen:
Thanks! Hmm, I wonder if we are just dealing with a factor of 255? shader sees 0-1 instead of 0-255 due to normalization?

ibgreen:
Fixed. #2184

xinaesthete:
> Thanks! Hmm, I wonder if we are just dealing with a factor of 255? shader sees 0-1 instead of 0-255 due to normalization?

I'm not quite sure in this particular case - that certainly was something I saw at one point, but here if I use a factor anywhere near that magnitude it definitely isn't right. For the rendering to look like the original it would surely need a different blend mode, so it seems like it may well be that there is another issue with how these parameters are set.

xinaesthete:
FWIW, the reason it appears white when the alpha isn't adjusted in the original sample is that it's writing zero-alpha for those fragments and showing the background colour of the HTML body itself. I think intuitively it's easy to see something that looks like that and imagine it's blown out, but it's the opposite in this case.

PR #2187, opened 2024-08-14
chore(core): follow up on device props docs and exports

chrisgervang:
For #1946 

#### Change List
- docs
- exports


chrisgervang:
Ah, not sure why yarn version changed. Will fix

ibgreen (approved):

PR #2285, opened 2024-11-07
chore: Revert Modernize Canvas handling

felixpalmer:
#### Background

Reverts the following commits to reinstate the old canvas resizing:

* #2237
* #2226
* #2238
* #2239
* #2245
* #2257
* #2266
* #2277


ibgreen (commented):
That is a lot of commits to back out. Hopefully it won't cause more pain than it solves.

Issue #2295, opened 2024-12-06
[Bug] WebGLDevice resizes external gl context


Pessimistress:
### Module and/or Feature

@luma.gl/webgl

### Description

(Originally reported on Slack)
When using deck.gl `MapboxOverlay` interleaved with maplibre, as soon as the overlay is added to MapLibre, the visible map is reduced to ca. 2278px width.

Cause:
`WebGLDevice` constructor calls `canvasContext.resize()` without any parameters, which leads to resizing the canvas to match `window.devicePixelRatio`.

### Expected Behavior

`WebGLDevice` should not automatically resize canvas if an external context is provided.

### Steps to Reproduce

https://gitlab.com/alexander.muedespacher/deck-gl-maplibre-demo

### Environment

- Framework version: luma.gl@9.0.27
- Browser: Chrome 130
- Node: 20
- OS: MacOS

PR #2296, opened 2024-12-06
fix(webgl): Prevent auto resizing when using external gl context (9.0-release)

Pessimistress:
For #2295

#### Change List
- Disable `autoResize` when using an external WebGL context
- `cssToDeviceRatio()` fallback to canvas clientWidth when dimension cache is unavailable (this happens when resize() is never called)

### DIFF
--- a/modules/webgl/src/adapter/webgl-device.ts
+++ b/modules/webgl/src/adapter/webgl-device.ts
@@ -203,7 +203,8 @@ ${device.info.vendor}, ${device.info.renderer} for canvas: ${device.canvasContex
 
     // Create and instrument context
     const canvas = props.gl?.canvas || props.canvas;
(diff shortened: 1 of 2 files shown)


modules/webgl/src/adapter/webgl-device.ts
```diff
@@ -203,7 +203,8 @@ ${device.info.vendor}, ${device.info.renderer} for canvas: ${device.canvasContex
 
     // Create and instrument context
     const canvas = props.gl?.canvas || props.canvas;
-    this.canvasContext = new WebGLCanvasContext(this, {...props, canvas});
+    const autoResize = !props.gl;
```
chrisgervang, line 206: Is `autoResize` possibly overridden by the user in device props?

chrisgervang (approved):

ibgreen (approved):

PR #2333, opened 2025-02-13
feat(textures): Support compressed textures, add example

donmccurdy:
Related:

- Fixes https://github.com/visgl/luma.gl/issues/2316#issuecomment-2657066743

Fixes rendering of compressed textures for ASTC4x4, BC1, BC3, BC5, BC7, and (uncompressed) RGBA8.

Not implemented yet:

- RGBA16
- RGBA32
- ETC2
- mipmaps
- WebGPU

Includes a minimal example, showing how to create a texture with compressed data extracted from a KTX2 file. Using loaders.gl instead of ktx-parse would of course be reasonable, but we need some fixes in loaders.gl as well, and luma.gl seemed like the place to start.


package.json
```diff
@@ -64,5 +64,6 @@
   "volta": {
     "node": "20.17.0",
     "yarn": "4.5.1"
-  }
+  },
+  "packageManager": "yarn@4.5.1"
```
donmccurdy, line 68: Optional, but this does make my life a bit easier, I'm having to enable/disable Volta when working in different repositories. Not sure if there's a better approach? Could be moved to a draft PR somewhere for discussion too.

- https://github.com/nodejs/corepack/issues/507#issuecomment-2625366567
- https://github.com/volta-cli/volta/issues/987

modules/webgl/src/adapter/resources/webgl-texture.ts
```diff
@@ -147,10 +147,14 @@ export class WEBGLTexture extends Texture {
...
     const glTarget = getWebGLCubeFaceTarget(this.glTarget, this.dimension, depth);
 
-    const glParameters: GLValueParameters = {
-      [GL.UNPACK_ROW_LENGTH]: options.bytesPerRow,
-      [GL.UNPACK_IMAGE_HEIGHT]: options.rowsPerImage
-    };
+    const glParameters: GLValueParameters = {};
+    // The parameters below will break for compressed textures, as well as
+    // uncompressed textures like `new Uint8Array(2 * 2 * 4).fill(220)`. Are
+    // these parameters needed for other cases?
```
ibgreen, line 153: These parameters are there to match the capabilities of the native WebGPU API.
Can we code it so that we only set them if they are provided by the user?

package.json
ibgreen: Yes, separate PR would be good for this, as this would be something that eventually should be replicated in all our repos.

ibgreen (approved):
The example images are somewhat big. 
- if only for an example we could load them from one of the `{deck.gl, loaders.gl}-data` repos.
- The reason for adding them to this repo would be to use them in tests. If so recommend putting them in a test/data directory, such as `core/test/data/textures/...` and load them via rawgithub urls

donmccurdy:
This current image is 512x512 with mipmaps, without zstd compression to simplify the example. Particularly for the uncompressed RGBA8 sample I agree that's too large, and these images should be replaced.

If you don't mind the image being trivial (8x8 color test grid?) we could keep the images under 1kb each. Or I could create more visually "interesting" images (name of each format on a colored background?) at moderate resolution, and then either compress them with ZSTD or host them somewhere else to avoid repository bloat.

I'm happy to do any of those. I was hoping to write up a unit test as well, so probably having something small in this repo would be useful. Do you have any preference?

ibgreen:
Sure. That said they are not THAT big. Just a few MB. If we use them in both unit tests and example then I think it is fine (the small extra bloat is "worth it" so to speak).

donmccurdy:
Sounds good! I don't like the current images that much anyway they say "ETC1S" because that's what they were generated from. I'll work up a cleaner/smaller set of images and update the PR, along with addressing the other couple comments. Thanks!

donmccurdy:
@ibgreen I've cleaned up the example images — they're now all in the 1-2 kb range (rgba16 and rgba32 proportionately bigger), without zstd compression. All hosted in the https://github.com/donmccurdy/KTX2-Samples, along with source files and steps to regenerate them. Preview:

preview_2d

ibgreen:
Looks like CI is hanging, I tried to merge in master...

ibgreen:
@donmccurdy Tried to get the example to work, without success.

ibgreen:
@donmccurdy I got it to work, but I only get a single image, I don't see a loop. I might have clobbered something in your new app.ts, if your local branch is still available you may want to compare.

I mostly had to fight a bunch of weird issues with vite config trying to load from a sub directory, etc.

- It looks like WebGL ignores the bytesPerRow and rowsPerImage so I removed your change to omit those.
- Also, in WebGPU these apparently still need to be provided , but now as width / blockWIdth and height / blockHeight so that will take more work. 

FWIW `device.getTextureFormatInfo()` does an effort of calculating blockWidth and blockHeight, not sure if it covers all cases.

PR #2338, opened 2025-02-18
fix(webgl): Do not set drawBuffers for external Framebuffer

felixpalmer:
The google-maps-overlay in deck sets `WebGLFramebuffer` as `Deck._framebuffer` in order to render into it: https://github.com/visgl/deck.gl/blob/master/modules/google-maps/src/google-maps-overlay.ts#L263-L268

This was working in luma 9.0, but 9.1 crashes as the `WebGLRenderPass` expects a luma.gl `WEBGLFramebuffer` object here: https://github.com/visgl/luma.gl/blob/master/modules/webgl/src/adapter/resources/webgl-render-pass.ts#L48-L50

This fix simply skips the `this.device.gl.drawBuffers()` calls when the passed `this.props.framebuffer` is not of the right type


Pessimistress (approved):

ibgreen (commented):
Could have been good with a comment explaining why the code has to handle this case.

PR #2402, opened 2025-07-02
fix(engine): AsyncTexture createTexture helper

felixpalmer:

#### Background

Currently when creating an `AsyncTexture` the Promise never resolves. Additionally to make it easier to work with the async initialization a helper `AsyncTexture.createTexture` is added, allowing this usage:

```ts
// create new texture
if (!this._texture) {
  const asyncTexture = await AsyncTexture.createTexture(this.device, {
    format: 'rgba8unorm',
    dimension: '2d',
    data: null,
    width: this._canvasWidth,
    height: this._canvasHeight,
    sampler: this._samplerParameters || DEFAULT_SAMPLER_PARAMETERS,
    mipmaps: true
  });
  this._texture = asyncTexture.texture;
}
```

Old signature: https://github.com/visgl/deck.gl/blob/master/modules/layers/src/icon-layer/icon-manager.ts#L414-L423

#### Change List
- Resolve `this.resolveReady`
- Add `AsyncTexture.createTexture`

### DIFF
--- a/modules/engine/src/async-texture/async-texture.ts
+++ b/modules/engine/src/async-texture/async-texture.ts
@@ -172,11 +172,12 @@ export class AsyncTexture {
   }
 
   async initAsync(props: AsyncTextureProps): Promise<void> {
-    let resolveReady;
-    let rejectReady;
-
     const asyncData: AsyncTextureData = props.data;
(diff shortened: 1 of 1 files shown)


ibgreen (approved):
Approved, thanks. The new method might attract some bike shedding, but fine to land if you feel it helps.

PR #2404, opened 2025-07-07
fix(lighting): Default to 3 MAX_LIGHTS

felixpalmer:

#### Background

https://github.com/visgl/luma.gl/pull/2190/files#diff-a9db84520c6b1bac02c1add1e117b9b1fbe19df69521554b26129f0cdc44ddf6 changed the default number of `MAX_LIGHTS` in the lighting module to 1. The default deck.gl `LightingEffect` uses 2 so this change makes the lighting look very different

#### Change List
- Default back to 3 max lights
- Consistent naming of `ambientColor` between TS and GLSL

### DIFF
--- a/modules/shadertools/src/modules/lighting/lights/lighting.ts
+++ b/modules/shadertools/src/modules/lighting/lights/lighting.ts
@@ -59,7 +59,7 @@ export type LightingProps = {
 
 export type LightingUniforms = {
   enabled: number;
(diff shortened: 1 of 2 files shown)


ibgreen (approved):
Fine for now.

The goal is to have the light infos as a dynamic length array of structs at the end of the uniform buffer, once our shader type system supports that.

PR #2405, opened 2025-07-08
fix(core): Support number for useDevicePixels

felixpalmer:

#### Background

deck.gl expects to be able to pass a number for `useDevicePixels`. Without this it is impossible to force HD rendering on a low def screen (or in tests). 

Repro of `'scatterplot-smoothedge'` test case in browser with DPR configured to 1:

### useDevicePixels: `2` (and fix in this PR)

Screenshot 2025-07-08 at 16 45 11

### useDevicePixels: `1 | true | false`
Screenshot 2025-07-08 at 16 44 50

I would argue that `getDevicePixelRatio()` should be removed - it is not currently used and I think trying to add it in the code block where I made the change makes the code harder to read

#### Change List
- Manually set DPR when `useDevicePixels` is a number

### DIFF
--- a/modules/core/src/adapter/canvas-context.ts
+++ b/modules/core/src/adapter/canvas-context.ts
@@ -23,8 +23,8 @@ export type CanvasContextProps = {
   height?: number;
   /** Visibility (only used if new canvas is created). */
   visible?: boolean;
-  /** Whether to size the drawing buffer to the pixel size during auto resize */
-  useDevicePixels?: boolean;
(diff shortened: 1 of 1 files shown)


ibgreen (approved):
OK. 

As you can see, I was trying to get rid of the number variant. 

Both because overloading types complicates things in TypeScript, and it seemed like the only use case was to support testing. 

Might have been cleaner to add a separate `_devicePixelRatio` prop instead, clearly marked as intended for testing.

PR #2406, opened 2025-07-09
feature(core): CanvasContext.setProps({useDevicePixels})

felixpalmer:
Followup to https://github.com/visgl/luma.gl/pull/2405, to avoid this hack: https://github.com/visgl/deck.gl/pull/9241/commits/1e786f2b39576a4b412c4b8acf4c0f727895d99d

#### Change List
- Provide way to update `useDevicePixels` prop


ibgreen (approved):
OK. 

Complexity is creeping back in to this surprisingly complicated class.
I think more props could be exposed to setProps, I should make pass later on.

PR #2458, opened 2025-10-29
Bump @vis.gl/ts-plugins

Pessimistress:
For https://github.com/visgl/deck.gl/issues/9840

Can confirm the build output now has 

modules/webgl/dist/adapter/webgl-adapter.js:
```js
        const { WebGLDevice } = await import("./webgl-device.js");
```

modules/webgpu/dist/adapter/webgpu-adapter.js:
```js
        const { WebGPUDevice } = await import("./webgpu-device.js");
```

#### Change List
- Upgrade `@vis.gl/ts-plugin` to bring in support for dynamic imports


ibgreen-openai (approved):
No longer alpha releases, great!

PR #2484, opened 2026-01-08
fix(engine): Avoid crash in generateMipmaps()

felixpalmer:
#### Background

While fixing another issue I notice the `hello-cube` test app is broken. Minimal fix to get it working again. The app will now log `Texture(dynamic-texture-1,rgba8unorm,512x512): generateMipmaps not supported in WebGPU` rather than crashing

#### Change List
- Use generateMipmapsWebGL() even in WebGPU codepath rather than throwing error
- Copy `failIfMajorPerformanceCaveat` prop

### DIFF
--- a/modules/engine/src/dynamic-texture/dynamic-texture.ts
+++ b/modules/engine/src/dynamic-texture/dynamic-texture.ts
@@ -236,7 +236,11 @@ export class DynamicTexture {
   generateMipmaps(): void {
     // Call the WebGL-style mipmap generation helper
     // WebGL implementation generates mipmaps, WebGPU logs a warning
-    this.texture.generateMipmapsWebGL();
+if (this.device.type === 'webgl') {
+      this.texture.generateMipmapsWebGL();
+    } else {
+      log.warn('Mipmap generation not yet implemented on WebGPU: your texture data will not be correctly initialized');
+    }
   }
 
   /** Set sampler or create one from props */


modules/engine/src/dynamic-texture/dynamic-texture.ts
```diff
@@ -234,12 +234,9 @@ export class DynamicTexture {
...
   generateMipmaps(): void {
-    // Only supported via WebGL helper (luma.gl path)
-    if (this.device.type === 'webgl') {
-      this.texture.generateMipmapsWebGL();
-    } else {
-      throw new Error('Automatic mipmap generation not supported on this device');
-    }
+    // Call the WebGL-style mipmap generation helper
+    // WebGL implementation generates mipmaps, WebGPU logs a warning
+    this.texture.generateMipmapsWebGL();
```
ibgreen:
```suggestion
if (this.device.type === 'webgl') {
      this.texture.generateMipmapsWebGL();
    } else {
      log.warn('Mipmap generation not yet implemented on WebGPU: your texture data will not be correctly initialized');
    }
```


- The goal is to implement programmatic mipmap generation in DynamicTexture (there is already some WIP code for that) and to use that on webgpu devices.
- Therefore, I think we should log a warning in DynamicTexture.generateMipmaps() instead of calling the WebGL specific function. 
- Conversely, we should throw, not warn, in generateMipmapsWebGL() when running on WebGPU.

PR #2486, opened 2026-01-14
fix(core): Auto-convert uint8 buffers to uint16

felixpalmer:
#### Background

WebGPU doesn't support Uint8Array buffers, but the GLTF loader can produce them (as could any app). Many small 3D tiles end up having a Uint8Array index buffer which causes luma to throw an error:

```
deck: initialization of ScenegraphLayer({id: 'google-3d-tiles-scenegraph-https://tile.googleapis.com/v1/3dtiles/datasets/XXX.glb'}): indices buffer content must be of type uint16 or uint32 Error: indices buffer content must be of type uint16 or uint32
    at WebGLDevice._normalizeBufferProps (device.ts:762:15)
``` 

As a result the tile doesn't load and there is a hole in the terrain:

Screenshot 2026-01-14 at 09 11 15

To resolve this, this PR automatically converts uint8 buffers to uint16. Given that the buffers will typically be small, the extra memory isn't significant.

After this PR:

Screenshot 2026-01-14 at 09 11 38

#### Change List
- Convert uint8 to uint16 for WebGPU compatibility
- Test case

### DIFF
--- a/docs/api-reference/core/resources/buffer.md
+++ b/docs/api-reference/core/resources/buffer.md
@@ -6,13 +6,13 @@ Holds a block of GPU memory. The length of a buffer cannot be changed after crea
 
 ### `BufferProps`
 
-| Property      | Type                             | Description                                                                  |
-| ------------- | -------------------------------- | ---------------------------------------------------------------------------- |
-| `usage?`      | `number`                         | Bit mask of Usage flags                                                      |
-| `byteLength?` | `number`                         | Length of buffer (cannot be changed after creation).                         |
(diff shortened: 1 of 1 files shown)


felixpalmer:
> Does this need to be mentioned in the docs, or are there no consequences to a user?

Done: https://github.com/visgl/luma.gl/pull/2486/commits/068be397f1ff278171b74f1c6eebcdfb6e6928bc

In general I don't expect the user will notice, this is more removing a gotcha

Issue #2340, opened 2025-02-19
[Feat] Generic GPU-based data manipulation


Pessimistress:
## Motivation

deck.gl has multiple feature areas that would benefit from a shared & standardized computation module. Example use cases:

- To consume data input in memory-efficient formats such as Arrow and Parquet
- Attribute animation/transition (currently implemented with ad-hoc transforms)
- Data aggregation (currently implemented with ad-hoc transforms)
- To perform one-time data transform such as 64bit split, coordinate system conversion (currently implemented on the CPU)

## Proposal

Create a new module `@luma.gl/gpgpu`.

> The proposed syntax is strongly inspired by [tensorflow.js](https://js.tensorflow.org/api/latest/), especially the functions in Creation, Slicing and Joining, Arithmetic, Basic Math & Reduction

### Example API

Creation: returns a wrapper of a GPU buffer

```ts
// A column with stride of 1, example: Int32Array [property0, property1, property2, ...]
gpu.array1d(data: TypedArray): GPUAttribute
// A column with stride > 1, example: Float64Array [x0, y0, x1, y1, x2, y2, ...]
gpu.array2d(data: TypedArray, shape: [number, number]): GPUAttribute
// Constant scaler|vec2|vec3|vec4
gpu.constant(value: number | NumericArray): GPUAttribute
```

Reshape: joining, slicing and/or rearranging GPU buffers

```ts
// Interleaving multiple columns, example: [x0, x1, x2, ...] + [y0, y1, y2, ...] -> [x0, y0, x1, y1, x2, y2, ...]
gpu.stack(values: GPUAttribute[]): GPUAttribute
// N-D slicing of interleaved buffer, example: [a0, b0, c1, a1, b1, c1, a2, b2, c2, ...] -> [a0, c0, a1, c1, c2, c2, ...]
gpu.slice(value: GPUAttribute, begin: number[], size: number[]): GPUAttribute
```

Transform: math operations on GPU buffers

```ts
// element-wise add
gpu.add(value: GPUAttribute, 1): GPUAttribute
// min value across dimensions
gpu.min(value: GPUAttribute): number
```

### Interface with loaders.gl

There is no direct dependency on loaders.gl, but the module can be "loaders friendly" by accepting a `Table` shaped input:

```ts
gpu.array1d(table: Table, columnNameOrIndex: string | number): GPUAttribute
```

### Interface with deck.gl

[...]

PR #2594, opened 2026-04-06
feat(gpgpu): add module for generic data processing

Pessimistress:
For #2340

#### Change List
- Bootstrap new module
- Foundation classes
- Operations for demonstration: add, interleave, fround
- Unit tests
- Docs

image


.yarnrc.yml
```diff
@@ -10,3 +10,5 @@ supportedArchitectures:
     - darwin
     - linux
     - win32
+
+yarnPath: .yarn/releases/yarn-4.13.0.cjs
```
ibgreen: is this needed?

modules/gpgpu/tsconfig.json
```diff
@@ -8,6 +8,10 @@
     "outDir": "dist"
   },
   "references": [
-    {"path": "../engine"}
+    {"path": "../core"},
+    {"path": "../engine"},
+    {"path": "../constants"},
+    {"path": "../shadertools"},
+    {"path": "../webgpu"}
```
ibgreen, line 15: We should avoid depending on the backend if we can. Let's look at how we could uncouple this.

modules/gpgpu/test/transforms/texture-transform.spec.ts
```diff
@@ -0,0 +1,205 @@
+// luma.gl
```
ibgreen: Are we copying these classes from engine into gpgpu?

modules/gpgpu/test/transforms/swap.spec.ts
```diff
@@ -0,0 +1,45 @@
+import {expect, test} from 'vitest';
```
ibgreen: Are we copying these classes from engine into gpgpu?

ibgreen (commented):
First comment:
Can we avoid copying all the Transform, Computation, Swap etc classes?
Let's make a quick PR that fixes issues with those in the engine module and get that landed first.
I am not concerned about minor breaking changes in those classes, they are in experimental state and unlikely to be used in application, and anyone who is actually using them would surely appreciate any improvements we make.

ibgreen:
I had codex look at the copied classes

What Changed
The PR adds a new @luma.gl/gpgpu module with:

WebGL backend using a copied BufferTransform
WebGPU backend using a copied Computation
Direct copies of TextureTransform, Swap, SwapBuffers, SwapFramebuffers
Copied/converted tests for those transform classes
Some unrelated package-manager churn: .yarn/releases/yarn-4.13.0.cjs, .yarnrc.yml, and large yarn.lock changes
Upstream Candidates
BufferTransform has two changes worth upstreaming into buffer-transform.ts (line 14):

Add support for interleaved transform feedback via a friendly feedbackBufferMode?: 'interleaved' | 'separate'
Widen run({outputBuffers}) from Record<string, Buffer> to Record<string, Buffer | BufferRange>
That said, engine already inherits numeric bufferMode through ModelProps, so the minimal alternative is for GPGPU to pass bufferMode: GL.INTERLEAVED_ATTRIBS without changing engine. The string prop is still nicer API, but it may require introducing or avoiding an engine dependency on @luma.gl/constants.

Do Not Upstream As-Is
Computation changes should not be upstreamed as written into computation.ts (line 34):

It imports getShaderLayoutFromWGSL from @luma.gl/webgpu, which would couple engine to a backend.
It removes mergeShaderModuleBindingsIntoLayout, likely breaking shader-module uniform/binding layout merging.
It removes the optional CommandEncoder path for uniform uploads, which current UniformStore.setUniforms() supports.
It creates managed uniform buffers for all modules, while engine currently guards on shaderModuleHasUniforms().
If layout inference is needed, the right upstream fix is probably a backend-neutral Device.getShaderLayout() / core path, not an engine import from @luma.gl/webgpu.

TextureTransform is effectively unchanged except import paths, so there is nothing to upstream into texture-transform.ts (line 11).

Swap should not be upstreamed. The PR copy is worse than swap.ts (line 14): it drops id, ignores supplied framebuffer width/height by creating 1x1 textures, drops generated texture ids, and changes one usage mask inconsistently.

ibgreen:
Updating the BufferTransform to use string constants seems like a good move!
We could add a getShaderLayoutWebGPU() method to the Device to unbreak the webgpu module coupling

modules/gpgpu/test/transforms/texture-transform.spec.ts
Pessimistress: Yes, at some point (a long time ago) the vision was that transforms should live in their dedicated module

Pessimistress: Also I started building the skeleton of this quite a while ago, maybe prior to 9.2, so the discrepancies noted by codex are not really surprising. I do not have strong preference either way.

docs/api-reference/gpgpu/gpu-table.md
```diff
@@ -0,0 +1,119 @@
...
+| `offset?` | `number` | Byte offset to the first element of the first row. Defaults to `0`. |
+| `stride?` | `number` | Byte distance between adjacent rows. Defaults to `ValueType.BYTES_PER_ELEMENT * size`. |
+| `value?` | `TypedArray` | CPU-side data for the table. Required unless `source` is provided. |
+| `source?` | `Operation \| GPUTable \| null` | Lazy data source for this table. |
+| `isConstant?` | `boolean` | Whether every row shares the same value. Defaults to `false`. |
+| `length?` | `number` | Row count. Optional when `isConstant` is `true` or `value` is provided. |
+
+## Static Methods
+
+### `GPUTable.fromArray(value, props?): GPUTable`
```
akre54, line 43: Would it be possible to add a `GPUTable.fromArrowColumn` to make it zero-copy to pass in an Arrow Vector from duckdb? It's already a contiguous typed array under the hood

docs/api-reference/gpgpu/operations.md
```diff
@@ -0,0 +1,94 @@
+# Operations
```
akre54, line 1: Can I also request:

* `mul(table, scalar)`:  scale a column
* `clamp(table, min, max)`: clip values
* `scaleLinear(table, domain, range)`: linear mapping
* `colorRamp(table, stops)`: map scalar to RGBA

docs/api-reference/gpgpu/operations.md
```diff
@@ -0,0 +1,94 @@
...
+]), {size: 3});
+
+const id = GPUTable.fromArray(new Float32Array([5, 6]), {size: 1});
+
+const result = interleave(xyz, id);
+```
+
+## `fround`
+
+### `fround(x: GPUTable): GPUTable`
```
akre54, line 71: Arrow returns `Int32`, `Int64`, `Float64` etc. GPU wants Float32 (positions/radii) or Uint8 (colors). Beyond `fround`, can we add `cast(table, 'float32')` and `normalizeToUint8(table, domain)`

docs/api-reference/gpgpu/operations.md
akre54: It would also be great to get `extent(table)` (min / max) and `normalize(table)` to get [0 .. 1]. I typically normalize the values in my columns to get a scalar that I can pass to `colorRamp`

akre54 (commented):
This is great! I've been looking for something like this for Noodles. Let me know if I can help test

Pessimistress:
@ibgreen cleaned up duplicate classes

ibgreen-openai:
I have rebased to latest and landed.

So far looks very good.

We should probably discuss:
- `GPUTable` naming. We will likely end up with multiple GPUTable type objects (I have so far added `GPUArrowTable`). I personally think that something called GPUTable should be concrete, not lazy evaluated. This would be more of a `LazyGPUTableEvaluator` and perhaps could return a more concrete table type when evaluated?
- The backend concept. Avoiding import of both types of shaders would be very desirable across luma.gl, however the backend approach seems like a specific solution that does not immediately extend to other areas.

PR #2596, opened 2026-04-08
docs: Don't mention removed constants module in upgrade-guide

felixpalmer:
This change didn't land in the end, see https://github.com/visgl/luma.gl/pull/2583


docs/upgrade-guide.md
```diff
@@ -19,7 +19,6 @@ luma.gl largely follows [SEMVER](https://semver.org) conventions. Breaking chang
 - glTF texture sampling now defaults to linear filtering when a glTF sampler omits explicit filter settings. Applications relying on the previous nearest-neighbor default should verify visual output and set sampler filters explicitly when nearest sampling is required.
 - The legacy feature flag `timer-query-webgl` has been removed. Replace checks for `timer-query-webgl` with `timestamp-query` for GPU timestamp/query support on both WebGPU and WebGL.
 - `PipelineFactory` and `ShaderFactory` now import from `@luma.gl/core` instead of `@luma.gl/engine`.
-- `@luma.gl/constants` has been removed. If raw numeric WebGL enums are still needed, import them from `@luma.gl/webgl/constants`.
```
ibgreen, line 22: @felixpalmer We can change this to saying that the module has been deprecated, not removed.

What happened was is that the official source of constants are now in `@luma.gl/webgl/constants`.

However, our dev tooling setup in deck.gl uses old typescript resolution configs that don't handle package.json `exports` sub path configs well.

So to unblock deck publish, I added back the constants module, but it should still be considered deprecated.

It is no longer a peer dependency, applications don't need to list it in their package.json anymore (e.g. deck examples no longer list it) unless the application directly imports GL constants from the deprecated module

PR #2597, opened 2026-04-08
fix(webgl): Do not assume luma WEBGLFramebuffer

felixpalmer:
#### Background

Fixes regression introduced first introduced in `9.3-alpha.10` breaking Google interleaved rendering together with https://github.com/visgl/deck.gl/pull/10253

Screenshot 2026-04-08 at 16 19 18

Error logged:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'map')
    at new WEBGLRenderPass (webgl-render-pass.ts:56:61)
    at WEBGLCommandEncoder.beginRenderPass (webgl-command-encoder.ts:53:12)
    at _WebGLDevice.beginRenderPass (device.ts:775:32)
    at DrawLayersPass._render (layers-pass.ts:109:36)
    at DrawLayersPass.render (draw-layers-pass.ts:15:17)
    at DeckRenderer.renderLayers (deck-renderer.ts:94:36)
    at Deck._drawLayers (deck.ts:1419:24)
    at google-maps-overlay.ts:344:16
    at withGLParameters (with-parameters.ts:37:13)
```

Repro using deck.gl `examples/basemap-browser`

#### Change List
- Don't autocreate textures for external framebuffer
- Prevent crashes when we have no colorAttachments

### DIFF
--- a/modules/webgl/src/adapter/resources/webgl-render-pass.ts
+++ b/modules/webgl/src/adapter/resources/webgl-render-pass.ts
@@ -54,16 +54,12 @@ export class WEBGLRenderPass extends RenderPass {
     this.setParameters({viewport, ...this.props.parameters});
 
     // Specify mapping of draw buffer locations to color attachments
-    if (webglFramebuffer instanceof WEBGLFramebuffer) {
-      if (!isDefaultFramebuffer) {
-        const drawBuffers = webglFramebuffer.colorAttachments.map(
-          (_, i) => GL.COLOR_ATTACHMENT0 + i
-        );
-        this.device.gl.drawBuffers(drawBuffers);
-      } else {
-        // Default framebuffer only supports GL.BACK/GL.NONE draw buffers
-        this.device.gl.drawBuffers([GL.BACK]);
-      }
+    if (!isDefaultFramebuffer && webglFramebuffer instanceof WEBGLFramebuffer) {
+      const drawBuffers = webglFramebuffer.colorAttachments.map((_, i) => GL.COLOR_ATTACHMENT0 + i);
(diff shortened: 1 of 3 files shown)


modules/webgl/src/adapter/resources/webgl-render-pass.ts
```diff
@@ -23,8 +23,11 @@ export class WEBGLRenderPass extends RenderPass {
   constructor(device: WebGLDevice, props: RenderPassProps) {
     super(device, props);
     this.device = device;
-    const webglFramebuffer = this.props.framebuffer as WEBGLFramebuffer | null;
-    const isDefaultFramebuffer = !webglFramebuffer || webglFramebuffer.handle === null;
+    // The framebuffer may be a luma.gl WEBGLFramebuffer, a raw WebGLFramebuffer
+    // handle (e.g. from an external context like Google Maps), or null (default).
+    const webglFramebuffer = this.props.framebuffer as WEBGLFramebuffer | WebGLFramebuffer | null;
```
ibgreen, line 28: @felixpalmer 
All the resource constructor should support a `handle` argumment.
If you already have a WebGLFramebuffer the idea was that you create a wrapper for it.
Is it possible to make that path work for this issue rather than make a framebuffer specific hack?

```ts
const webglFramebufferHandle = googleMaps.getFramebuffer();
webglDevice.createFramebuffer({handle: webglFramebufferHandle});
```

felixpalmer: @ibgreen I've reworked the fix to perform the wrapping in deck: https://github.com/visgl/deck.gl/pull/10253

PR #2628, opened 2026-05-19
feat(gpgpu) operations on GPUVectors

ibgreen-openai:
## Goals

- Add first-class `GPUVector` support to `GPUTableEvaluator`.
- Reuse existing gpgpu operations instead of adding a parallel GPUVector compute layer.
- Keep Arrow usage in gpgpu limited to `GPUVector` schema/type metadata.

image

## Changes

- Renamed `modules/gpgpu/src/operation/gpu-table.ts` to `gpu-table-evaluator.ts`.
- Added `GPUTableEvaluator.fromGPUVector()` for packed numeric `GPUVector` inputs.
- Added `GPUTableEvaluator.evaluateToGPUVector()` for GPU-resident output wrapping.
- Updated `add`, `interleave`, and `fround` to accept `GPUTableEvaluator | GPUVector`.
- Preserved output `GPUVector` logical type metadata where possible.
- Added interleaved `GPUVector` output support through `evaluateToGPUVector({interleaved})`.
- Exported `getGPUTableEvaluator` and GPUVector-related evaluator types from `@luma.gl/gpgpu`.
- Added browser tests covering add, interleave, fround, lazy chaining, and ownership behavior.
- Added `apache-arrow` dependency to gpgpu for schema/type metadata used by `GPUVector`.

## Verification

- `env -u YARN_NO_PROXY corepack yarn lint fix` passes.
- `env -u YARN_NO_PROXY corepack yarn build` passes.
- `env -u YARN_NO_PROXY corepack yarn test-headless modules/gpgpu/test/gpu-vector/gpu-vector-compute.spec.ts` passes, 8 tests.

## Risks / Follow-Ups

- Only packed numeric `GPUVector` inputs are supported for v1.
- Interleaved input vectors are rejected for now.
- `apache-arrow` in gpgpu is schema metadata only, not Arrow table/vector upload logic.


modules/gpgpu/package.json
```diff
@@ -48,6 +48,7 @@
     "@math.gl/core": "^4.1.0",
     "@math.gl/types": "^4.1.0",
     "@probe.gl/log": "^4.0.8",
-    "@probe.gl/stats": "^4.0.8"
+    "@probe.gl/stats": "^4.0.8",
+    "apache-arrow": "^17.0.0"
```
Pessimistress, line 52: No.....

ibgreen-openai: - As mentioned we use schemas and data types from apache-arrow.
- the table module / GPUVector already has this dependency, so removing it from here would be cosmetic only.
- If we want to get rid of this dependency, we should implement a parallel type system.
- I spent a lot of time in loaders.gl v4 trying to duplicate the arrow type system to avoid this dependency. It really slowed down my progress and in the end gained me little.

My position right now is that we go all in on arrow and just accept this dependency.

- If we wanted to make sure we just use the type system I could re-export it from the tables module, so we don't use anything else from arrow by mistake
- but it would be quite clunky as it is a fairly long list of exports.

ibgreen-openai: arrow usage updated to use explicit imports

PR #2661, opened 2026-06-04
feat(gpgpu) Backend registration changes

Pessimistress:
- Backend registration for built-in device types is no longer required. WebGL and WebGPU devices automatically fallback to the respective backend with dynamic import (separate bundle, only loaded on demand)
- Backend registry still exposed for custom device types and/or operations
- Backend imports are moved to dedicated endpoints: 

  ```ts
  // before
  import {webgpuBackend} from '@luma.gl/webgpu';
  // after
  import * as webgpuBackend from '@luma.gl/gpgpu/webgpu';
  ```

  This also allows users to pick and choose only the operations they need.

### DIFF
--- a/docs/api-reference/gpgpu/README.md
+++ b/docs/api-reference/gpgpu/README.md
@@ -21,7 +21,7 @@ Interleaving two buffers together
 ```ts
 import {luma} from '@luma.gl/core';
 import {webglAdapter} from '@luma.gl/webgl';
-import {GPUTableEvaluator, backendRegistry, webglBackend, add, interleave} from '@luma.gl/gpgpu';
+import {GPUTableEvaluator, add, interleave} from '@luma.gl/gpgpu';
 
 const inputA = GPUTableEvaluator.fromArray(new Float32Array([0, 0, 0, 1, 0, 0]), {size: 3});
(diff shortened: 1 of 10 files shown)


ibgreen-openai:
> * Backend registration for built-in device types is no longer required. WebGL and WebGPU devices automatically fallback to the respective backend with dynamic import (separate bundle, only loaded on demand)
> * Backend registry still exposed for custom device types and/or operations
> * Backend imports are moved to dedicated endpoints:
>   ```ts
>   // before
>   import {webgpuBackend} from '@luma.gl/webgpu';
>   // after
>   import * as webgpuBackend from '@luma.gl/gpgpu/webgpu';

- It is certainly a step in the right direction, certainly the webgpu module was never intended to be a container for random bits of webgpu related code.

>   ```
>     
>   This also allows users to pick and choose only the operations they need.

Yes, I still feel that slicing per backend and pushing the burden of managing that onto the app is undesirable. It is not a logical split, it is an implementation imposed split.

It seems to me that someone wanting to support both webgpu and webgp now needs to mess with composing mulitple backends.

But as discussed lets continue and see how it plays out:
- if this really provides the meaningful wins in terms of bundle size etc that we are hoping for
- and how different the operation implementations actually will end up being between the backends.

ibgreen-openai (approved):
Approved, added comments separately

PR #2664, opened 2026-06-04
feat(shadertools) ShaderPlugins

ibgreen-openai:


chrisgervang (commented):
I currently take for granted how easy it is to inject shaders with WebGL using luma.gl today within deck Extensions or custom layers - is ShaderPlugin a way to continue supporting this when using a WebGPU device?

ibgreen-openai:
> I currently take for granted how easy it is to inject shaders with WebGL using luma.gl today within deck Extensions or custom layers - is ShaderPlugin a way to continue supporting this when using a WebGPU device?

Supporting deck Extensions on WebGPU is complicated, so I am creating a similar concept in luma.gl (ShaderPlugins) minus the deck.gl lifecycle integration.

ShaderPlugins would likely be the "substrate" on top of which deck.gl v10 Extensions are re-implemented in a portable way (with minimal user facing changes), alternatively ShaderPlugins could just an end-to-end demonstration of how WebGPU can be supported in Extensions.

PR #2674, opened 2026-06-12
feat(experimental): HTMLTexture via HTML-in-Canvas

ibgreen:
image


chrisgervang (approved):
Very exciting. I think we'll see some cool applications of this overtime as support reaches maturity.

Were there any surprises to get the integration to work?

ibgreen-openai:
> Very exciting. I think we'll see some cool applications of this overtime as support reaches maturity.

Yes. We could also consider whether a deck widget integration could make sense - ability to embed the widgets in the 3D scene?

> Were there any surprises to get the integration to work?

-The most surprising part was that “clickable HTML texture” is not really texture picking.
- The texture is only the visual copy. 
- The actual interactive/accessibility surface is still the live DOM element, which must remain a direct child of the layoutsubtree canvas and have its CSS transform kept in sync with where its pixels are drawn. 
- Browser hit testing happens on that transformed DOM, not on the GPU quad.

The other surprise was API churn. The prose in the announcement was not enough to implement against reliably; the WICG examples showed the working paint / requestPaint() flow and the current WebGL/WebGPU copy signatures.

chrisgervang:
Oh interesting! I wonder the performance implications are with rapid DOM tree changes? Maybe they figured out a great solution for this.

I'm curious if we'll see similar upper limits to DOM-based visualizations (too many elements hurting performance).

Also curious if mapping the texture to extremely complex geometries would hurt performance.

Embedding widgets is an interesting idea - the first use case I can think of is embedding legends and charts in the canvas for screenshot and screen recording application (essentially, rastering widgets into screen space textures for capture)

PR #2677, opened 2026-06-16
feat(engine) add portable video textures

ibgreen-openai:
## Goals

Add portable engine-level video textures while letting the shader binding declaration select the backend representation.

## Changes

- Add engine `VideoTexture` for caller-owned `HTMLVideoElement` and `VideoFrame` sources.
- Resolve GLSL `sampler2D` and WGSL `texture_2d` through copied luma textures; resolve WGSL `texture_external` through native WebGPU external acquisition only.
- Keep `VideoTexture` v1 deliberately narrow: source, color space, sampler, pull-based frame observation, and a one-mip copied `rgba8unorm` texture. Use an ordinary `Texture` when full texture props or mipmaps are required.
- Reject copied luma textures for WebGPU `external-texture` bind-group slots.
- Add focused `VideoTexture` readiness, frame replacement, ownership, frame timestamp, and external-slot tests.
- Add video texture docs plus the experimental rotating cylinder example with procedural default video and optional camera input.

## Verification

- `nvm use` passed after sourcing `/Users/ib/.nvm/nvm.sh`.
- `yarn install` passed through `corepack yarn` with existing peer-dependency warnings.
- `yarn lint fix` passed through `corepack yarn`.
- `yarn test-node` passed.
- `yarn build` passed.
- `yarn examples:typecheck` passed.


docs/api-reference/core/resources/external-texture.md
```diff
@@ -8,25 +8,21 @@ import {CoreDocsTabs} from '@site/src/components/docs/core-docs-tabs';
...
-    sampler
-  ])
+  model.setBindings({videoTexture: externalTexture});
 
   model.draw(renderPass);
 }
 requestAnimationFrame(onFrame);
 ```
+
+`ExternalTextureProps` accepts `source?: HTMLVideoElement | VideoFrame`, `colorSpace?: 'srgb'`, an optional default `sampler`, and normal `ResourceProps`. Handle-backed opaque WebGPU external textures also require `width` and `height` when luma cannot infer them from a source.
```
ibgreen-openai, line 28: @chrisgervang

docs/api-guide/gpu/video-textures.mdx
```diff
@@ -0,0 +1,87 @@
+# Working With Video Textures
```
chrisgervang, line 1: Very exciting to see support come back for this!! We could bring back the deck.gl example that rendered 360 video on a sphere with an orbit view

Or, I think an even better use would be demonstrating a time series plot synchronized with video frames

docs/api-guide/gpu/video-textures.mdx
```diff
@@ -0,0 +1,87 @@
...
+@group(0) @binding(auto) var videoTextureSampler: sampler;
+let color = textureSampleBaseClampToEdge(videoTexture, videoTextureSampler, uv);
+```
+
+When a `VideoTexture` resolves against `texture_external`, luma.gl tries to acquire a native WebGPU [`GPUExternalTexture`](https://gpuweb.github.io/gpuweb/#gpu-external-texture). A copied `Texture` cannot satisfy that WebGPU bind-group slot, so use a `texture_2d<f32>` shader binding when native external import is unavailable or copied texture behavior is required.
+
+This native path is an optimization, not a more general texture type. The [WebGPU Fundamentals external-video guide](https://webgpufundamentals.org/webgpu/lessons/webgpu-textures-external-video.html) explains why: browsers often receive decoded video in YUV-like planes rather than already-expanded RGBA pixels. `texture_external` lets WebGPU sample that browser-owned representation directly and perform the needed conversion during sampling instead of first copying every frame into an RGBA texture. The browser implementation decides whether a copy is avoided, but the API exists so it can avoid that conversion/copy when possible.
```
chrisgervang, line 56: Nice

docs/api-guide/gpu/video-textures.mdx
```diff
@@ -0,0 +1,87 @@
...
+- Sampling uses `textureSampleBaseClampToEdge`, so the native path is base-level and clamp-style. It does not provide mipmaps or ordinary repeat sampling.
+- Native external acquisition can force bind-group invalidation because the concrete external binding may change between draws.
+- `VideoTexture` copies frames during draw binding resolution. WebGL can generate copied video mipmaps there, but WebGPU `VideoTexture` rejects `mipmaps: true` because WebGPU mipmap generation must be recorded before the render pass. Upload into an ordinary `Texture` before `beginRenderPass()` when WebGPU video mipmaps are required.
+
+The WebGPU Fundamentals article also notes two useful implementation details: an external texture may conceptually hide multiple underlying video planes, and WebGPU can inject the format conversion needed to return shader-visible RGBA. That is why an external texture needs special WGSL syntax instead of pretending to be a normal `texture_2d<f32>`.
+
+Use the copied path when flexibility matters. Use `texture_external` when the shader can accept its restrictions and video sampling cost matters.
+
+## Camera Video
```
chrisgervang, line 70: Another complimentary example would be demuxing a video file with `mediabunny`.


## Timeline Scrubbing (Data Visualization Use Cases)

This setup uses the `VideoSampleSink` random-access method (`.getSample()`). Ideal for synchronizing a video timeline with analytical charts or slider inputs.

```ts

import { Input, ALL_FORMATS, BlobSource, VideoSampleSink } from 'mediabunny';

const input = new Input({
  formats: ALL_FORMATS,
  source: new BlobSource(videoFileBlob)
});

const videoTrack = await input.getPrimaryVideoTrack();

if (videoTrack) {
  const sink = new VideoSampleSink(videoTrack); 
  const slider = document.querySelector('#timeline-slider') as HTMLInputElement;

  slider.addEventListener('input', async () => {
    const targetTimeInSeconds = parseFloat(slider.value);
    
    // Fetch the sample matching the timestamp
    const sample = await sink.getSample(targetTimeInSeconds);
    
    if (sample) {
      // 1. Extract the underlying browser VideoFrame
      const frame = await sample.toVideoFrame();
      
      if (frame) {
        // 2. Map frame directly to your graphics texture context
        const videoTexture = new VideoTexture(device, { source: frame });
        
        // 3. Clean up the WebCodecs hardware resource instantly
        frame.close(); 
      }
      
      // 4. Clean up the Mediabunny object container
      sample.close();
    }
  });
}
```

chrisgervang (approved):
Super cool!! Suggested a use case around video data syntonization that I think luma and deck are uniquely positioned to excel at

docs/api-guide/gpu/video-textures.mdx
ibgreen-openai: @chrisgervang These sound like great example ideas, maybe we can capture them in a tracker task, unless you are ready to put up a PR?

PR #2765, opened 2026-07-30
RFC: simplify CanvasContext drawing buffer sizing

ibgreen:
## Motivation

deck.gl's canvas integrations are behavior-oriented:

- **Interleaved:** attach to an externally owned canvas/context and read its drawing-buffer size without writing it.
- **Overlaid:** render to a separate canvas whose drawing-buffer size follows the basemap canvas.

luma.gl currently exposes lower-level, overlapping sizing controls (`autoResize`, `useDevicePixels`, `pixelSizeSource`, and the boolean/object `createCanvasContext` union). This makes deck.gl reconstruct those behaviors through configuration merging and sizing-algorithm selection in [deck.gl #10370](https://github.com/visgl/deck.gl/pull/10370) and [deck.gl #10332](https://github.com/visgl/deck.gl/pull/10332).

This draft proposes fresh props that describe drawing-buffer ownership and source directly, deprecates the overloaded legacy props, and includes a tested proof of concept.

## Proposal

`drawingBufferSizeTracking` selects where the target drawing-buffer dimensions come from:

| Value | Behavior |
| --- | --- |
| `'none'` | The application owns the drawing-buffer size; luma.gl never changes it. |
| `'canvas'` | luma.gl derives the drawing-buffer size from the target canvas and resizes it when needed. |
| `'external-canvas'` | luma.gl mirrors `drawingBufferSizeSource.width/height`; if source and target are the same canvas, luma.gl only reads. |

The setting controls only the drawing buffer (`canvas.width` and `canvas.height`). It does not copy an external canvas's CSS size, style, position, transforms, borders, scroll state, or containing block. The target's existing CSS-size, position, visibility, and callback observation remains separate.

The default is `'canvas'`. With no fixed ratio it uses exact device-pixel observation when supported, falling back to CSS size × browser DPR. A numeric `pixelRatio` selects content-box sizing at that fixed ratio.

## API

```ts
export type DrawingBufferSizeTracking =
  | 'none'
  | 'canvas'
  | 'external-canvas';

export type CanvasContextProps = {
  drawingBufferSizeTracking?: DrawingBufferSizeTracking;
  drawingBufferSizeSource?: HTMLCanvasElement | OffscreenCanvas | null;
  pixelRatio?: number;

  /** @deprecated */
  autoResize?: boolean;
  /** @deprecated */
  useDevicePixels?: boolean | number;
  /** @deprecated */
  pixelSizeSource?: 'exact' | 'css-dpr';
};

export type DeviceProps = {
  createCanvasContext?: boolean | CanvasContextProps;
  canvasContextProps?: CanvasContextProps;
};
```

Validation is deliberately explicit:

- `drawingBufferSizeSource` is required for `'external-canvas'` and invalid for the other values.
- `pixelRatio` is valid only for `'canvas'`, must be finite and positive, and requires an explicit tracking value.
- New drawing-buffer props take precedence over legacy sizing props.
- `CanvasContext.setProps()` supports dynamic tracking, source, and ratio transitions and reconfigures the observer box when needed.

`canvasContextProps` is the preferred default/attached-context configuration. If the deprecated object form of `createCanvasContext` is also supplied, it is merged first and `canvasContextProps` wins.

## Migration

| Existing configuration | Preferred configuration |
| --- | --- |
| `autoResize: false` | `drawingBufferSizeTracking: 'none'` |
| `useDevicePixels: false` | `drawingBufferSizeTracking: 'canvas', pixelRatio: 1` |
| numeric `useDevicePixels` | `drawingBufferSizeTracking: 'canvas', pixelRatio: number` |
| exact automatic sizing | `drawingBufferSizeTracking: 'canvas'` |
| reproduce another renderer's CSS/DPR algorithm | `drawingBufferSizeTracking: 'external-canvas'` plus that canvas as `drawingBufferSizeSource` |

The legacy CSS × live browser-DPR algorithm remains supported internally through v9 normalization, but is intentionally not a separate value in the fresh API. Its main integration use case is represented more directly by tracking the external backing store.

The RFC proposes deprecation in v9 and removal of `autoResize`, `useDevicePixels`, `pixelSizeSource`, and object-form `createCanvasContext` in v10.

## PoC Implementation

The PoC:

- implements all three tracking behaviors in the shared `CanvasSurface`, covering both `CanvasContext` and `PresentationContext`;
- reads an external canvas's two integer backing-store dimensions at size-query and render boundaries;
- changes the target only when those dimensions differ, avoiding layout reads, polling, and observer-ordering dependencies;
- retains legacy exact and CSS-DPR behavior through internal normalization;
- supports dynamic tracking/source/ratio changes;
- adds and tests `DeviceProps.canvasContextProps` precedence; and
- leaves stable API docs and release notes unchanged while the RFC is under review.

The durable proposal is in `dev-docs/RFCs/vNext/canvas-context-configuration-rfc.md`.

## External Contexts

WebGL attachment defaults to:

```ts
{
  drawingBufferSizeTracking: 'external-canvas',
  drawingBufferSizeSource: gl.canvas
}
```

Because source and target are the same object, luma.gl updates bookkeeping before use but never writes the externally owned canvas. Explicit `canvasContextProps` can select target-canvas sizing or no tracking instead.

For an overlaid deck.gl canvas:

```ts
{
  canvas: overlayCanvas,
  drawingBufferSizeTracking: 'external-canvas',
  drawingBufferSizeSource: map.getCanvas()
}
```

This copies only the basemap canvas's actual drawing-buffer dimensions. Layout alignment remains the responsibility of deck.gl/the host application.

## Validation

- `nvm use` — Node v22.22.1
- `yarn lint fix` — passed
- Focused CanvasContext/CanvasObserver/device/WebGL attachment suite — 4 files, 44 tests passed
- `yarn build` — passed
- `yarn test` node phase — 53 files passed, 2 skipped; 195 tests passed, 2 skipped
- `yarn test` headless phase — 235 files passed; 1291 tests passed, 25 skipped; 3 known unrelated WebGPU/DGGS failures remain:
  - `modules/shadertools/test/modules/geospatial/dggs.spec.ts` (A5 produces NaN)
  - `modules/arrow/test/arrow/dggs-gpu-polygons.spec.ts` (A5 produces NaN)
  - `modules/arrow-layers/test/layers/arrow-layers.spec.ts` (storage-backed instances are not drawable)

## Open Questions

- Should `drawingBufferSizeSource` remain canvas-only or accept a structural `{width, height}` source?
- Should `'none'` retain current exact-device-pixel bookkeeping and callbacks, as the PoC does?
- Is the compatibility-only legacy CSS-DPR path sufficient for v9 now that new overlays can track the external backing store directly?

### DIFF
--- a/dev-docs/RFCs/vNext/canvas-context-configuration-rfc.md
+++ b/dev-docs/RFCs/vNext/canvas-context-configuration-rfc.md
@@ -1,21 +1,19 @@
-# RFC: Explicit CanvasContext Drawing-Buffer Sizing
+# RFC: Explicit CanvasContext Drawing-Buffer Tracking
 
 * **Date**: July 2026
 * **Status**: Draft
 
 ## Summary
 
-This RFC proposes fresh, non-overloaded `CanvasContext` properties:
+This RFC proposes fresh `CanvasContext` properties that describe the application's intended
+canvas integration:
 
 ```ts
-export type DrawingBufferSizingMode =
-  | 'manual'
-  | 'track-device-pixels'
-  | 'track-css-pixels';
+export type DrawingBufferSizeTracking = 'none' | 'canvas' | 'external-canvas';
 
 export type CanvasContextProps = {
-  drawingBufferSizingMode?: DrawingBufferSizingMode;
-  trackCanvas?: HTMLCanvasElement | OffscreenCanvas | null;
+  drawingBufferSizeTracking?: DrawingBufferSizeTracking;
+  drawingBufferSizeSource?: HTMLCanvasElement | OffscreenCanvas | null;
   pixelRatio?: number;
 };
 ```
@@ -50,32 +48,67 @@ The individual properties are overloaded or interact through implicit precedence
 
 These interactions leak into higher-level integrations. [deck.gl
 #10370](https://github.com/visgl/deck.gl/pull/10370) must merge nested
-`deviceProps.createCanvasContext` objects and narrow the boolean/object union to select legacy
-CSS-DPR sizing. [deck.gl #10332](https://github.com/visgl/deck.gl/pull/10332) demonstrates the
-branching required to interpret and forward `useDevicePixels: boolean | number`.
+`deviceProps.createCanvasContext` objects and narrow the boolean/object union to reproduce a
+basemap's CSS/DPR sizing algorithm. [deck.gl
+#10332](https://github.com/visgl/deck.gl/pull/10332) demonstrates the branching required to
+interpret and forward `useDevicePixels: boolean | number`.
 
-The proposed API directly represents the two canvas-integration behaviors used by deck.gl:
+That is a symptom of an abstraction mismatch. deck.gl has two concrete integration behaviors:
 
 - **Interleaved:** attach to an externally owned context and canvas. The external renderer changes
   the drawing-buffer dimensions; luma.gl reads them but never writes them.
 - **Overlaid:** render to a separate luma.gl canvas whose drawing-buffer dimensions mirror the
   basemap canvas.
 
-The lower-level self-sizing properties remain useful for standalone canvases, but deck.gl no
-longer needs to reproduce the basemap's CSS/DPR sizing algorithm to implement either integration.
+luma.gl currently exposes lower-level sizing algorithms, leaving deck.gl to reconstruct these
+behaviors. The proposed API represents drawing-buffer ownership and source directly.
 
 ## Proposal
 
-### Track another canvas
+### Drawing-buffer tracking
 
-`trackCanvas` makes another canvas's actual `width` and `height` authoritative:
+`drawingBufferSizeTracking` selects where the target drawing-buffer dimensions come from:
+
+| Tracking | Size source | Does luma.gl write the target `width` and `height`? | `pixelRatio` |
+| --- | --- | --- | --- |
+| `'none'` | Application | Never | Not allowed |
+| `'canvas'` | Target canvas CSS/physical size | Yes, when size changes | Optional fixed ratio |
+| `'external-canvas'` | `drawingBufferSizeSource.width/height` | Yes, unless source is target | Not allowed |
+
+The default is `'canvas'`, preserving standalone canvas auto-sizing. With no `pixelRatio`, luma.gl
+uses `ResizeObserver.devicePixelContentBoxSize` when available and falls back to content-box size
+multiplied by browser DPR. With a numeric `pixelRatio`, luma.gl observes the content box and uses
+`Math.floor(cssSize * pixelRatio)`. A ratio must be finite and greater than zero.
+
+```ts
+// The application owns absolute drawing-buffer dimensions.
+{drawingBufferSizeTracking: 'none'}
+
+// Track the target canvas at its exact physical pixel coverage when available.
+{drawingBufferSizeTracking: 'canvas'}
+
+// Disable device-pixel scaling while continuing to track target CSS resizes.
+{drawingBufferSizeTracking: 'canvas', pixelRatio: 1}
+
+// Use a fixed pixel ratio while continuing to track target CSS resizes.
+{drawingBufferSizeTracking: 'canvas', pixelRatio: 1.5}
+```
+
+In `'none'` mode, target canvas CSS size, position, visibility, and DPR observation continue.
+Resize callbacks still run, but luma.gl does not change the drawing buffer automatically.
+
+### Track an external canvas
+
+`'external-canvas'` tracking makes another canvas's actual backing-store `width` and `height`
+authoritative:
 
 ```ts
 await luma.createDevice({
   createCanvasContext: true,
   canvasContextProps: {
     canvas: overlayCanvas,
-    trackCanvas: map.getCanvas()
+    drawingBufferSizeTracking: 'external-canvas',
+    drawingBufferSizeSource: map.getCanvas()
   }
 });
 ```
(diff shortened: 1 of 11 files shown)


chrisgervang:
If we go with this it's likely safe to remove `pixelSizeSource`.. it might have technically released in a luma 9.3 patch, which is shame since it wasn't used in a deck release

dev-docs/RFCs/vNext/canvas-context-configuration-rfc.md
```diff
@@ -0,0 +1,212 @@
...
+### Add a policy object
+
+A discriminated policy object can encode every valid state statically, but adds nesting to the
+common configuration path. The proposed flat mode and optional ratio remain explicit without
+requiring another object.
+
+### Use ResizeObserver box names as modes
+
+Names such as `content-box` and `device-pixel-content-box` map directly to browser terminology,
+but expose an implementation detail and become misleading when exact device-pixel observation is
```
chrisgervang: I might favor these if it wasn't possible to be misleading, but agree with descriptive naming

modules/core/src/adapter/canvas-surface.ts
```diff
@@ -27,7 +30,15 @@ export type CanvasContextProps = {
   height?: number;
   /** Visibility (only used if new canvas is created). */
   visible?: boolean;
-  /** Whether to size the drawing buffer to the pixel size during auto resize. If a number is provided it is used as a static pixel ratio */
+  /** How luma.gl should resize the drawing buffer. */
```
chrisgervang: Could we explain the modes in the JSdoc?

modules/core/src/adapter/canvas-surface.ts
```diff
@@ -200,9 +229,29 @@ export abstract class CanvasSurface {
...
+        'drawingBufferSizingMode' in props &&
+        (drawingBufferSizingMode !== 'track-css-pixels' || previousUsesLegacyDrawingBufferSizing)
+      ) {
+        pixelRatio = props.pixelRatio;
+      }
+      validateDrawingBufferSizingProps(drawingBufferSizingMode, pixelRatio);
+      this._usesLegacyDrawingBufferSizing = false;
+      this._setDrawingBufferSizingProps(drawingBufferSizingMode, pixelRatio);
+    } else if ('useDevicePixels' in props && this._usesLegacyDrawingBufferSizing) {
+      this.props.useDevicePixels = props.useDevicePixels ?? false;
```
chrisgervang: Should we keep the old `||` behavior of explicitly setting to false if falsy for compatibility?

modules/core/src/adapter/device.ts
```diff
@@ -362,8 +362,14 @@ export type DeviceTextureFormatCapabilities = {
...
-  /** Properties for creating a default canvas context */
-  createCanvasContext?: CanvasContextProps | true;
+  /**
+   * Whether to create a default canvas context.
+   * Supplying a `CanvasContextProps` object is deprecated; set this to `true` and use
+   * `canvasContextProps` instead.
+   */
+  createCanvasContext?: CanvasContextProps | boolean;
+  /** Properties for creating or attaching the default canvas context. */
+  canvasContextProps?: CanvasContextProps;
```
chrisgervang, line 372: 👍

modules/webgl/src/adapter/webgl-adapter.ts
```diff
@@ -63,14 +63,21 @@ export class WebGLAdapter extends Adapter {
...
     // We create a new device using the provided WebGL context and its canvas
     // Assume that whoever created the external context will be handling resizes.
     return new WebGLDevice({
       ...props,
       _handle: gl,
-      createCanvasContext: {canvas: gl.canvas, autoResize: false, ...createCanvasContext}
+      createCanvasContext: true,
+      canvasContextProps: {
+        canvas: gl.canvas,
+        autoResize: false,
```
chrisgervang, line 94: Should this set the sizing mode to manual?

chrisgervang (approved):

chrisgervang:
> Should manual mode continue observing the device-pixel content box, as implemented, or use the cheaper content box when only callbacks/position are needed?

This seems like the risky part of this. If it comes up as a perf issue, we might first think to add a `performantDrawingBufferSizeTracking: Boolean` flag, but then it confuses the device pixel tracking mode. Maybe instead this could be solved with a 4th mode like `"manual-cheaper-observation"`

ibgreen:
@chrisgervang After looking at this again, it struck me that we are addressing the problem rather indirectly.

it seems to me that what deck wants is to either
- create a canvas that track the size of an external context (overlay mode)
- attach to a canvas that is managed externally (interleaved mode)

Perhaps the right API is to offer modes that take care of these things.

E.g. the fact that mapbox uses non-exact pixels and DPR multiplication is kind of an implementation detail that deck.gl doesn't really need to know about.

I am updating along these lines

chrisgervang:
What would interleaved deck.gl canvas use with this new pattern?

modules/core/src/adapter/canvas-surface.ts
ibgreen-openai: Addressed. `DrawingBufferSizeTracking` now documents all three modes directly in its exported TSDoc: application-owned (`none`), target-canvas-derived (`canvas`), and externally mirrored (`external-canvas`). The property docs also spell out the source and ratio constraints.

modules/core/src/adapter/canvas-surface.ts
ibgreen-openai: Yes—restored the legacy falsy normalization in `730438e90`. The setter uses `props.useDevicePixels || false` again, and a regression test verifies that numeric `0` becomes `false` and normalizes to ratio 1.

modules/webgl/src/adapter/webgl-adapter.ts
ibgreen-openai: The redesigned API uses `external-canvas` here rather than `none`/manual. An attached context defaults its canvas as both target and source, so luma.gl refreshes drawing-buffer bookkeeping from the external owner's `width`/`height` but never writes them. That preserves external ownership while keeping size queries current.

ibgreen-openai:
For interleaved deck.gl, attach the externally owned WebGL context. WebGLAdapter.attach now defaults the resulting CanvasContext to external-canvas tracking with gl.canvas as its source. Because source and target are the same canvas, luma.gl only refreshes size bookkeeping and never writes canvas.width or canvas.height; the basemap renderer remains the owner. Overlaid mode uses a separate target canvas with map.getCanvas() as the external source.

ibgreen-openai:
Will close temporarily until we refocus on this again

PR #2800, opened 2026-08-01
fix(shadertools): make fp64 reliable on Apple WebGPU

ibgreen-openai:
## Summary

- Add an optimizer-independent integer-controlled WGSL implementation of double-single arithmetic, selected automatically on Apple WebGPU/Metal while preserving the existing `vec2<f32>` API.
- Add assembly and real WebGPU compute coverage for automatic selection, explicit overrides, arithmetic primitives, boundary values, division, and square root.
- Turn the Mandelbrot example into an fp32/fp64 showcase with an opt-in WebGPU compute benchmark covering automatic, classic, integer-controlled, and native-f32 modes.
- Add shared precision-documentation tabs and embed the showcase in the guide and reference pages.
- Follow up on #2795 by documenting that the full `fp64` function library is GLSL-only; WebGPU/WGSL support belongs to the lower-level `fp64arithmetic` module.

## Why

Classic Dekker/Knuth error transforms depend on intermediate `f32` rounding points. WGSL permits reassociation and contraction, and Metal can optimize those residual-producing expressions away. The Apple path instead decodes each limb into integer sign, exponent, and significand fields, performs exact sum/product transforms and rounding with integer operations, then packs the result back into the public double-single representation.

## Apple Metal benchmark

The interactive benchmark ran 8,192 lanes, 32 dependent operations per lane, and three timed dispatches. All 16 operation/mode rows completed on Apple WebGPU/Metal. Representative maximum relative errors:

| Operation | Automatic | Integer-controlled | Classic | Native f32 |
| --- | ---: | ---: | ---: | ---: |
| Add | 4.87e-15 | 4.87e-15 | 1.14e-7 | 1.14e-7 |
| Multiply | 1.27e-13 | 1.27e-13 | 3.69e-6 | 2.01e-6 |
| Divide | 1.15e-13 | 1.15e-13 | 3.76e-6 | 2.02e-6 |
| Square root | 1.02e-14 | 1.02e-14 | 8.83e-8 | 1.29e-7 |

Timing is reported by the example using GPU timestamps when available, otherwise queue completion; no performance threshold is imposed.

## Validation

- `nvm use` — Node 22.22.1
- `yarn install` — passed (existing peer warnings only)
- `yarn lint fix` — passed on the final sources and again in the commit hook
- `yarn build` — passed
- Commit-hook node suite — 253 passed, 2 skipped
- Focused Apple WebGPU browser suite — 40 passed across shader assembly and fp64 arithmetic compute tests
- Live Apple WebGPU/Metal check — both Mandelbrot canvases rendered and all 16 benchmark rows completed
- `yarn website:build` — passed
- `(cd website && yarn build)` — passed on the exact final sources, including static generation and validation of 429 raw documentation pages
- `yarn test` — node passed. An earlier code-complete headless run reached 1,323 passed / 25 skipped with only the known Arrow polygon/text failure reproduced on `master`; the final shared-machine rerun was stopped after broad unrelated GPU/browser timeouts made it non-diagnostic.

## Limits

This remains double-single arithmetic rather than IEEE-754 binary64. It retains the `f32` exponent range and provides approximately 48 significand bits while both limbs are normal; precision tapers as the low limb becomes subnormal. The documented portable contract covers finite normalized double-single inputs and finite normal results.

### DIFF
--- a/docs/api-guide/shaders/gpu-floating-point-precision.md
+++ b/docs/api-guide/shaders/gpu-floating-point-precision.md
@@ -1,8 +1,9 @@
-import {ShaderLevelDocsTabs} from '@site/src/components/docs/shader-level-docs-tabs';
+import {ShaderModuleDocsTabs} from '@site/src/components/docs/shader-module-docs-tabs';
+import {FP64Example} from '@site/src/examples';
 
 # GPU Floating-Point Precision Techniques
 
-<ShaderLevelDocsTabs active="gpu-floating-point-precision" />
+<ShaderModuleDocsTabs group="precision" active="precision-guide" />
 
 "fp64" is often used to mean "more precision than `f32`," but the available
 techniques do not all implement the same number system. A pair of `f32` values,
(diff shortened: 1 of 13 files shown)


ibgreen-openai:
CI timeout follow-up: pushed 8a4bf9542. The failure came from forcing the Apple-oriented integer WGSL matrix on CI's software WebGPU/SwiftShader adapter; several pipelines stalled for 60 seconds and then invalidated the shared Dawn instance. The high-cost arithmetic, helper, division, and square-root matrix now skips software/CPU/fallback adapters, while the compact exact twoSum/twoProd integer smoke test still executes there and the full matrix still runs on hardware GPUs. Verification: CI-mode SwiftShader 14/14 in 6.8s, Apple WebGPU 40/40, and the replacement GitHub Actions run is fully green (examples 2m50s, test 7m02s, website 10m05s).

PR #2822, opened 2026-08-02
Add GPU trace batch index foundation

ibgreen-openai:
## Summary

- add indirect compute dispatch support to `Computation` and a reusable `DispatchCommandBuffer`
- introduce stable, group-aligned trace span batches with global source identity
- upload a compact GPU batch index containing span ranges, time bounds, lane bounds, group IDs, and batch IDs
- make synthetic trace source ordering temporally clustered so batch bounds are selective
- request the maximum WebGPU feature level while keeping the interactive example capped at a proven 4M spans
- expose batch counts, negotiated buffer limits, and a usable standalone controls panel
- increase density-bin resolution to make automatic scrolling visually smoother

## Why

The trace viewer currently binds and scans one canonical span table. A 10M-span allocation can fit on high-limit adapters, but replaying the complete graph every frame overwhelms presentation. This tranche establishes the data and dispatch contracts needed to select candidate batches and drive later chunked or streamed execution without changing stable span IDs or dependency semantics.

This is stacked on #2819 so the batch/index work remains separate from adaptive density LOD.

## Verification

- `yarn lint fix`
- `yarn build`
- `yarn test-node test/examples/gpu-trace-viewer.node.spec.ts` — 7 passed
- `yarn test-browser test/examples/gpu-trace-viewer.spec.ts` — 1 passed on hardware WebGPU
- pre-commit `test-fast` — 258 passed, 2 skipped
- live standalone viewer — 4,000,000 spans, 6 batches, 937,917 dependencies

## Follow-up

- evaluate the GPU batch index against the current viewport
- compact candidate batch IDs and write indirect dispatch commands
- split or stream exact-span storage and reuse batch-local scratch
- re-enable the 10M target once per-frame work scales with candidate batches rather than total spans

### DIFF
--- a/modules/experimental/src/gpu-primitives/dispatch-command-buffer.ts
+++ b/modules/experimental/src/gpu-primitives/dispatch-command-buffer.ts
@@ -1,0 +1,154 @@
+// luma.gl
+// SPDX-License-Identifier: MIT
+// Copyright (c) vis.gl contributors
+
+import {Buffer, type Device} from '@luma.gl/core';
+import {GPUData} from '@luma.gl/tables';
+
+const UINT32_BYTE_LENGTH = Uint32Array.BYTES_PER_ELEMENT;
+const DISPATCH_RECORD_WORDS = 3;
+
+/** CPU description of one WebGPU `dispatchWorkgroupsIndirect` record. */
+export type DispatchCommand = {
+  /** Workgroup count in the X dimension. */
+  x: number;
+  /** Workgroup count in the Y dimension. Defaults to `1`. */
+  y?: number;
+  /** Workgroup count in the Z dimension. Defaults to `1`. */
+  z?: number;
+};
+
+/** Properties for one typed WebGPU indirect-dispatch buffer. */
+export type DispatchCommandBufferProps = {
+  /** Buffer identifier. */
+  id?: string;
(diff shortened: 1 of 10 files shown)


modules/experimental/src/gpu-primitives/dispatch-command-buffer.ts
```diff
@@ -0,0 +1,154 @@
...
+};
+
+/** Properties for one typed WebGPU indirect-dispatch buffer. */
+export type DispatchCommandBufferProps = {
+  /** Buffer identifier. */
+  id?: string;
+  /** Record capacity. Defaults to `commands.length`. */
+  capacity?: number;
+  /** Optional initial CPU records. Unspecified capacity slots are zero-filled. */
+  commands?: DispatchCommand[];
```
ibgreen-openai, line 28: Is this a WebGPU concept or something more abstract we are defining?

ibgreen-openai: This directly models a native WebGPU indirect-dispatch record. dispatchWorkgroupsIndirect reads three consecutive u32 values for the x, y, and z workgroup counts; DispatchCommandBuffer gives that record a typed buffer wrapper plus ownership and GPUData views. It does not define a separate scheduling abstraction.

PR #2824, opened 2026-08-02
Add independent GPU trace capacity controls

ibgreen-openai:
## Summary

- split trace generation into independent span and dependency capacity controls
- calculate limit-aware dropdown options from the respective 32-byte span and 16-byte dependency records
- enforce the selected dependency upper bound while generating deterministic topology
- report generated dependencies versus the requested limit in the inspector

## Stack

- stacked on #2819 (`codex/trace-density-lod`)
- GPU batch candidate selection was promoted into #2819 to make that PR self-contained after review

## Verification

- `yarn lint fix`
- `yarn build`
- `yarn test-node test/examples/gpu-trace-viewer.node.spec.ts` — 8 passed
- `yarn test-browser test/examples/gpu-trace-viewer.spec.ts` — 1 passed on hardware WebGPU
- commit hook `test-fast` — 259 passed, 2 skipped

## Notes

The 10M options are limit-aware. The inspector reports generated dependencies versus the selected upper bound because sparse topology may naturally produce fewer edges.

### DIFF
--- a/examples/experimental/gpu-trace-viewer/app.ts
+++ b/examples/experimental/gpu-trace-viewer/app.ts
@@ -1187,6 +1187,9 @@ export default class GPUTraceViewerAnimationLoopTemplate extends AnimationLoopTe
       if (!(target instanceof HTMLInputElement || target instanceof HTMLSelectElement)) {
         return;
       }
+      if (target instanceof HTMLSelectElement && event.type !== 'change') {
+        return;
+      }
       if (target.matches('[data-span-capacity]')) {
         this.rebuild(Number(target.value), this.dependencyCapacity);
       } else if (target.matches('[data-dependency-capacity]')) {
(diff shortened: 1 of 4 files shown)


ibgreen-openai:
Addressed the duplicate-rebuild finding in e3681b5e1: capacity selects now ignore input events and rebuild only on change. The hardware WebGPU regression dispatches both events and verifies that the compile count advances exactly once.

The candidate-list concern is valid, but it is not introduced by this PR after the rebase: candidate batch selection came from the merged #2819 work now on master. That stage is currently preparatory indexing/telemetry; exact span visibility and density still scan the full span domain, so it does not yet provide the claimed O(N) reduction. This PR remains scoped to independent span/dependency capacity controls and does not claim that optimization. I am leaving the broader candidate-consumption work as a follow-up rather than presenting it as fixed here.

PR #2838, opened 2026-08-02
fix(core): skip buffer debug copies outside debug mode

### LINKED ISSUE #2785: Question: unconditional `_setDebugData` in `WEBGlBuffer`
hey, and huge thanks for this amazing library :)

I noticed that `_setDebugData` is called unconditionally in `WEBGlBuffer` (https://raw.githubusercontent.com/visgl/luma.gl/refs/heads/master/modules/webgl/src/adapter/webgl-device.ts), and I'm wondering if it's intentional. I'm working with some extremely performance sensitive code, and ran into this while looking at some graphs.

ibgreen-openai:
Fixes #2785.

## Summary

- skip `Buffer._setDebugData()` allocations and copies when device debugging is disabled
- keep existing debug-data behavior for `debug: true` devices
- cover WebGL construction, writes, and reads with debugging disabled
- document that the CPU-side buffer mirror is debug-only

The guard lives in the core `Buffer` helper so every backend using it gets the same behavior without duplicating checks at each WebGL call site.

@organize, does this match the overhead you observed? In particular, it leaves `buffer.debugData` empty unless the device was created with `debug: true`.

## Validation

- `nvm use`
- `yarn install`
- `yarn lint fix`
- `yarn build`
- `yarn test-node` (also rerun by the pre-commit hook): 70 files passed, 2 skipped; 293 tests passed, 2 skipped
- `yarn test-headless modules/core/test/adapter/resources/buffer.spec.ts`: 17 tests passed
- focused rerun of `modules/effects/test/passes/advanced-effects.spec.ts`: 4 tests passed

The complete `yarn test` headless phase was not fully green under heavy concurrent GPU-test load: 245 files passed and 13 failed, predominantly 60-second WebGPU timeouts. A timed-out effects suite passed immediately in isolation. One unrelated Arrow-layer assertion remained reproducible.

`yarn website:build` completed the server compilation, but the client build was interrupted after prolonged CPU contention and consequently failed its final `llms.txt` output check.

### DIFF
--- a/docs/developer-guide/debugging.md
+++ b/docs/developer-guide/debugging.md
@@ -105,7 +105,7 @@ shader is involved when debugging shader parsing errors occur.
 
 ## Buffer data inspection
 
-`Buffer` objects contain a CPU side copy of the first few bytes of data in the `buffer.debugData` field. This field is refreshed whenever data is written or read from the CPU side, using `buffer.write()`, `buffer.readAsync()` etc and can be inspected in the debugger to inspect the contents of the buffer.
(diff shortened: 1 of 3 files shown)


organize:
🥇 thank you for the quick fix!

PR #3132, opened 2026-08-25
feat(gpgpu): extend GPUDataEvaluator to support segmented data

Pessimistress:
#### Change List
- Accept optional `startIndices`


ibgreen:
Thanks @Pessimistress — I merged current master into this branch with a normal, non-force push, then added follow-up commit e880ecb52 to align segmented evaluators with the newer batched-data model.

- Resolved the #3134 overlap by keeping master’s numeric-input implementation.
- Read offsets from canonical GPUData.valueOffsets instead of Arrow readback metadata.
- Preserved chunk-local offsets and list formats through GPUVectorEvaluator batching and rematerialization, with explicit ownership and offset validation.
- Added value-list support and a two-chunk regression test.

Verification: yarn lint fix and yarn build pass; all 3,226 Node tests pass. The full headless run reached four failures in unrelated GPU raster/splats/lighting tests. CI is running on the pushed commits.

Issue #3177, opened 2026-09-06
[Bug] Default framebuffer keeps its old size when an attached canvas is resized by its owner


vlumi:
### Module and/or Feature

`@luma.gl/core` `CanvasSurface._resizeDrawingBufferIfNeeded` (9.4.0, also 9.4.0-beta.7 and `master`), with `@luma.gl/webgl` `WebGLCanvasContext._configureDevice`.

### Description

When a device is attached to an existing WebGL context (`luma.attachDevice(gl, ...)`, `autoResize: false`) and the owner of the canvas resizes it, the canvas context's default framebuffer keeps the old drawing-buffer size. Everything else follows the resize: the `ResizeObserver` updates `cssWidth`/`cssHeight`, the host calls `setDrawingBufferSize(canvas.width, canvas.height)`, and `drawingBufferWidth`/`drawingBufferHeight` are correct. But `_resizeDrawingBufferIfNeeded` only calls `_configureDevice()` when it has to write the new size onto the canvas, and here the canvas already has that size, so the `canvas-context-framebuffer` is never resized.

Because the default framebuffer's `height` is what consumers flip GL viewports against, every subsequent draw lands offset by `oldHeight - newHeight` pixels. Shrinking the window pushes the whole layer stack below the canvas; growing it leaves a gap at the top.

Phase 4 of #2743 ("Attached and externally owned canvas sizing") and its `Canvas sizing` row say attached contexts must respect the external owner's sizing authority. This is one place where they still do not: the owner sized the canvas, luma tracks the new size, but the framebuffer wrapper is left behind.

**Where it shows up**

deck.gl `MapLibreOverlay` / `MapboxOverlay` in interleaved mode: deck attaches to MapLibre's context, MapLibre resizes its own canvas, deck's `_onCanvasContextResize` calls `setDrawingBufferSize` with the canvas size (so luma's own state is right), then draws with a y offset. This is the remaining piece of visgl/deck.gl#9666 (closed) and looks like the same mechanism behind other "interleaved layers drift after resize" reports.

Probe from such an app after shrinking the container from 819 px to 479 px, read before the next frame:

```text
deck:   [1400, 479]
css:    [1400, 479]
db:     [1400, 479]
canvas: [1400, 479]   (width, height)
fb:     [1400, 819]   <- canvas-context-framebuffer
```

**Fix**

[...]

Issue #2743, opened 2026-07-17
Tracker: Device+Canvas lifecycle


ibgreen-openai:
## Why

deck.gl has accumulated a long stream of canvas, device, resize, attach, detach, async initialization, and teardown fixes. We should address the underlying ownership and lifecycle contracts in luma.gl instead of stacking integration-specific fixes in deck.gl.

One round of bigger root-level work is in [#2742](https://github.com/visgl/luma.gl/pull/2742).

## Ownership model

| Concern | Owner | Contract |
|---|---|---|
| luma `CanvasContext` / `PresentationContext` wrapper | `Device` | Wrappers created through a device are managed children and are destroyed on final device teardown unless destroyed explicitly first. |
| DOM canvas | Caller or integrating framework | luma never removes caller-supplied canvases. |
| Native WebGL/WebGPU handle | Creating or attaching caller | `destroy()` releases owned handles; `detach()` removes luma ownership and returns the reusable backend handle. |
| Canvas sizing | Standalone luma context or external host | Standalone contexts may resize their canvas; attached/shared contexts must respect the external owner’s sizing authority. |

## Evidence from deck.gl

| Family | Representative evidence | What it tells us | Root luma.gl action |
|---|---|---|---|
| Wrapper and observer teardown | [deck.gl#10355](https://github.com/visgl/deck.gl/issues/10355), [deck.gl#10356](https://github.com/visgl/deck.gl/pull/10356), [deck.gl#10266](https://github.com/visgl/deck.gl/issues/10266), [deck.gl#10078](https://github.com/visgl/deck.gl/issues/10078), [deck.gl#9817](https://github.com/visgl/deck.gl/issues/9817), [deck.gl#9853](https://github.com/visgl/deck.gl/issues/9853) | Dropped wrappers and late observer callbacks can retain canvases or access partially destroyed devices. | Device-owned wrappers, idempotent teardown, observer shutdown, and reverse-order destruction. |
| Async finalize and device reuse races | [deck.gl#9379](https://github.com/visgl/deck.gl/issues/9379), [deck.gl#9421](https://github.com/visgl/deck.gl/pull/9421), [deck.gl#8463](https://github.com/visgl/deck.gl/issues/8463) | React StrictMode and rapid create/finalize/create cycles can complete device creation after the original owner was finalized. `_reuseDevices` avoided crashes but previously prevented real teardown. | Reference-counted reusable devices, plus follow-up cancellation or late-result disposal for async creation. |
[...]

PR #3178, opened 2026-09-06
fix(core): resize the default framebuffer after an external canvas resize

vlumi:
For #3177

#### Background

When a device is attached to an externally owned canvas (`autoResize: false`), the owner resizes the canvas and then calls `setDrawingBufferSize()` with the size the canvas already has. `_resizeDrawingBufferIfNeeded()` saw no difference between the tracked size and the canvas, so it skipped `_configureDevice()`, and the default `canvas-context-framebuffer` kept its previous size. Consumers that flip viewports against the framebuffer height (deck.gl's interleaved MapLibre/Mapbox overlays) then draw offset by the height difference after every resize. See visgl/deck.gl#9666 and the `Canvas sizing` row of #2743.

#### Change List

- `CanvasSurface` remembers the drawing-buffer size the device was last configured for (the initial canvas size, which the subclass constructors configure at). `_resizeDrawingBufferIfNeeded()` reconfigures when the tracked size differs from that, whether or not it had to write the size onto the canvas itself. Startup and luma-driven resizes behave exactly as before; the only added reconfigure is the case where the canvas owner already resized the canvas. No extra work on WebGPU, whose `_configureDevice()` is unconditional.
- Regression test in `webgl-render-pass.spec.ts`: an externally resized canvas plus `setDrawingBufferSize()` resizes the default framebuffer on the next default render pass.

Reproduces on the 9.4.0 release and on `master`. Verified with the standalone repro from the issue (framebuffer 300x150 stays before the fix, becomes 640x480 with it) and in a deck.gl 9.4.0-beta.4 + MapLibre 6 interleaved app with this change applied to the installed `@luma.gl/core`: shrinking, growing, globe and mercator all stay aligned after a resize.

### DIFF
--- a/modules/core/src/adapter/canvas-surface.ts
+++ b/modules/core/src/adapter/canvas-surface.ts
@@ -140,6 +140,8 @@ export abstract class CanvasSurface {
   protected destroyed = false;
   /** Whether the drawing buffer size needs to be resized (deferred resizing to avoid flicker) */
   protected _needsDrawingBufferResize: boolean = true;
+  /** Drawing buffer size the device was last configured for; subclass constructors configure at the initial canvas size */
+  protected _configuredDrawingBufferSize: [number, number] = [0, 0];
 
   abstract get [Symbol.toStringTag](): string;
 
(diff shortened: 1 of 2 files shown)


ibgreen (approved):
Looks good, appreciate the fix.

felixpalmer:
Added to 9.4-release branch

PR #3179, opened 2026-09-07
fix(core): adopt external drawing buffer size

felixpalmer:
#### Background

Basemap browser demo in deck broken in 9.4

https://github.com/user-attachments/assets/ddca20fe-a83e-4166-9c04-f99246dfec69

Not sure if this is the intended design in luma, putting up this as a potential fix

#### Change List
- Detect out of sync buffer size and adopt in luma
- Tests

### DIFF
--- a/modules/core/src/adapter/canvas-surface.ts
+++ b/modules/core/src/adapter/canvas-surface.ts
@@ -441,9 +441,29 @@ export abstract class CanvasSurface {
       if (sizeChanged) {
         this.canvas.width = this.drawingBufferWidth;
         this.canvas.height = this.drawingBufferHeight;
-        this._configureDevice();
       }
+      // Reconfigure even when the canvas already had the requested size. An external owner of
+      // canvas sizing (e.g. a base map in deck.gl's interleaved mode) resizes the canvas itself
+      // and then reports the new size via setDrawingBufferSize(), so the canvas never needs to be
+      // written here. Backend resources derived from the drawing buffer size still need to be
+      // reconciled, so a requested resize must not be dropped just because the canvas agrees.
+      this._configureDevice();
+      return;
     }
+    // No resize was requested, but the canvas may still have been resized behind luma.gl's back:
+    // an owner that changes only the backing store (e.g. a host pixel ratio change) does not move
+    // the content box, so the ResizeObserver never fires and no size is ever reported.
+    this._adoptExternalDrawingBufferSize();
+  }
+
+  protected _adoptExternalDrawingBufferSize(): void {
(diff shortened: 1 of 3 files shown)


modules/core/src/adapter/canvas-surface.ts
```diff
@@ -441,9 +441,29 @@ export abstract class CanvasSurface {
...
         this.canvas.width = this.drawingBufferWidth;
         this.canvas.height = this.drawingBufferHeight;
-        this._configureDevice();
       }
+      // Reconfigure even when the canvas already had the requested size. An external owner of
+      // canvas sizing (e.g. a base map in deck.gl's interleaved mode) resizes the canvas itself
+      // and then reports the new size via setDrawingBufferSize(), so the canvas never needs to be
+      // written here. Backend resources derived from the drawing buffer size still need to be
+      // reconciled, so a requested resize must not be dropped just because the canvas agrees.
+      this._configureDevice();
```
ibgreen, line 450: The reconfigure is necessary when an external owner reports a genuinely new size, but making it unconditional also reconfigures on the initial pending flush even though each subclass constructor already configured that size. On WebGPU that means another `GPUCanvasContext.configure()` plus depth-resource churn. A coalesced A → B → A resize similarly performs work despite no net change. #3178 tracks the last configured dimensions, so it catches the reported external-resize case while preserving those no-op paths; I favor that narrower condition.

modules/core/src/adapter/canvas-surface.ts
```diff
@@ -441,9 +441,29 @@ export abstract class CanvasSurface {
...
+      return;
+    }
+    // No resize was requested, but the canvas may still have been resized behind luma.gl's back:
+    // an owner that changes only the backing store (e.g. a host pixel ratio change) does not move
+    // the content box, so the ResizeObserver never fires and no size is ever reported.
+    this._adoptExternalDrawingBufferSize();
```
ibgreen, line 456: This implicit adoption is broader than the reported bug and changes the sizing source of truth. It runs for `autoResize: true` too, where luma is meant to derive the drawing-buffer size from CSS/device pixels, and `canvas.width/height` are not necessarily the actual WebGL drawing-buffer dimensions. For attached contexts our documented contract has the external owner report `gl.drawingBufferWidth/Height` through `setDrawingBufferSize()`. #3178 repairs exactly that explicit path without silently replacing it. I suggest leaving automatic adoption for a separate ownership-contract change (with guards, callbacks, and backend cases defined).

ibgreen:
closing in favor of #3178.
the extra tests here could be worthwhile though.

PR #3192, opened 2026-09-08
fix(webgl): accept GLSL bool and struct array members in uniform block reflection validation

### LINKED ISSUE #3187: [Bug] 9.4.0 uniform block reflection validation rejects GLSL bool members and array-of-struct members that match in layout
### Description

Since 9.4.0, `@luma.gl/webgl` validates each WebGL uniform block by reflecting the linked program's layout and comparing it with the std140 metadata built from the module's `uniformTypes`. For two common cases the comparison fails even though the byte offsets agree, so every affected program logs a warning at level 0 and falls back to the supplied metadata:

```
luma.gl: WebGL uniform block reflection failed for "dataFilterUniforms"; using supplied std140 metadata. Failed to validate WebGL uniform block "dataFilterUniforms": reflected layout for "dataFilterUniforms.useSoftMargin" does not match supplied std140 metadata
luma.gl: WebGL uniform block reflection failed for "gouraudMaterialUniforms"; using supplied std140 metadata. Failed to validate WebGL uniform block "gouraudMaterialUniforms": reflected layout for "gouraudMaterialUniforms.unlit" does not match supplied std140 metadata
luma.gl: WebGL uniform block reflection failed for "lightingUniforms"; using supplied std140 metadata. Failed to validate WebGL uniform block "lightingUniforms": reflected layout for "lightingUniforms.lights[0].color" does not match supplied std140 metadata
```

Rendering is unaffected because the fallback is the pre-9.4 behaviour, but the warnings fire for luma's own `gouraudMaterial` and `lighting` modules and for any deck.gl 9.4 layer using them or `DataFilterExtension`, so they show up in most deck.gl applications.

### Root cause

Both are in `modules/webgl/src/adapter/helpers/get-shader-layout-from-glsl.ts`.

**1. GLSL `bool` reflects as `f32`, declarations use `i32`.** The GL-type table maps `GL.BOOL` to `"f32"` and `BOOL_VEC2..4` to `vecN<f32>`, while `GL.INT` maps to `"i32"`. luma's shader type vocabulary has no bool, so modules declare GLSL bools as `'i32'` (e.g. `unlit: 'i32'` in `gouraudMaterial`, `useSoftMargin: 'i32'` in deck's data filter), which also matches std140, where a bool occupies a 4-byte integer slot. `validateReflectedUniformBlock` then compares `"f32"` with `"i32"` and throws although both are 4-byte scalars at the same offset.

**2. Arrays of structs are flattened differently on each side.** `getUniformInfosFromTypes` turns `lights: [LightType, 5]` into a single entry named `lights[0].color` with `arrayLength: 5` and `byteStride` equal to the struct stride. The driver reports each struct member as its own non-array uniform: `lights[0].color` with `UNIFORM_SIZE` 1 and no array stride, then `lights[1].color`, and so on. The validator compares `arrayLength` 5 with 1 and the strides, and throws. The byte offsets match.

### Expected behaviour

Either map the `BOOL` family to the `i32` family in the reflection table, or treat 4-byte scalar formats as layout-equivalent in the validator; and expand declared struct arrays per element (or compare only byte offsets for struct members) before validating. With those two changes the three warnings above disappear and the validator still catches real layout divergence.

### Environment

- `@luma.gl/webgl` 9.4.0, `@luma.gl/shadertools` 9.4.0, `@luma.gl/core` 9.4.0
- deck.gl 9.4.0 (`@deck.gl/layers` `PolygonLayer`, `@deck.gl/extensions` `DataFilterExtension`)
- Chrome on macOS, WebGL2

### Reproduction

Render any deck.gl 9.4.0 `PolygonLayer`, or any layer with `DataFilterExtension`, on a WebGL2 device and open the console. The three messages appear once per uniform block per program link.

jsve:
Fixes #3187

### Background

Since 9.4.0 `@luma.gl/webgl` reflects each linked uniform block and compares it with the std140 metadata built from module `uniformTypes`. Two common declarations fail that comparison although their byte layouts agree, so every affected program logs a level-0 warning and falls back to the supplied metadata. This fires for luma's own `gouraudMaterial` / `lighting` modules and for deck.gl 9.4 layers that use them or `DataFilterExtension`.

### Change List

- `webgl-shadertypes.ts`: map `GL.BOOL` / `BOOL_VEC2..4` to `i32` / `vecN<i32>` instead of `f32`. A GLSL bool occupies a 4-byte integer slot in std140 and is set through the integer uniform setters (see `set-uniform.ts`), and luma's own modules already declare GLSL bools as `'i32'` (`unlit`, `enabled`, `invert`, the pbr `*Enabled` flags).
- `get-shader-layout-from-glsl.ts`: expand declared struct arrays per element (`lights[0].color`, `lights[1].color`, …, each `arrayLength: 1`, `byteStride: 0`). This matches how WebGL reports struct array members and how raw GLSL blocks were already reflected, so the validator still compares every member's offset and format strictly.
- Tests (first commit, red until the fix commit is applied on top):
  - `get-shader-layout-from-glsl.spec.ts`: the mocked reflection context now accepts per-uniform driver results; new cases cover bool members (raw and module-backed), per-element struct arrays, and a real offset divergence that must still warn and fall back.
  - `model-uniform-block-reflection.spec.ts` (engine, headless): builds a `Model` with `gouraudMaterial` and `lighting` on a WebGL device and asserts the linked blocks validate without a reflection fallback. This is the reproduction from the issue.

### Commits

The branch is two commits:

**test(webgl): cover uniform block reflection of GLSL bools and struct arrays**

Test-only. Extends the mocked reflection fixture and adds the new unit cases and the headless engine spec. Checked out by itself, the two specs report 5 failures highlighting the warnings from issue #3187 

**fix(webgl): accept GLSL bool and struct array members in uniform block reflection validation**

Source-only. The two changes in `webgl-shadertypes.ts` and `get-shader-layout-from-glsl.ts` described above.

### DIFF
--- a/modules/engine/test/lib/model-uniform-block-reflection.spec.ts
+++ b/modules/engine/test/lib/model-uniform-block-reflection.spec.ts
@@ -1,0 +1,68 @@
+// luma.gl
+// SPDX-License-Identifier: MIT
+// SPDX-FileCopyrightText: Copyright (c) vis.gl contributors
+
+import {expect, it, vi} from 'vitest';
+import {log} from '@luma.gl/core';
+import {Model} from '@luma.gl/engine';
+import {gouraudMaterial} from '@luma.gl/shadertools';
+import {getWebGLTestDevice} from '@luma.gl/test-utils';
+
+// References a GLSL bool member (`material.unlit`) and struct array members
+// (`lighting.lights[i]`) so both module blocks stay active after linking.
+const LIGHTING_VS = /* glsl */ `\
+#version 300 es
+out vec3 vColor;
+void main() {
+  vColor = lighting.ambientColor + lighting.lights[0].color + lighting.lights[4].position;
+  if (material.unlit) {
+    vColor = vec3(material.ambient);
+  }
+  gl_Position = vec4(0.0, 0.0, 0.0, 1.0);
+}
+`;
+
+const LIGHTING_FS = /* glsl */ `\
+#version 300 es
+precision highp float;
+in vec3 vColor;
+out vec4 fragColor;
+void main() {
+  fragColor = vec4(vColor, 1.0);
+}
+`;
+
+it('Model validates gouraudMaterial and lighting uniform blocks against WebGL reflection', async () => {
+  const webglDevice = await getWebGLTestDevice();
+  const fallbackMessages: string[] = [];
+  const logOnce = vi.spyOn(log, 'once').mockImplementation((_level, message) => {
+    fallbackMessages.push(String(message));
+    return () => {};
+  });
+
+  try {
+    const model = new Model(webglDevice, {
+      id: 'uniform-block-reflection-test',
+      topology: 'point-list',
+      vertexCount: 0,
+      vs: LIGHTING_VS,
+      fs: LIGHTING_FS,
+      modules: [gouraudMaterial]
+    });
+
+    const uniformBlockNames = model.pipeline.shaderLayout.bindings
+      .filter(binding => binding.type === 'uniform')
+      .map(binding => binding.name);
+    expect(uniformBlockNames, 'module uniform blocks are active').toEqual(
+      expect.arrayContaining(['gouraudMaterialUniforms', 'lightingUniforms'])
+    );
+    expect(
+      fallbackMessages.filter(message => message.includes('uniform block reflection failed')),
+      'reflected layouts match the module std140 metadata'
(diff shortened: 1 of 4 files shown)


jsve:
I was too slow :) Issue is already resolved though https://github.com/visgl/luma.gl/pull/3190

Closing this PR. Thanks for the quick resolve!

PR #3200, opened 2026-09-08
[9.4] Address example mobile support review feedback

ibgreen:
#### Background
Follow-up to #3198, which merged while its automated review was still running. This addresses both review findings from that PR.

#### Change List
- Commit the generated mobile-support bootstrap and metadata into all 75 standalone example HTML pages.
- Add regression assertions for standalone bootstrap installation and canonical example IDs.
- Restore each shared canvas's previous `touchAction` value when the five affected examples finalize.

#### Verification
- `nvm use`
- `yarn lint fix`
- `yarn build`
- `yarn examples:typecheck`
- `yarn website:build`
- `(cd website && yarn build)`
- `yarn test-node`: 2,932 passed, 3 skipped
- `yarn test`: node suite passed; headless suite passed 1,812 tests and reproduced the same six unrelated WebGPU failures in untouched DGGS, raster, splats, and scene-renderer tests.

### DIFF
--- a/examples/api/animation/index.html
+++ b/examples/api/animation/index.html
@@ -1,4 +1,22 @@
 <!doctype html>
+<head>
+  <!-- luma-example-support -->
+  <meta name="luma-example-id" content="api/animation" />
+  <meta name="luma-example-backends" content="webgpu,webgl2" />
+  <meta name="luma-example-mobile" content="full" />
+  <meta name="luma-example-mobile-profile" content="standard" />
+  <script type="module" data-luma-example-support-bootstrap>
+    import {installStandaloneExampleSupport} from '../../example-support.ts';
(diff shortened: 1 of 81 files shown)


ibgreen:
The three automated findings posted after this PR merged are addressed in #3202:

- Standalone application imports are now conditional dynamic imports gated on the support preflight, while remaining visible to Vite for production bundling.
- GPGPU, GPU Sort, and GPU Data Analysis remain in `loading` until initialization returns a handle.
- Unsupported-environment messaging is now device-neutral.

Regression coverage verifies all 75 standalone pages await preflight and contain no static application imports. There were no unresolved inline review threads on this merged PR to resolve; the findings were posted as this top-level review summary.

PR #3205, opened 2026-09-09
fix(examples): scope legacy deck.gl shader shim to initialization

akre54:
## Summary

- Moves the legacy deck.gl shader-assembler compatibility shim into the shared `ArrowDeck` wrapper so every Deck example gets the fix without duplicating lifecycle callbacks.
- Keeps the shim scoped to deck.gl's single legacy no-argument assembler lookup during `LayerManager` construction, then immediately restores strict explicit-language behavior.
- Preserves caller-provided `onDeviceInitialized` callbacks and restores an unconsumed shim during finalization.
- Renames the luSpatial taxi export to `createGPUSpatialTaxiDeck`, matching the website loader and standalone entry point.
- Adds focused Vitest coverage for callback ordering, WebGPU assembler selection, immediate restoration, and finalization cleanup.

## Verification

- [x] `nvm use`
- [x] `yarn install --immutable`
- [x] `yarn lint fix`
- [x] `yarn build`
- [x] `yarn test-node` — 410 files passed, 1 skipped; 3271 tests passed, 7 skipped
- [x] `yarn examples:typecheck`
- [x] Focused lifecycle test: `test/examples/deck-example-device.node.spec.ts`

The compatibility shim can be removed after the examples move to a deck.gl release that passes the known shader language to `ShaderAssembler.getDefaultShaderAssembler(language)`.

### DIFF
--- a/examples/deck/deck-example-device.test.mjs
+++ b/examples/deck/deck-example-device.test.mjs
@@ -1,0 +1,107 @@
+// luma.gl
+// SPDX-License-Identifier: MIT
+// SPDX-FileCopyrightText: Copyright (c) vis.gl contributors
+
+// Regression coverage for installLegacyDeckShaderAssemblerCompatibility(): deck.gl calls
+// ShaderAssembler.getDefaultShaderAssembler() with no argument on every device init AND on
+// every subsequent render/pick, so the patch must stay installed across repeated no-argument
+// calls until the caller explicitly restores it. A prior version restored automatically after
+// the first no-argument call, which crashed deck.gl website examples with
+// "shadertools: assertion failed." on the very next render or pointer pick.
+
+import assert from 'node:assert/strict';
+import test from 'node:test';
+
+function makeFakeShaderAssembler() {
+  const glslAssembler = {language: 'glsl'};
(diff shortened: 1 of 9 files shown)


examples/deck/deck-example-device.test.mjs
```diff
@@ -0,0 +1,107 @@
...
+  // Mocks are scoped to this test's own tracker, so each test gets an independent module mock
+  // (node:test rejects re-mocking a specifier that a prior test's tracker never restored).
+  t.mock.module('@luma.gl/shadertools', {namedExports: {ShaderAssembler}});
+  t.mock.module('@luma.gl/webgl', {namedExports: {webgl2Adapter: {}}});
+  t.mock.module('@luma.gl/webgpu', {namedExports: {webgpuAdapter: {}}});
+  // Cache-bust so each test gets a fresh module instance bound to its own fake ShaderAssembler.
+  return import(`./deck-example-device.ts?test=${Math.random()}`);
+}
+
+test('keeps the legacy patch installed across repeated no-argument calls', async t => {
```
akre54, line 40: Not sure how necessary this test file is but just added to catch future regressions

PR #3290, opened 2026-09-17
feat(shadertools): add public resetShaderHooks() method

akre54:
## Summary
Adds a public `resetShaderHooks()` method to `ShaderAssembler`, replacing the brittle type-casting pattern that downstream consumers (particularly deck.gl) currently use to clear shader hooks.

## Background
After the refactor in #3283 (commit 6bbb201), deck.gl manages shader assembler instances per-language and needs to clear hooks when switching between GLSL and WGSL or during re-initialization. Without a public API, consumers must resort to:

```typescript
(shaderAssembler as any)._hookFunctions.length = 0;
```

This breaks encapsulation, couples to private implementation details, and fails if luma.gl refactors internals.

## Changes
- **modules/shadertools/src/lib/shader-assembler.ts**: Add public `resetShaderHooks()` method to `ShaderAssembler` base class
- Clear `_hookFunctions` in-place (`length = 0`) to preserve reference stability
- JSDoc with usage example for GLSL→WGSL switching
- **modules/shadertools/test/lib/shader-assembler.spec.ts**: Add test verifying hooks clear completely and new ones can be added post-reset

## Example Usage
```typescript
// deck.gl pattern when switching shader languages
const glslAssembler = new GLSLShaderAssembler();
glslAssembler.addShaderHook('vs:HOOK(inout vec4 value)');

// Switch to WGSL
glslAssembler.resetShaderHooks();
const wgslAssembler = new WGSLShaderAssembler();
wgslAssembler.addShaderHook('vs:HOOK(value: ptr<function, vec4<f32>>)');
```

## Notes
- Small compatibility fix for deck.gl and other downstream consumers
- May become less critical once PR #3288 (singleton removal) fully stabilizes
- Scoped to just the method, doc, and test—no refactoring


ibgreen:
Thanks for taking this on — having a public API for hook cleanup is a useful goal, and the isolated-assembler test is helpful.

I’m concerned that resetShaderHooks() addresses the wrong ownership boundary, though. _hookFunctions is already instance-owned, while the default fallback assemblers are shared per language. Resetting a shared GLSL assembler can therefore clear hooks installed by unrelated consumers. The example also resets a GLSL assembler and then creates a WGSL assembler, so it does not actually switch the state of one assembler.

With the explicit assembler injection in #3288, I think the cleaner direction is for deck.gl to own language-specific assembler instances and keep their hooks isolated. If this API is retained, could we document that it clears all hooks for that assembler and add a test using the shared/default assembler to make the cross-consumer behavior explicit?

The public API goal makes sense — I’d just prefer the fix to preserve hook ownership per assembler (or at least per language) so one consumer cannot affect another. Appreciate the effort here!

akre54:
Makes sense, thanks. This API could encourage misuse. Closing in favor of #3288 which is more comprehensive

PR #3301, opened 2026-09-18
perf(shadertools): make WGSL binding allocation order independent and stop scanning every shader twice

akre54:
Two independent findings in the WGSL assembly path — one a correctness bug, one pure duplicated work. Both are small and self-contained; **happy to split this into two PRs if you'd prefer to review them separately.**

## 1. Explicit module bindings are now reserved before auto allocation

`reserveModuleBindings()` (was `reserveRegisteredModuleBindings`) only pre-reserved bindings already present in the binding registry — i.e. ones a *previous* assembly on the same assembler had assigned. Explicit `@group(N) @binding(M)` declarations in module source were never reserved, so a module declaring `@binding(auto)` that happened to be assembled first could take a location another module declares explicitly, and assembly threw `Duplicate WGSL binding assignment`.

**Whether a shader compiled depended on the order the caller happened to list modules in.** Reproduced with two modules in group 2, one explicit at binding 1 and one auto — listing the auto module first threw.

Explicit module bindings are now claimed in a pass that runs before any auto allocation, so allocation is a function of the module *set*. The explicit relocation path claims its own reservation rather than registering the location a second time.

One behavioural change worth noting: conflicting explicit bindings are now reported by the reservation pass, which names **both** modules rather than only the second one. The existing test assertion was updated for the new message.

## 2. The WGSL debug tail no longer runs on every assembly

`getShaderBindingDebugRowsFromWGSL` and `scanWGSLInterface` each ran **twice** per assembly: `assembleWGSLShader` computed both, then `WGSLShaderAssembler` discarded them and recomputed from the preprocessed source. The binding table is debug metadata that no render path reads, so both are now computed on first access.

Measured with a direct in-process A/B — two arms interleaved twice, 300 assemblies per sample, 9 samples per arm:

| | per assembly |
|---|---|
| eager (before) | 145.3 µs |
| lazy (after) | 93.7 µs |
| **reduction** | **35%** |

`performance.now()` is clamped to 0.1 ms in a page that isn't cross-origin-isolated, so that's batch wall-clock, not a sum of per-call timers.

## 3. `getShaderModuleDependencies({canonicalOrder})` — added, off by default

Reordering modules produces assembled sources that are permutations of each other: identical lines, identical binding assignments, different bytes. The same program therefore misses in the shader and pipeline caches and builds duplicate pipelines. `canonicalOrder` orders modules with no dependency relation by name, making the result a function of the module set.

**It is off by default and no caller enables it**, because it isn't safe for assembly yet: GLSL order is part of the language, and modules here rely on caller order instead of declaring their dependencies — the gpgpu `source_values_texture` module needs a `TYPE` define supplied by another module, and sorting broke 9 gpgpu tests. The option and its test are included so the finding isn't lost; enabling it needs those undeclared dependencies declared first. Glad to drop it from this PR if you'd rather not carry a disabled option.

## Testing

- `yarn test-node`: 3677 passed, 8 skipped — **+5 tests** over master's 3672.
- `yarn test-headless`: 2527 passed, 1 failed — **identical to master**, where `test/examples/gpu-trace-viewer.spec.ts` already fails (verified by running clean master).

New tests cover an explicit binding keeping its location across three module orderings with identical assignments and no collisions; conflicts being reported in either order; what reordering does and doesn't change; and `canonicalOrder` being independent of caller order while the default preserves it.

## Related work not in this PR

While investigating I also found that a `Model` drawn into render passes with differing attachment formats (a shadow or picking pass alongside the main pass) rebuilds its pipeline on **every** alternation and takes a fresh `PipelineFactory` reference each time, so `useCount` grows without bound. Measured on the `ShadowMapRenderer` shape — a depth-only target alternating with a colour target, 100 frames:

| | before | after a local fix |
|---|---|---|
| pipeline rebuilds | 200 | 2 |
| references held | 201 | 3 |
| `useCount` after `destroy()` | `[100, 100]` | `[0, 0]` |

Both `_setPipelineNeedsUpdate('attachment formats')` and the single-release `destroy()` are still present on master. I have a working fix but it predates the `_preparePipelineUpdate`/`_finishPipelineUpdate` refactor and the new `_updatePipelineAsync` path, so it needs reworking against those before it's proposable — I'd rather land it separately than bundle a rewrite in here. Let me know if you'd like me to open that as a follow-up.


ibgreen:
Thanks for identifying and reproducing the explicit/auto binding collision and the redundant metadata scans. Both findings were useful. We addressed them in two focused PRs, which are now merged:

- #3304 — reserve explicit WGSL bindings before allocation.
- #3305 — scan WGSL metadata only after final preprocessing.

We chose a slightly different implementation for each:

**Binding allocation:** #3304 preprocesses module sources once, reserves explicit declarations and active registry assignments, and resolves bindings before emitting source. It preserves module emission order and existing auto-binding assignments. Declaration validation remains separate from registry reuse, so repeated explicit declarations—including identical declarations within one module—still throw instead of being treated as reuse of the same reservation. Tests cover all six permutations of an explicit module and two auto modules, plus repeated assembly, conditional declarations, and conflicts with existing registry assignments.

The intended guarantee is collision avoidance around explicit bindings, rather than full order independence: reversing two auto-binding modules on fresh assemblers can still swap their assigned locations. Existing registry assignments also deliberately preserve assembly history.

**Metadata scanning:** #3305 introduces an internal source-only assembly step, then computes metadata once from the final preprocessed source. The lower-level assembly API continues returning eager metadata. We preferred this to lazy getters because `Model` immediately destructures `bindingTable`, which would trigger the getter during construction anyway. A regression test verifies that both scans run exactly once even when the returned metadata is consumed immediately. We have not carried over the 35% figure as an end-to-end Model performance claim.

**Canonical ordering:** We left this out for now. With no assembly caller enabling it, the option would add API surface without delivering the cache benefit yet. It would be better to revisit it separately with evidence of duplicate cache entries and tests covering dependencies, preprocessing, and injections before changing emission order.

Both replacement PRs passed CI. Thanks again for surfacing these issues and providing the initial investigation.

ibgreen:
Closing this PR as superseded.

The two actionable fixes have landed separately:

- #3304 reserves explicit WGSL bindings before auto-allocation.
- #3305 removes the redundant WGSL metadata scans.

The optional canonical module-ordering change is being deferred because it can affect shader semantics and requires additional dependency work.

PR #3302, opened 2026-09-18
perf(engine): cache one render pipeline per attachment-format variant

akre54:
Stacks on #3301. GitHub will not accept a fork branch as a PR base, so this targets `master` and its diff currently carries #3301's commit as well — **review the top two commits only** (`perf(engine): cache one render pipeline per attachment-format variant` and `perf(engine): compute the WGSL binding debug table on first access`). The diff collapses to just those once #3301 merges.

## The problem

A `Model` drawn into render passes with different attachment formats — the common case being a screen pass plus a picking or shadow pass — rebuilt its pipeline on every alternation. `_syncAttachmentFormats()` marks the pipeline dirty whenever the formats differ from the last draw, and `_updatePipeline()` then ran a full rebuild.

The rebuild always hit `PipelineFactory`'s cache, so it produced no new GPU pipeline. But it still paid a full `_preparePipelineUpdate()` and took a fresh `PipelineFactory` reference each time, so `useCount` grew without bound — which also defeated eviction for applications running with `_destroyPipelines: true`, since `destroy()` released exactly one reference.

## Measured

One WebGPU model, 200 render passes alternating between an `rgba8unorm` and an `rgba16float` target, both variants warmed first. Timing covers `_syncAttachmentFormats()` + `_updatePipeline()` only, excluding `beginRenderPass`/`submit`; 3 runs each.

| | before | after |
|---|---|---|
| pipeline rebuilds over 200 passes | 200 | **0** |
| total factory `useCount` after 200 passes | 3 → 203 | **3 → 3** |
| `useCount` still held after `destroy()` | **202** | **0** |
| per-pass cost | 3.5–8.5 µs | 1.0–1.5 µs |

The remaining ~1 µs is `_syncAttachmentFormats()` itself — reading the framebuffer's formats and comparing them — which is unavoidable per-pass work. The pipeline update on top of it is now free.

The 202 leaked references are the part worth looking at: that is one per alternation, never handed back.

## Approach

Keep one pipeline per attachment-format signature, bounded at 8 variants, and move reference ownership into the cache. Each entry owns one `PipelineFactory` reference plus one `ShaderFactory` reference per distinct shader, released on stale wipe, eviction, or destroy. This replaces `_finishPipelineUpdate()`'s previous immediate release of the outgoing shaders — keeping both would double-count the same references.

Obsolete variants are released only *after* the replacement exists, so a shader or pipeline shared with the replacement is never destroyed and recreated.

Two placement details a reviewer may want to check:

- **The reuse check runs before `_preparePipelineUpdate()`**, not inside it, because that function creates shaders (taking `ShaderFactory` references) that a cache hit does not need.
- **`_updatePipelineAsync()` needs no in-flight bookkeeping.** Its only caller is `_initializePipelineAsync()`, once per model from the constructor, and `PipelineFactory._createPipelineAsync()` already shares one pending backend compilation per hash. On failure, only `_finishPipelineUpdate()` inserts, so the existing `catch` leaves prior variants untouched and valid.

Eviction never drops the currently bound pipeline, whose reference would otherwise be left without an owner.

## Second commit

`assembleWGSLShader()` returns `bindingTable` as a lazy getter (added in #3301), but `Model` destructured it straight out of the result, forcing the scan on every model construction. Nothing on a render path reads it — it is debug metadata for `getBindingDebugTable()`. This is what makes #3301's getter actually pay off.

## Tests

Five new headless WebGPU tests in `modules/engine/test/lib/model.spec.ts`, covering reuse across alternating passes, full release on destroy, discarding variants when the pipeline shape changes, the eviction bound, and the lazy binding table. They assert on `_pipelineCache` size and on factory reference counts, so a future regression in reference ownership fails loudly.

Also raises the `waitForPipelineError` poll budget from 500 ms to 5 s. The new tests create more pipelines in parallel, which is the load profile that made the old budget give up early and report what looked like a product regression.

## Validation

- `yarn lint` clean.
- `yarn test-node`: 3677 passed, 0 failed.
- `yarn test-headless`: 2532 passed, 1 failed — `gpu-trace-viewer`, which I confirmed fails identically on the base commit with the same `expected 0 to be greater than 0`.


ibgreen:
Updated the branch to rebase onto current `master`, removing the superseded #3301 commits now covered by #3304 and #3305.

The remaining change focuses on the pipeline variant cache. It now uses structured attachment-format keys and LRU eviction so frequently reused render targets remain cached while the cache stays bounded.

Validation: `yarn build` passed and the focused model tests passed. The full suite is currently blocked by an unrelated `@luma.gl/gpgpu/gpu-vector-search` package-export failure.

